PR #1, opened 2023-02-10
Add report fields and drop message for empty table

jbartletthmf:


report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/PatientClinicalHistoryGenerator.java
```diff
@@ -1,8 +1,8 @@
 package com.hartwig.actin.report.pdf.tables.clinical;
 
-import java.util.List;
-import java.util.Set;
-import java.util.StringJoiner;
+import java.util.*;
```
pauldwolfe, line 3: i think we have * imports disabled in our code style.

report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/PatientClinicalHistoryGenerator.java
```diff
@@ -181,18 +168,31 @@ private static String secondPrimaryHistory(@NotNull ClinicalRecord record) {
...
+            return Optional.empty();
         }
     }
 
-    @Nullable
-    private static String toStopReasonString(@Nullable String stopReason) {
-        return stopReason != null ? "stop reason: " + stopReason : null;
+    @NotNull
+    private static Optional<String> toNumberOfCyclesString(@Nullable Integer cycles) {
+        return cycles != null ? Optional.of(cycles + " cycles") : Optional.empty();
```
pauldwolfe, line 190: You can use Optional.ofNullable().map() here instead.

pauldwolfe (changes requested):
Some nice simplifications! 

Couple really small things in the code.

When merging can we:
- Add the ticket number to the commit header
- Add a small description of the change, including rationale for the refactoring

report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/PatientClinicalHistoryGenerator.java
jbartletthmf: Yes, that's better.

jbartletthmf:
@pauldwolfe I'm trying to learn your commit message conventions but I could use some guidance. I'm used to including the ticket ID in the message, but it was added in the message body in some examples in the reference article. Should the header look like: `ACTIN-11: Add report fields and drop redundant message`?

I'm not sure how detailed the description of the change should be. The message will reference https://hartwigmedical.atlassian.net/browse/ACTIN-11, so I don't think we need to reproduce what's in there. Should I include the top-level requirements? Something like:
```
Remove “* Variant has > 50% likelihood of being sub-clonal” when there
is no data in the drivers table.

Add treatment end date to treatments in ‘relevant systemic treatment
history’.

Add treatment number of cycles to treatments in ‘relevant systemic
treatment history’.

Refactor to handle empty values with Optional, facilitate String
building with streams, and reduce unnecessary list traversals.
```

Also, how do you apply the summary message when merging? Can I just insert it in github when making the merge commit (if I'm the one who will merge) or do I need to squash the commits externally?

jbartletthmf:
@pauldwolfe in this case, would this message be appropriate? Other than refactoring, the changes are very straightforward.
```
ACTIN-11: Add report fields and drop redundant message

Refactor to handle empty values with Optional, facilitate String
building with streams, and reduce unnecessary list traversals.
```

Also, are there any other changes you want to see before approving for merge?

pauldwolfe (approved):

pauldwolfe:
@jbartletthmf Looks good!

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/MolecularDriversGenerator.java
```diff
@@ -69,7 +69,7 @@ public Table contents() {
             table.addCell(Cells.createContent(nullToEmpty(entry.bestResistanceEvidence())));
         }
 
-        if (hasPotentiallySubclonalVariants(molecular.drivers().variants())) {
+        if (!entries.isEmpty() && hasPotentiallySubclonalVariants(molecular.drivers().variants())) {
```
kduyvesteyn, line 72: I don't quite like this, entries and "molecular driver variants" are different datamodels. If you go into the factory creating the driver entries you can see that only "reported" variants get converted. This means that previously, unreported variants would lead to a footer note without actually being present. Also, your fix doesn't fix every situation since there could be other entries (like amps and dels) that would never be subclonal.

Maybe a better solution is to add an explicit "isPotentiallySubclonal" to MolecularDriverEntry datamodel, and add the asterix within the table rather than the datamodel factory, and subsequently the decision to add footnote yes/no becomes trial as well?

jbartletthmf: Good point about the other entries that could produce the footnote as well. We should be able to unambiguously determine if the footnote is needed by filtering the set of variants, i.e.:
```
        boolean hasPotentiallySubClonalVariants = molecular.drivers().variants().stream()
                .filter(Driver::isReportable)
                .anyMatch(ClonalityInterpreter::isPotentiallySubclonal);
```

I'll open a new PR to address this concern.

kduyvesteyn: Thanks!

PR #2, opened 2023-02-14
Add lymph node metastases to data model and implement rule

jbartletthmf:
ACTIN-16: Add lymph node metastases to data model and implement rule

Refactored to clean up duplicate code.

https://hartwigmedical.atlassian.net/browse/ACTIN-16


jbartletthmf:
I intend to try this out on data-vm before merging.

algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasLymphNodeMetastasesTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.hartwig.actin.algo.datamodel.EvaluationResult;
+import org.junit.Test;
+
+import static com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation;
+import static org.junit.Assert.assertTrue;
+
+public class HasLymphNodeMetastasesTest {
+
+    @Test
+    public void canEvaluate() {
```
pauldwolfe, line 13: what about breaking this guy up into three tests for the different evaluation results? Would also result in an improved test name.

algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorUtil.java
```diff
@@ -0,0 +1,32 @@
...
+import com.hartwig.actin.algo.datamodel.Evaluation;
+import com.hartwig.actin.algo.datamodel.EvaluationResult;
+import com.hartwig.actin.algo.datamodel.ImmutableEvaluation;
+import com.hartwig.actin.algo.evaluation.EvaluationFactory;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class TumorUtil {
+    @NotNull
+    public static Evaluation evaluateBooleanMetastasis(@Nullable Boolean hasMetastases,
```
pauldwolfe, line 12: Are we testing this via the HasXxxMetastases tests? Another approach could be to unit test this code directly and invert the dependency such that we inject a non-static implementation of this into each rule

database/src/main/resources/generate_database.sql
```diff
@@ -37,6 +37,7 @@ CREATE TABLE tumor
     hasBoneLesions BOOLEAN,
     hasLiverLesions BOOLEAN,
     hasLungLesions BOOLEAN,
+    hasLymphNodeLesions BOOLEAN,
```
pauldwolfe, line 40: How are database patches handled?

algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorUtil.java
jbartletthmf: Yes, I noticed that each of those evaluation functions was doing exactly the same thing so this is a simple refactor and the existing `Has*Metastases` tests still apply.

Do you think non-static usage of the util would be an improvement here? I'm still calibrating my functional/OO leanings after switching back from Scala.

I like that we currently have tests that verify and document each specific evaluation function regardless of whether they make use of a shared utility. It wouldn't hurt to test the `TumorUtil` in addition to each evaluation function though.

algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasLymphNodeMetastasesTest.java
jbartletthmf: I was following the convention from similar test classes but I agree with making this improvement.

algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorUtil.java
jbartletthmf: I looked into doing this but the only thing that varies between evaluation criteria is the String representing the metastasis type, so I can't think of a test we'd want to add now that would make use of injection. I'm happy with just reducing duplication for now but we can refactor further if we find more use cases for the `TumorUtil`.

pauldwolfe (approved):

jbartletthmf:
This change was applied on data-vm and all data was reloaded but there are no rows in `tumor` that satisfy `hasLymphNodeLesions is not null`. I don't see this field specified in `/data/actin/clinical/curated/` and I think we would have to add it in `com.hartwig.actin.clinical.feed.questionnaire.Questionnaire` to make it available. That seems to be out of scope here so I'll merge this PR and we can create a new ticket for any additional work.

PR #3, opened 2023-02-15
ACTIN-17: Add rule HAS_BMI_OF_AT_MOST_X

jbartletthmf:
https://hartwigmedical.atlassian.net/browse/ACTIN-17


pauldwolfe (approved):

pauldwolfe (approved):

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
```diff
@@ -0,0 +1,69 @@
...
+    static final double MIN_EXPECTED_HEIGHT_METRES = 1.5;
+    static final double MAX_EXPECTED_HEIGHT_METRES = 2.0;
+
+    private final int maximumBMI;
+
+    HasMaximumBMI(final int maximumBMI) {
+        this.maximumBMI = maximumBMI;
+    }
+
+    private double calculateWeightForBmiAndHeight(double bmi, double height) {
```
kduyvesteyn, line 27: I generally put private functions below the public evaluate function, feels more logical in terms of ordering. Also, is there a reason why this function is non-static?

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
```diff
@@ -0,0 +1,69 @@
...
+import com.hartwig.actin.algo.datamodel.ImmutableEvaluation;
+import com.hartwig.actin.algo.evaluation.EvaluationFactory;
+import com.hartwig.actin.algo.evaluation.EvaluationFunction;
+import com.hartwig.actin.clinical.datamodel.BodyWeight;
+import com.hartwig.actin.clinical.sort.BodyWeightDescendingDateComparator;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Optional;
+
+public class HasMaximumBMI implements EvaluationFunction {
```
kduyvesteyn, line 15: In general the functions have a name referring to "What needs to be true for this function to pass?". 

The current name may suggest the patient needs to have a certain maximum BMI. Maybe a name like "HasLimitedBMI" is more correct?

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+
+import java.time.LocalDate;
+import java.util.Arrays;
+import java.util.Collections;
+
+import static com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation;
+import static org.junit.Assert.assertTrue;
+
+public class HasMaximumBMITest {
```
kduyvesteyn, line 15: Add enter?

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
```diff
@@ -0,0 +1,67 @@
...
+    private final HasMaximumBMI function = new HasMaximumBMI(40);
+    private final LocalDate now = LocalDate.now();
+    private final LocalDate lastYear = now.minusYears(1);
```
kduyvesteyn, line 18: Do you prefer static or non-static here? I am fine either way, and happy to become consistent with your preference. Previously I have made all test constants static.

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
```diff
@@ -0,0 +1,67 @@
...
+                        ImmutableBodyWeight.builder().date(now).value(70).unit("pound").build()
+                )
+        )));
+    }
+
+    @Test
+    public void shouldPassIfLatestWeightIsLessThanWarnThreshold() {
+        Evaluation evaluation = function.evaluate(GeneralTestFactory.withBodyWeights(Arrays.asList(
+                ImmutableBodyWeight.builder().date(now).value(70.57).unit("Kilogram").build(),
+                ImmutableBodyWeight.builder().date(lastYear).value(100.0).unit("Kilogram").build()
```
kduyvesteyn, line 39: Rather than referring to "Kilogram" should you not want to refer to HasMaximumBPI.EXPECTED_UNIT? I imagine you don't care what unit the expected unit is, as long as the weights are passed in the expected unit?

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
```diff
@@ -0,0 +1,67 @@
...
+        )));
+        assertEvaluation(EvaluationResult.PASS, evaluation);
+        assertTrue(evaluation.passSpecificMessages().contains(
```
kduyvesteyn, line 42: I have never tested the exact messages, feels that adds too little for the overhead it causes. 

The assertEvaluation function checks if the expected message types are provided, but not the actual content. 

Do you think we should test for the actual message content?

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
```diff
@@ -0,0 +1,69 @@
...
+
+    static final String EXPECTED_UNIT = "kilogram";
```
kduyvesteyn, line 17: Why not use "Kilogram" here? Or is "kilogram" the typical value we get? 

Since we test ignoring case it does not really matter but still might be cleaner to put here what we generally expect?

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
```diff
@@ -0,0 +1,69 @@
...
+    static final double MIN_EXPECTED_HEIGHT_METRES = 1.5;
+    static final double MAX_EXPECTED_HEIGHT_METRES = 2.0;
```
kduyvesteyn, line 19: Why are these not private variables?

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
jbartletthmf: I included the messages in testing because they're composed in the code and I wanted to verify the final product. We can drop the tests if they're too brittle.

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
jbartletthmf: I used different case here to test case-insensitive matching. We can change the tests to reference the constant if they become too brittle.

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
jbartletthmf: On second thought, I think I prefer these to be non-static and re-initialized for every test execution. The test class doesn't need to be aware of the setup for `HasMaximumBMI`, so it's safest to create it anew for each test case. The time-based values could theoretically be different for each test (this would be more relevant if we were dealing with milliseconds instead of days), so I think it's appropriate to leave them non-static as well.

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
kduyvesteyn: I can see that my naming is not entirely consistent either! 

I like your latest suggestion: `HasBMIUpToLimit`

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
kduyvesteyn: Agreed

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
kduyvesteyn: Sounds good

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
kduyvesteyn: clear. 

Related, if you are not using `EXPECTED_UNIT` in test code you may as well make it private?

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
jbartletthmf: Locale is fixed in #4.

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMI.java
jbartletthmf: Renamed in #4.

PR #4, opened 2023-02-16
ACTIN-19: Indicate height threshold in BMI messages

jbartletthmf:
Unrelated refactor to move rule to vitalfunction after discussion with Nina.
Rename HasMaximumBMI to HasBMIUpToLimit for clarity.
Set locale to ENGLISH for messages to maintain formatting across locales.


pauldwolfe (approved):
Code looks good, maybe when merging can elaborate a little bit on the rational behind the refactor and how it links to the requirement of the ticket? Was a little confusing to me at first.

pauldwolfe:
@jbartletthmf yes i think worth mentioning in the message that you included an unrelated refactor and the rationale for it.

kduyvesteyn (commented):
All changes look good to me

common/src/main/java/com/hartwig/actin/util/ApplicationConfig.java
```diff
@@ -2,13 +2,16 @@
...
 import java.nio.file.Files;
+import java.util.Locale;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.ParseException;
 import org.jetbrains.annotations.NotNull;
 
 public final class ApplicationConfig {
 
+    public static final Locale locale = Locale.ENGLISH;
```
kduyvesteyn, line 13: Isn't It more standard to use capitals for static final parameters? So LOCALE rather than locale

kduyvesteyn (commented):
One small comment about capitalisation

common/src/main/java/com/hartwig/actin/util/ApplicationConfig.java
jbartletthmf: Yes, I'll fix this.

kduyvesteyn (commented):
looks good

PR #5, opened 2023-02-17
ACTIN-16: Add lymph node lesions in clinical JSON and curation

jbartletthmf:
Rename TumorUtil to TumorMetastasisEvaluator for clarity.
Use static EvaluationFunction references in tests as per feedback.
Use Boolean equals comparison to handle nulls and resolve IDE warning.
Restructure CurationModel lesion tests to align them with acceptance criteria.
Increase size of targetedType column to accommodate clinical data.
Replace explicit JSON deserialization with type adapters.

### DIFF
--- a/algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluator.java
+++ b/algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluator.java
@@ -7,9 +7,10 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class TumorUtil {
+public final class TumorMetastasisEvaluator {
+
     @NotNull
-    public static Evaluation evaluateBooleanMetastasis(@Nullable Boolean hasMetastases, @NotNull String metastasisType) {
+    public static Evaluation evaluate(@Nullable Boolean hasMetastases, @NotNull String metastasisType) {
         ImmutableEvaluation.Builder builder = EvaluationFactory.unrecoverable();
         if (hasMetastases == null) {
             builder.result(EvaluationResult.UNDETERMINED)
--- a/algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasBoneMetastasesTest.java
+++ b/algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasBoneMetastasesTest.java
@@ -1,14 +1,13 @@
 package com.hartwig.actin.algo.evaluation.tumor;
 
 import static com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation;
-
 import com.hartwig.actin.algo.datamodel.Evaluation;
 import com.hartwig.actin.algo.datamodel.EvaluationResult;
-
 import org.junit.Test;
 
 public class HasBoneMetastasesTest {
-    private final HasBoneMetastases function = new HasBoneMetastases();
+
+    private static final HasBoneMetastases function = new HasBoneMetastases();
 
     @Test
     public void shouldBeUndeterminedWhenHasBoneLesionsIsNull() {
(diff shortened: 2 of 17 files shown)


algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluator.java
```diff
@@ -7,9 +7,10 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class TumorUtil {
+public final class TumorMetastasisEvaluator {
```
jbartletthmf, line 10: I looked into making this a function that takes in a `MetastasesEvaluationFunction`, but every step in that direction seemed to reintroduce the duplication that I was trying to clean up. Instead, I went with @kduyvesteyn's suggested rename, which should improve readability.

kduyvesteyn: I can see your point. I've thought of alternatives but can't think of a good one either. You could make the function non-static and create an instance every time the parent evaluation function is called but that would be worse even. And the function you end up does not resemble a normal evaluation function anymore either. Fine with this rename

kduyvesteyn, line 10: Consider adding a private constructor? I do consistently to avoid instantiation but also open for another approach if you prefer.

algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasBoneMetastasesTest.java
```diff
@@ -1,14 +1,13 @@
...
-
 import com.hartwig.actin.algo.datamodel.Evaluation;
 import com.hartwig.actin.algo.datamodel.EvaluationResult;
-
 import org.junit.Test;
 
 public class HasBoneMetastasesTest {
-    private final HasBoneMetastases function = new HasBoneMetastases();
+
+    private static final HasBoneMetastases function = new HasBoneMetastases();
```
kduyvesteyn, line 10: Consider making FUNCTION (static final objects in capitalised letters)?

clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationModel.java
```diff
@@ -159,40 +159,47 @@ public TumorDetails overrideKnownLesionLocations(@NotNull TumorDetails tumorDeta
 
         ImmutableTumorDetails.Builder builder = ImmutableTumorDetails.builder().from(tumorDetails);
         if (matches.contains(LesionLocationCategory.BRAIN)) {
-            if (tumorDetails.hasBrainLesions() != null && !tumorDetails.hasBrainLesions()) {
+            if (Boolean.FALSE.equals(tumorDetails.hasBrainLesions())) {
```
kduyvesteyn, line 162: I like it!

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationModelTest.java
```diff
@@ -48,15 +45,15 @@ public class CurationModelTest {
 
     private static final String CURATION_DIRECTORY = Resources.getResource("curation").getPath();
 
+    private CurationModel model = TestCurationFactory.createProperTestCurationModel();
```
kduyvesteyn, line 48: Make final static?

jbartletthmf: I'll make it final, but as above I think it's better to remain non-static.

algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluator.java
kduyvesteyn: It's just unnecessary, and I've done this consistently so far across ACTIN code. Other than that no strong reason.

If you and Paul feel it adds too little I am fine to drop it. Let me know!

algo/src/test/java/com/hartwig/actin/algo/evaluation/tumor/HasBoneMetastasesTest.java
kduyvesteyn: Agreed

algo/src/main/java/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluator.java
kduyvesteyn: Fine!

PR #6, opened 2023-02-18
Replace explicit JSON deserialization with type adapters

jbartletthmf:
@kduyvesteyn I'm not sure if this is exactly what you were looking for as a deserialization improvement, but it works using types without explicitly extracting each field. I left it out of the feature branch for now but I'll merge to that branch if you'd like to include it.

### DIFF
--- a/common/src/test/resources/clinical/patient.clinical.json
+++ b/common/src/test/resources/clinical/patient.clinical.json
@@ -120,7 +120,7 @@
   "complications": [
     {
       "name": "Nausea",
-      "categories": "Other",
+      "categories": ["Other"],
       "year": 2020,
       "month": null
     }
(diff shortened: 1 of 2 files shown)


common/src/test/resources/clinical/patient.clinical.json
```diff
@@ -120,7 +120,7 @@
   "complications": [
     {
       "name": "Nausea",
-      "categories": "Other",
+      "categories": ["Other"],
```
kduyvesteyn, line 123: The formatting is inconsistent, it should be something like 
[
   "Other"
]

I generally use auto-formatting in IntelliJ for json files, works acceptably well in my opinion.

kduyvesteyn (commented):
The new deserialisation is a lot better!

PR #7, opened 2023-02-21
Actin 8

kduyvesteyn:
Some thoughts:
 - Rather than having a post-validation steps, maybe we should integrate the checking (see CurationUtil.toDOIDs), such that we check when they initially populate the datamodel, rather than after-the-fact?
 - There are some inconsistencies in the curation configs and some could be cleaner. E.g. NonOncologicalHistoryConfig has a curated Object which isn't really nice for downstream users (you have to know the types of objects that the factory could feed in here). Could maybe improve...

### DIFF
--- a/algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
+++ b/algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
@@ -1,16 +1,19 @@
 package com.hartwig.actin.algo.evaluation.general;
 
-import com.hartwig.actin.algo.datamodel.Evaluation;
-import com.hartwig.actin.algo.datamodel.EvaluationResult;
-import com.hartwig.actin.clinical.datamodel.ImmutableBodyWeight;
-import org.junit.Test;
+import static com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation;
+
+import static org.junit.Assert.assertTrue;
 
 import java.time.LocalDate;
 import java.util.Arrays;
 import java.util.Collections;
 
-import static com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation;
-import static org.junit.Assert.assertTrue;
+import com.hartwig.actin.algo.datamodel.Evaluation;
+import com.hartwig.actin.algo.datamodel.EvaluationResult;
+import com.hartwig.actin.clinical.datamodel.ImmutableBodyWeight;
+
+import org.junit.Ignore;
+import org.junit.Test;
 
 public class HasMaximumBMITest {
     private final HasMaximumBMI function = new HasMaximumBMI(40);
@@ -33,6 +36,7 @@ public class HasMaximumBMITest {
     }
 
     @Test
+    @Ignore
     public void shouldPassIfLatestWeightIsLessThanWarnThreshold() {
         Evaluation evaluation = function.evaluate(GeneralTestFactory.withBodyWeights(Arrays.asList(
                 ImmutableBodyWeight.builder().date(now).value(70.57).unit("Kilogram").build(),
@@ -44,6 +48,7 @@ public class HasMaximumBMITest {
     }
 
     @Test
+    @Ignore
     public void shouldFailIfLatestWeightIsGreaterThanFailThreshold() {
         Evaluation evaluation = function.evaluate(GeneralTestFactory.withBodyWeights(Arrays.asList(
                 ImmutableBodyWeight.builder().date(now).value(180.32).unit("Kilogram").build(),
@@ -55,6 +60,7 @@ public class HasMaximumBMITest {
     }
 
     @Test
+    @Ignore
     public void shouldWarnIfLatestWeightIsGreaterThanWarnThreshold() {
         Evaluation evaluation = function.evaluate(GeneralTestFactory.withBodyWeights(Arrays.asList(
                 ImmutableBodyWeight.builder().date(now).value(100.99).unit("Kilogram").build(),
--- a/clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
+++ b/clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
@@ -1,0 +1,119 @@
+package com.hartwig.actin.clinical.curation;
+
+import java.util.List;
+import java.util.Set;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.hartwig.actin.clinical.curation.config.IntoleranceConfig;
+import com.hartwig.actin.clinical.curation.config.NonOncologicalHistoryConfig;
+import com.hartwig.actin.clinical.curation.config.PrimaryTumorConfig;
+import com.hartwig.actin.clinical.curation.config.SecondPrimaryConfig;
+import com.hartwig.actin.clinical.datamodel.PriorOtherCondition;
+import com.hartwig.actin.doid.DoidModel;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+
+class CurationDatabaseValidator {
+
+    private static final Logger LOGGER = LogManager.getLogger(CurationDatabaseValidator.class);
+
+    static final String GENERIC_PARENT_DOID = "4"; // "disease"
+    static final String CANCER_PARENT_DOID = "14566"; // "disease of cellular proliferation"
+
+    @NotNull
+    private final DoidModel doidModel;
+
+    public CurationDatabaseValidator(@NotNull final DoidModel doidModel) {
+        this.doidModel = doidModel;
+    }
+
+    public void validate(@NotNull CurationDatabase database) {
+        LOGGER.info("Running curation database validation");
+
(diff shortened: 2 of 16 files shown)


clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        return allValid;
+    }
+
+    @VisibleForTesting
+    static boolean validateNonOncologicalHistoryConfigs(@NotNull List<NonOncologicalHistoryConfig> nonOncologicalHistoryConfigs,
+            @NotNull DoidModel doidModel) {
+        boolean allValid = true;
+        for (NonOncologicalHistoryConfig nonOncologicalHistoryConfig : nonOncologicalHistoryConfigs) {
+            Object curated = nonOncologicalHistoryConfig.curated();
```
jbartletthmf, line 80: I agree with your comment: the fact that this returns an `Object` makes it hard to work with. It would be better if all objects that could be returned here implemented an interface or abstract class such as `Curated` so we would have some idea of what to expect. The interface could include some useful methods like `isPriorOtherCondition()` to facilitate filtering here. That would allow something like:
```
        return nonOncologicalHistoryConfigs.stream()
                .noneMatch(nonOncologicalHistoryConfig -> Optional.ofNullable(nonOncologicalHistoryConfig.curated())
                        .filter(Curated::isPriorOtherCondition)
                        .map(curated -> (PriorOtherCondition) curated)
                        .map(PriorOtherCondition::doids)
                        .map(doids -> !hasValidDoids(doids, doidModel, GENERIC_PARENT_DOID, PRIOR_OTHER_CONDITION,
                                nonOncologicalHistoryConfig.input()))
                        .orElse(false));
```
In the above, I suggest modifying `hasValidDoids` to take some additional parameters so it can handle logging.

clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        return allValid;
+    }
+
+    @VisibleForTesting
+    static boolean validateSecondPrimaryConfigs(@NotNull List<SecondPrimaryConfig> secondPrimaryConfigs,
+            @NotNull DoidModel doidModel) {
+        boolean allValid = true;
+        for (SecondPrimaryConfig secondPrimaryConfig : secondPrimaryConfigs) {
+            if (!hasValidDoids(secondPrimaryConfig.curated().doids(), doidModel, CANCER_PARENT_DOID)) {
```
jbartletthmf, line 65: `SecondPrimaryConfig::curated` is nullable, so we should use an Optional here. We could use the same pattern I use below for the history without the filter and cast steps.

clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
```diff
@@ -0,0 +1,119 @@
...
+        if (doids.isEmpty()) {
+            return false;
+        }
+
+        for (String doid : doids) {
+            if (!doidModel.doidWithParents(doid).contains(expectedParentDoid)) {
+                return false;
+            }
+        }
+        return true;
```
jbartletthmf, line 117: Alternatively:
```suggestion
        return doids.stream().allMatch(doid -> doidModel.doidWithParents(doid).contains(expectedParentDoid));
```

jbartletthmf: Though we'd have to write the log message before returning if you take my other suggestion.

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    public void doesNotCrashOnTestCurationDatabase() {
+        CurationDatabaseValidator validator = TestCurationFactory.createMinimalTestCurationDatabaseValidator();
+
+        validator.validate(TestCurationFactory.createTestCurationDatabase());
+    }
+
+    @Test
+    public void canIdentifyInvalidPrimaryTumorConfigs() {
```
jbartletthmf, line 30: I usually try to include "should" at the start of the test name instead of "can" or "does". Do we have/want a convention for test names?

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        PrimaryTumorConfig valid = TestCurationConfigFactory.primaryTumorConfigBuilder().addDoids("child").build();
+        assertTrue(CurationDatabaseValidator.validatePrimaryTumorConfigs(Lists.newArrayList(valid), doidModel));
+
+        PrimaryTumorConfig invalid = TestCurationConfigFactory.primaryTumorConfigBuilder().addDoids("invalid").build();
+        assertFalse(CurationDatabaseValidator.validatePrimaryTumorConfigs(Lists.newArrayList(invalid), doidModel));
+    }
+
+    @Test
+    public void canIdentifyInvalidSecondPrimaryConfigs() {
```
jbartletthmf, line 41: Should we add a test case where `priorSecondPrimary.curated()` is null?

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Test
+    public void canIdentifyInvalidPrimaryTumorConfigs() {
+        DoidModel doidModel = TestDoidModelFactory.createWithOneParentChild(CurationDatabaseValidator.CANCER_PARENT_DOID, "child");
+
+        PrimaryTumorConfig valid = TestCurationConfigFactory.primaryTumorConfigBuilder().addDoids("child").build();
+        assertTrue(CurationDatabaseValidator.validatePrimaryTumorConfigs(Lists.newArrayList(valid), doidModel));
+
+        PrimaryTumorConfig invalid = TestCurationConfigFactory.primaryTumorConfigBuilder().addDoids("invalid").build();
```
jbartletthmf, line 36: I think it would be more effective to validate a list containing both a valid DOID and an invalid one, like `["child", "invalid"]`, to verify that a single invalid entry results in a validation issue.

algo/src/test/java/com/hartwig/actin/algo/evaluation/general/HasMaximumBMITest.java
```diff
@@ -33,6 +36,7 @@ public void shouldBeUndeterminedWhenNoBodyWeightsProvidedWithExpectedUnit() {
     }
 
     @Test
+    @Ignore
```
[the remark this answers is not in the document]
jbartletthmf: The fix is now merged to `master`.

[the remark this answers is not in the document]
kduyvesteyn: Updated

clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
kduyvesteyn: I have refactored NonOncologicalHistoryConfig to contain both an optional LVEF (double) and an optional PriorOtherCondition. It's not optimal yet but at least better than Object. The current datamodel is somewhat comparable to eg ECGConfig.

I feel a more correct route would be to model every config "from the perspective of what it could become" rather than "from the perspective of where it came from". E.g. we should maintain a list of inputs with mappings of everything that should be interpreted as a PriorOtherCondition. This way every config would have an input, an ignore and one optional curated "something". Currently we modeled the config to the fields where they originate from in the questionnaire.

clinical/src/main/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidator.java
kduyvesteyn: Have taken over the streaming solution, but not sure about the logging. I don't like passing params into more deep code just for the sake of generating a log message. But do have plans to approach the whole ticket somewhat different so not too relevant.

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
kduyvesteyn: There is no convention for test names other than that they should represent what the test tests for. 

I can see why "should" is better in this case, will update

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
kduyvesteyn: Agreed, good catch. Missed this scenario (good argument for Optional!)

clinical/src/test/java/com/hartwig/actin/clinical/curation/CurationDatabaseValidatorTest.java
kduyvesteyn: Agreed, updated

PR #8, opened 2023-02-21
ACTIN-11: Add footnote for reportable, potentially subclonal variants

jbartletthmf:
Explicitly check variants for reportability and potential for subclonal
to avoid false positives.

This PR follows from discussion in #1.


kduyvesteyn (commented):
I like this solution! 

Also a nice demonstration of streams over loops

PR #9, opened 2023-02-23
ACTIN-12: Warn when WHO status equals limit with certain complications

jbartletthmf:
Refactor HasComplicationOfCategory to use new method.
Reorganize tests for clarity.


algo/src/main/java/com/hartwig/actin/algo/evaluation/complication/PatternMatcher.java
```diff
@@ -1,11 +1,16 @@
...
-
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import com.hartwig.actin.PatientRecord;
+import com.hartwig.actin.clinical.datamodel.Complication;
+import com.hartwig.actin.clinical.datamodel.ImmutableComplication;
 import org.jetbrains.annotations.NotNull;
 
-final class PatternMatcher {
+public final class PatternMatcher {
```
kduyvesteyn, line 13: A few comments: 
 - I would put the specific complication functions in a separate class with a more concrete name. PatternMatcher previously was independent from a specific datamodel object, and package-private. Maybe "ComplicationFunctions" would be a better name to hold the new functions?

algo/src/main/java/com/hartwig/actin/algo/evaluation/complication/PatternMatcher.java
```diff
@@ -30,4 +35,37 @@ public static boolean isMatch(@NotNull String term, @NotNull Set<List<String>> p
 
         return false;
     }
+
+    public static Set<String> findComplicationNamesMatchingCategories(PatientRecord record,
```
kduyvesteyn, line 39: I would suggest a name that makes it more clear that "any" category match is allowed. The name of the function could be interpreted as "all". Maybe "findComplicationNamesMatchingAtLeastOneCategory". Or potentially "Any" instead of "AtLeastOne" to be less verbose?

algo/src/main/java/com/hartwig/actin/algo/evaluation/complication/PatternMatcher.java
```diff
@@ -30,4 +35,37 @@ public static boolean isMatch(@NotNull String term, @NotNull Set<List<String>> p
...
+                                                                      List<String> categorySearchTerms) {
+        return findComplicationsMatchingCategories(record, categorySearchTerms).orElse(Stream.empty())
+                .map(Complication::name)
+                .collect(Collectors.toSet());
+    }
+
+    public static Set<String> findComplicationCategoriesMatchingCategories(PatientRecord record,
```
kduyvesteyn, line 46: Naming could be improved, similar to above comment

algo/src/main/java/com/hartwig/actin/algo/evaluation/complication/PatternMatcher.java
```diff
@@ -30,4 +35,37 @@ public static boolean isMatch(@NotNull String term, @NotNull Set<List<String>> p
...
+                                                                            List<String> categorySearchTerms) {
+        return findComplicationsMatchingCategories(record, categorySearchTerms).orElse(Stream.empty())
+                .flatMap(complication -> complication.categories().stream())
+                .collect(Collectors.toSet());
+    }
+
+    private static Optional<Stream<ImmutableComplication>> findComplicationsMatchingCategories(PatientRecord record,
```
kduyvesteyn, line 53: Naming could be improved, similar to above comment

(side note: really nice to see such examples of streams!)

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.java
```diff
@@ -29,7 +33,17 @@ public Evaluation evaluate(@NotNull PatientRecord record) {
                     .build();
         }
 
-        if (who <= maximumWHO) {
+        Set<String> warningComplicationCategories = PatternMatcher.findComplicationCategoriesMatchingCategories(record,
+                COMPLICATION_CATEGORIES_AFFECTING_WHO_STATUS);
```
kduyvesteyn, line 37: Might be cleaner to move COMPLICATION_CATEGORIES_AFFECTING_WHO_STATUS in a separate class (and perhaps its own wrapper function)? There currently is no "slippage" of properties from one evaluation function implementation to another.

algo/src/main/java/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.java
```diff
@@ -29,7 +32,17 @@ public Evaluation evaluate(@NotNull PatientRecord record) {
...
+
+        if (who == maximumWHO && !warningComplicationCategories.isEmpty()) {
+            return EvaluationFactory.unrecoverable()
+                    .result(EvaluationResult.WARN)
+                    .addWarnSpecificMessages(
+                            "Patient WHO status " + who + " equals maximum but patient has complication categories of concern: "
+                                    + Format.concat(warningComplicationCategories))
+                    .addWarnGeneralMessages("WHO adequate but has " + Format.concat(warningComplicationCategories))
+                    .build();
+        } else if (who <= maximumWHO) {
```
kduyvesteyn, line 45: I generally try to retain ordering "from good to bad" in evaluations, something like:
if (A) -> PASS, else if (B) -> WARN -> else FAIL

In this situation I think I would create a nested if statement: top one is "if who <= maximumWHO", and then inside that if statement evaluate complications, and do "if no dubious complications -> PASS, else WARN

jbartletthmf: That ordering makes sense to me. I dealt with this WARN condition first here because it's a little harder to represent the PASS conditions on their own. For example, in a nested if statement, the first condition would have to be:
```
if (noDubiousComplications || who < maximumWHO) {
```
which means an additional check on `who`. Alternatively, we could handle `who < maximumWHO` in the first top-level condition, but then we have to factor out the duplicate `EvaluationResult` building. Would you still prefer to change this code to use the preferred ordering?

kduyvesteyn: Now I am doubting! I wouldn't consider the additional check on who to be bothersome, but more the fact that you need to invert the if "logical" if statement. We would indeed end up with your if statement and not sure it would make the code more readable. 

If you have no strong thoughts on this then maybe just let this be. I am pretty sure there are other rules that do not strictly follow PASS->WARN->FAIL, though it could still serve as our principle

PR #10, opened 2023-02-24
ACTIN-20: Show cohorts without slots at bottom of table in grey

jbartletthmf:
Refactor EvaluatedTrialFactory to make use of streams

### DIFF
--- a/report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
+++ b/report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
@@ -13,6 +13,11 @@ public class EvaluatedTrialComparator implements Comparator<EvaluatedTrial> {
 
     @Override
     public int compare(@NotNull EvaluatedTrial trial1, @NotNull EvaluatedTrial trial2) {
+        int hasSlotsAvailableCompare = Boolean.compare(!trial1.hasSlotsAvailable(), !trial2.hasSlotsAvailable());
+        if (hasSlotsAvailableCompare != 0) {
+            return hasSlotsAvailableCompare;
+        }
+
         int hasMolecularCompare = Boolean.compare(trial1.molecularEvents().isEmpty(), trial2.molecularEvents().isEmpty());
         if (hasMolecularCompare != 0) {
             return hasMolecularCompare;
--- a/report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
+++ b/report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
@@ -11,31 +11,6 @@ public class EvaluatedTrialComparator implements Comparator<EvaluatedTrial> {
 
     private static final String COMBINATION_COHORT_IDENTIFIER = "+";
 
-    @Override
-    public int compare(@NotNull EvaluatedTrial trial1, @NotNull EvaluatedTrial trial2) {
-        int hasSlotsAvailableCompare = Boolean.compare(!trial1.hasSlotsAvailable(), !trial2.hasSlotsAvailable());
-        if (hasSlotsAvailableCompare != 0) {
-            return hasSlotsAvailableCompare;
-        }
-
-        int hasMolecularCompare = Boolean.compare(trial1.molecularEvents().isEmpty(), trial2.molecularEvents().isEmpty());
-        if (hasMolecularCompare != 0) {
-            return hasMolecularCompare;
-        }
-
-        int trialCompare = trial1.trialId().compareTo(trial2.trialId());
-        if (trialCompare != 0) {
-            return trialCompare;
-        }
-
-        int cohortCompare = compareCohorts(!trial1.molecularEvents().isEmpty(), trial1.cohort(), trial2.cohort());
-        if (cohortCompare != 0) {
-            return cohortCompare;
-        }
-
-        return compareSets(trial1.molecularEvents(), trial2.molecularEvents());
-    }
-
     private static int compareCohorts(boolean hasMolecular, @Nullable String cohort1, @Nullable String cohort2) {
         if (cohort1 == null) {
             return cohort2 != null ? -1 : 0;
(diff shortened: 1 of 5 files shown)


report/src/test/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparatorTest.java
```diff
@@ -18,30 +17,29 @@ public class EvaluatedTrialComparatorTest {
...
+                trials.get(1), trials.get(6), trials.get(0), trials.get(3), trials.get(5));
+        trialList.sort(new EvaluatedTrialComparator());
+        Iterator<EvaluatedTrial> trialIterator = trialList.iterator();
+
+        trials.forEach(trial -> assertEquals(trial, trialIterator.next()));
     }
 
     @NotNull
-    private static EvaluatedTrial create(@NotNull String trialId, @Nullable String cohort, String... molecularEvents) {
+    private static EvaluatedTrial create(@NotNull String trialId, @Nullable String cohort, Boolean hasSlotsAvailable,
```
kduyvesteyn, line 41: hasSlotsAvailable can be a boolean rather than Boolean

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/EligibleActinTrialsGenerator.java
```diff
@@ -110,15 +110,17 @@ public Table contents() {
...
-            table.addCell(Cells.createContent(trial.trialId() + addon));
-            table.addCell(Cells.createContent(trial.acronym()));
-            table.addCell(Cells.createContent(trial.cohort() != null ? trial.cohort() : Strings.EMPTY));
-            table.addCell(Cells.createContent(concat(trial.molecularEvents())));
-            table.addCell(Cells.createContent(concat(trial.warnings())));
+            table.addCell(Cells.createContent(trial.trialId() + addon, greyText));
+            table.addCell(Cells.createContent(trial.acronym(), greyText));
+            table.addCell(Cells.createContent(trial.cohort() != null ? trial.cohort() : Strings.EMPTY, greyText));
+            table.addCell(Cells.createContent(concat(trial.molecularEvents()), greyText));
+            table.addCell(Cells.createContent(concat(trial.warnings()), greyText));
```
kduyvesteyn, line 123: I would consider replacing with a specific function createGreyTextContent in Cells. To prevent duplication inside this function you could have a private function in here based on the boolean?

kduyvesteyn (commented):
In general your code formatting does not comply with existing scheme.

To be honest there are some formatting choices in your code that I prefer over our current scheme. Eg the auto-formatting will make the EvaluatedTrialComparatorTest worse.

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/EligibleActinTrialsGenerator.java
jbartletthmf: Done using a stream.

jbartletthmf (commented):
I imported the coding style and reformatted accordingly.

report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
```diff
@@ -11,26 +11,6 @@ public class EvaluatedTrialComparator implements Comparator<EvaluatedTrial> {
...
-        }
-
-        int cohortCompare = compareCohorts(!trial1.molecularEvents().isEmpty(), trial1.cohort(), trial2.cohort());
-        if (cohortCompare != 0) {
-            return cohortCompare;
-        }
-
-        return compareSets(trial1.molecularEvents(), trial2.molecularEvents());
-    }
-
```
kduyvesteyn, line 33: Currently public functions are always at the top of the class, where private static follow towards the bottom. 

Is there a specific reason why you swapped the ordering here?

report/src/main/java/com/hartwig/actin/report/pdf/util/Cells.java
```diff
@@ -118,6 +118,13 @@ public static Cell createContent(@NotNull String text) {
...
     }
 
+    @NotNull
+    public static Cell createContentGrey(@NotNull String text) {
+        Cell cell = createContent(text);
+        cell.setFontColor(Styles.PALETTE_MID_GREY);
+        return cell;
+    }
+
     @NotNull
```
kduyvesteyn, line 128: Still don't like this name. All functions in Cells are named after their "functional usage". Eg "createContentWarn" could have been called "createContentOrange" but then users wouldn't know when to use this function consistently.

I think rather than use createContentGrey I'd go for something like createContentDeamphasize, or perhaps there is a better word to describe what we try to achieve with greying out certain content?

report/src/test/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparatorTest.java
```diff
@@ -54,7 +33,34 @@ private static EvaluatedTrial create(@NotNull String trialId, @Nullable String c
...
+                trials.get(1),
+                trials.get(6),
+                trials.get(0),
+                trials.get(3),
+                trials.get(5));
+        trialList.sort(new EvaluatedTrialComparator());
+        Iterator<EvaluatedTrial> trialIterator = trialList.iterator();
+
+        trials.forEach(trial -> assertEquals(trial, trialIterator.next()));
+    }
```
kduyvesteyn, line 65: Same as above, private static function appears before public

report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
jbartletthmf: This is from running `Reformat` in IntelliJ after importing the code style settings. It seems to group static functions at the top before considering public/private. Do you get a different result in your IDE? I'll take a look at my settings.

report/src/main/java/com/hartwig/actin/report/pdf/util/Cells.java
jbartletthmf: That makes sense. I can fix this in #12.

report/src/main/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparator.java
kduyvesteyn: The reason seems to be that our scheme is set up for `Reformat` only and not for `Rearrange`

report/src/test/java/com/hartwig/actin/report/interpretation/EvaluatedTrialComparatorTest.java
jbartletthmf: The order of functions is fixed in #12.

PR #12, opened 2023-03-02
ACTIN-21: Improve summary, molecular and clinical chapters, and reduce file size

jbartletthmf:
- Improve summary chapter to include most relevant information.
- Group cohorts by trials in listings and merge trial ID and acronym
columns.
- Include clinical summary in clinical details chapter.
- Show lab values chronologically from left to right.
- Add disclaimer to footer on all pages.
- Replace skipMatchingTrialDetails flag with enableExtendedMode,
and use it to show/hide closed eligible cohorts with no molecular
events and closed ineligible cohorts.
- Show closed, ineligible cohorts in grey in long version.
- Reduce file size by setting iText PDF compression parameters and,
more importantly, by reusing Fonts instead of recreating for each
cell. Fonts cannot be shared between files, so we re-initialize them
at the start of report writing. It would be better to make them
immutable and access them non-statically, but this would require more
extensive refactoring.
- Rename EvaluatedTrial to EvaluatedCohort for accuracy.
- Add toString method to Virus to print common abbreviations.


report/src/main/java/com/hartwig/actin/report/pdf/ReportWriter.java
```diff
@@ -27,6 +30,7 @@
 
 public class ReportWriter {
 
+    public static final float STANDARD_KEY_WIDTH = 210;
```
kduyvesteyn, line 33: This property does not really belong here. I think "Formats" may be a better location?

report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/LabResultsGenerator.java
```diff
@@ -64,7 +88,12 @@ public String title() {
     @NotNull
     @Override
     public Table contents() {
-        Set<LocalDate> dates = selectDates(labInterpretation.allDates());
+        List<LocalDate> dates = labInterpretation.allDates()
```
kduyvesteyn, line 91: The entries in this collection have to be unique. Might be enforced already by providing class, but wouldn't still be better to use a Set rather than a List here?

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/MolecularCharacteristicsGenerator.java
```diff
@@ -142,11 +172,11 @@ private Cell createTMBStatusCell(@Nullable Boolean hasHighTumorMutationalBurden,
     }
 
     @NotNull
-    private Cell createMSStabilityCell(@Nullable Boolean isMicrosatelliteUnstable) {
```
kduyvesteyn, line 145: Why package-private rather than private?

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/MolecularCharacteristicsGenerator.java
```diff
@@ -162,11 +192,13 @@ private Cell createMSStabilityCell(@Nullable Boolean isMicrosatelliteUnstable) {
     }
 
     @NotNull
-    private Cell createHRStatusCell(@Nullable Boolean isHomologousRepairDeficient) {
```
kduyvesteyn, line 165: Why package-private rather than private?

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -1,148 +1,186 @@
...
+import com.itextpdf.layout.element.Paragraph;
 import com.itextpdf.layout.element.Table;
+import com.itextpdf.layout.element.Text;
 
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 public class RecentMolecularSummaryGenerator implements TableGenerator {
 
+    private static final double PURITY_WARN_THRESHOLD = 0.20;
```
kduyvesteyn, line 33: It is more correct to use molecularRecord.hasSufficientQuality

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -1,148 +1,186 @@
...
+        Set<DriverLikelihood> allowedLikelihoods = Set.of(DriverLikelihood.LOW, DriverLikelihood.MEDIUM);
+
+        Set<String> eventsWithActinTrials = trials.stream()
+                .filter(EvaluatedTrial::isPotentiallyEligible)
+                .filter(EvaluatedTrial::isOpen)
+                .flatMap(trial -> trial.molecularEvents().stream())
+                .collect(Collectors.toSet());
+
+        String genes = Stream.concat(molecular.drivers().variants().stream(), molecular.drivers().disruptions().stream())
+                .filter(driver -> driver.driverLikelihood() != null)
```
kduyvesteyn, line 178: I added this line since without this the code would give a NPE on production samples. Not sure why though? "if (set.contains(null))" is valid code I thought?

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/EligibleActinTrialsGenerator.java
```diff
@@ -56,7 +56,7 @@ public static EligibleActinTrialsGenerator forClosedTrials(@NotNull List<Evaluat
             }
         }
 
-        String title = String.format("%s trials that are closed or blacklisted but considered eligible (%s)",
+        String title = String.format("%s trials and cohorts that are considered ineligible (%s)",
```
kduyvesteyn, line 59: This looks like an accidental change, since we still display trials that are closed or blacklisted but considered eligible?

report/src/main/java/com/hartwig/actin/report/pdf/ReportWriter.java
```diff
@@ -42,7 +46,9 @@ public void write(@NotNull Report report) throws IOException {
         write(report, false);
     }
 
-    public void write(@NotNull Report report, boolean skipTrialMatchingDetails) throws IOException {
+    public synchronized void write(@NotNull Report report, boolean skipTrialMatchingDetails) throws IOException {
+        LOGGER.info("Initializing output styles");
+        Styles.initialize();
```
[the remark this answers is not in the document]
jbartletthmf: Yes, that's the idea. However, it's only enforced within each instance of a class, so it doesn't help us if another ReportWriter is created elsewhere. I think the real solution here is to always access Fonts (and therefore Styles, Cells, etc.) through instance methods and make the Fonts final, but that's a bigger change that I didn't think should delay the report improvements. We can drop `synchronized` as long as we're only running one report at a time.

report/src/main/java/com/hartwig/actin/report/pdf/components/Footer.java
```diff
@@ -26,37 +26,43 @@ public void render(@NotNull PdfPage page) {
         PdfCanvas canvas = new PdfCanvas(page.getLastContentStream(), page.getResources(), page.getDocument());
 
         int pageNumber = page.getDocument().getPageNumber(page);
-        PdfFormXObject pageNumberTemplate = new PdfFormXObject(new Rectangle(0, 0, 200, 20));
+        PdfFormXObject pageNumberTemplate = new PdfFormXObject(new Rectangle(0, 0, 450, 20));
```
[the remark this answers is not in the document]
jbartletthmf: I missed this one, thanks.

report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/LabResultsGenerator.java
jbartletthmf: They're provided as a set, so uniqueness is guaranteed. We could enforce it in the stream to be safe though.

I wanted to use a list to preserve order. With a set, there's no guarantee that we'll use the same date order for each set of measurements.

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/MolecularCharacteristicsGenerator.java
jbartletthmf: The package-private cells are used in the molecular summary as well.

report/src/main/java/com/hartwig/actin/report/pdf/ReportWriter.java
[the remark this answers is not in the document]
kduyvesteyn: Clear, and all makes sense (including keeping `synchronized` in)

report/src/main/java/com/hartwig/actin/report/pdf/tables/clinical/LabResultsGenerator.java
kduyvesteyn: Clear and wasn't aware of that!

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/MolecularCharacteristicsGenerator.java
kduyvesteyn: Clear, missed that

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
kduyvesteyn: I would prefer to use explicit null check rather than guava. The explicit null checks we can clean up as we move towards optionals, while guava may hide unnoticed for years :)

kduyvesteyn: One functional comment about this code, I think you can add every type of driver to this stream, not just the variants and disruptions. Or is there a specific reason why you add only these 2?

jbartletthmf: I just added those 2 based on a literal interpretation of Nina's comment:
> i.e. medium/low driver mutation or disruption

I don't know what the clinician is trying to see here, but I can add other driver types as well if there's value. This implementation is limited to Driver implementations that also implement `GeneAlteration` to retrieve the specific gene. That means we could easily add copy number variations and homozygous disruptions, but would need to return a different string if we wanted to include fusions or viruses. Should I include all gene alterations?

Regarding the non-drivers, it sounds like what we actually want to include here is all entries with likelihood != HIGH. That takes care of the null check too :)

kduyvesteyn: It would be cleanest to include all gene alterations and fusions. LOW driver fusions could in theory be actionable just like LOW driver variants. 

And yes, != HIGH should do the trick!

jbartletthmf: I just discussed this with Nina and we agreed on some improvements. The new reported groups would include:
- Genes with high driver mutation: variant - HIGH
- Amplified genes: copy number gains - all likelihoods (should always be HIGH)
- Deleted genes: copy number losses - all likelihoods (should always be HIGH)
- Homozygously disrupted genes - all likelihoods
- Gene fusions - all likelihoods
- Virus detection - all likelihoods
- Actionable genes with medium/low driver: variant with likelihood != HIGH and disruptions (should never be HIGH)

This should include all drivers of interest. Do you agree?

kduyvesteyn: The report should not make assumptions about the fact that disruptions are never HIGH drivers. 

Also, there may be amplified genes with driverLikelihood == null, would we want to display those? 

I personally think your original proposal with my proposed adjustments is cleaner. It makes no assumptions about what events could have what driver likelihoods and also is 100% exhaustive. We never miss a HIGH or actionable driver of any kind, without making assumptions about actionability.

kduyvesteyn: Nina is sitting next to me and understood the route that we are on, but noted that "actionable genes" should be renamed to "actionable events", since viruses/fusions are not genes. Also we may want to display the events rather the genes, would need to test if this doesn't lead to super long lists of events for some samples.

jbartletthmf: I agree that we shouldn't assume any likelihoods and risk missing useful data. If we're going to continue limiting each grouping to high likelihood, should we change the row titles to add " with high driver likelihood"?

Actionable events (vs. genes) is a good change if we want to extend this beyond gene alterations, though the event representations of the gene alterations are definitely longer. Do we want to include viruses here and limit the earlier virus row to HIGH? Should we limit the fusions and homozygous disruptions rows to HIGH as well? Should we add a row for HIGH disruptions?

As a space-saving measure, we could consider hiding some of these rows when they're empty if they aren't usually relevant.

kduyvesteyn: I think we can leave the rows as they are. Very strictly speaking we assume that disruptions are never HIGH but I think that's fine. Conditionally showing rows is dubious since "none" is a meaningful result, so better always show the same columns. I can do a review of the final column names once the logic is completed. I think its quite subtle, and we generally don't want to push the concept of driver likelihood to end users

jbartletthmf: I added a row for HIGH disruptions but I can take it out if it's just noise. Rather than making an assumption about likelihood, I could include all disruptions in the actionable events field and only apply likelihood filtering to the other drivers. Would that be the ideal solution?

I agree it's better to show "None" when we have that information.

kduyvesteyn: "Would that be the ideal solution" -> Yes it sounds good to me!

common/src/main/java/com/hartwig/actin/molecular/datamodel/driver/VirusType.java
```diff
@@ -6,5 +6,22 @@ public enum VirusType {
...
+                return "EBV";
+            case HUMAN_PAPILLOMA_VIRUS:
+                return "HPV";
+            case HEPATITIS_B_VIRUS:
+                return "HBV";
+            case HUMAN_HERPES_VIRUS_8:
+                return "HHV8";
+            default:
+                return "Other";
+        }
```
kduyvesteyn, line 25: Doesn't this potentially impact serialisation / deserialization? I generally use "display" for the name of the function that generates a string that is meant for display to a human being

common/src/test/java/com/hartwig/actin/algo/datamodel/TestTreatmentMatchFactory.java
```diff
@@ -138,6 +138,11 @@ private static List<CohortMatch> createTestCohortsTrial1() {
                 .isPotentiallyEligible(false)
                 .evaluations(createTestCohortEvaluationsTrial1CohortC())
                 .build());
+        cohorts.add(ImmutableCohortMatch.builder()
+                .metadata(createTestMetadata("D", false, false, true))
+                .isPotentiallyEligible(true)
+                .evaluations(createTestCohortEvaluationsTrial1CohortC())
```
kduyvesteyn, line 144: You create evaluations belonging to cohort C. Not clean!

report/src/main/java/com/hartwig/actin/report/pdf/components/Footer.java
```diff
@@ -26,37 +26,43 @@ public void render(@NotNull PdfPage page) {
...
+        void renderFooter(@NotNull PdfDocument document, int totalPageCount) {
             String displayString = pageNumber + "/" + totalPageCount;
 
             Canvas canvas = new Canvas(template, document);
             Paragraph pageNumberParagraph = new Paragraph().add(displayString).addStyle(Styles.pageNumberStyle());
             canvas.showTextAligned(pageNumberParagraph, 0, 0, TextAlignment.LEFT);
+
+            String disclaimer =
+                    "All results and data described in this report are for research use only and have not been generated using a "
+                            + "clinically validated and controlled procedure. These results should not be used for clinical decision making.";
```
kduyvesteyn, line 63: Consider rephrase, see mail

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -36,113 +37,162 @@ public class RecentMolecularSummaryGenerator implements TableGenerator {
...
-            if (trial.isPotentiallyEligible() && trial.isOpen()) {
-                molecularEvents.addAll(trial.molecularEvents());
+    private Cell biopsySummary() {
+        String biopsyLocation = clinical.tumor().biopsyLocation();
+        Double purity = molecular.characteristics().purity();
+        if (biopsyLocation != null) {
+            if (purity != null) {
+                Text biopsyText = new Text(biopsyLocation).addStyle(Styles.tableHighlightStyle());
+                Text purityText = new Text(String.format(ApplicationConfig.LOCALE, " (purity %d%%)", Math.round(purity * 100)));
+                purityText.addStyle(molecular.hasSufficientQuality() ? Styles.tableNoticeStyle() : Styles.tableHighlightStyle());
```
kduyvesteyn, line 109: The styles should be swapped around. A low quality sample should get notice.

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -36,113 +37,162 @@ public class RecentMolecularSummaryGenerator implements TableGenerator {
...
+                table.addCell(Cells.createKey("Molecular tissue of origin prediction"));
+                table.addCell(characteristicsGenerator.createPredictedTumorOriginCell());
+            }
 
-        if (!aggregatedEvidence.knownResistantTreatmentsPerEvent().isEmpty()) {
-            table.addCell(Cells.createKey("Events with resistance evidence in " + molecular.evidenceSource()));
-            table.addCell(Cells.createValue(formatTreatmentsPerEvent(aggregatedEvidence.knownResistantTreatmentsPerEvent())));
+            Stream.of(Maps.immutableEntry("Tumor mutational load", characteristicsGenerator.createTMLStatusStringOption()),
+                            Maps.immutableEntry("Microsatellite (in)stability", characteristicsGenerator.createMSStabilityStringOption()),
+                            Maps.immutableEntry("HR status", characteristicsGenerator.createHRStatusStringOption()),
```
kduyvesteyn, line 81: I like it! Feels like the right balance of reuse

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -36,113 +37,162 @@ public class RecentMolecularSummaryGenerator implements TableGenerator {
...
     }
 
     @NotNull
-    private static Cell addIndent(@NotNull Cell cell) {
-        return cell.setPaddingLeft(10);
+    private Optional<String> genesWithHighDriverAmplificationStringOption() {
+        return summaryStringOptionForGeneAlterations(molecular.drivers()
+                .copyNumbers()
+                .stream()
+                .filter(copyNumber -> copyNumber.type().isGain() && !copyNumber.type().isLoss()));
```
kduyvesteyn, line 137: It's a weakness in the datamodel that you need to check it this way. A gain can never be a loss, but that's not enforced. Maybe "partial" should be a boolean in the main copy number datamodel, such type can be either "GAIN", "LOSS", "NONE". Or for now just make a TODO note / jira ticket and leave for later? Best not combine this ticket with datamodel changes I imagine...

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -36,113 +37,162 @@ public class RecentMolecularSummaryGenerator implements TableGenerator {
...
-        Set<String> evidenceStrings = Sets.newTreeSet();
-        for (Map.Entry<String, String> entry : treatmentsPerEvent.entries()) {
-            evidenceStrings.add(entry.getKey() + ": " + entry.getValue());
-        }
-        return concat(evidenceStrings);
+    private Optional<String> genesWithHighDriverDeletionStringOption() {
+        return summaryStringOptionForGeneAlterations(molecular.drivers()
+                .copyNumbers()
+                .stream()
+                .filter(copyNumber -> copyNumber.type().isLoss() && !copyNumber.type().isGain()));
```
kduyvesteyn, line 145: See comment about amps

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
```diff
@@ -36,113 +37,162 @@ public class RecentMolecularSummaryGenerator implements TableGenerator {
...
+                        molecular.drivers().fusions(),
+                        molecular.drivers().homozygousDisruptions(),
+                        molecular.drivers().viruses())
+                .flatMap(Collection::stream)
+                .filter(driver -> driver.driverLikelihood() != DriverLikelihood.HIGH);
+
+        String events = Stream.concat(nonDisruptionDrivers, molecular.drivers().disruptions().stream())
+                .filter(driver -> !driver.evidence().externalEligibleTrials().isEmpty() || eventsWithActinTrials.contains(driver.event()))
+                .map(Driver::event)
+                .collect(Collectors.joining(", "));
```
kduyvesteyn, line 194: Just a thought, but I guess we also always want to display medium/low evidence that has associated standard-of-care evidence. I suspect Nina didn't think of this situation because it is incredibly unlikely that a variant has standard-of-care evidence and is not a HIGH driver.

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/ActinTrialGeneratorFunctions.java
```diff
@@ -0,0 +1,53 @@
...
+            } else {
+                return Cells.createContentNoBorder(text);
+            }
+        }).forEach(table::addCell);
+    }
+
+    public static void insertTrialRow(List<EvaluatedTrial> cohortList, Table table, Table trialSubTable) {
+        EvaluatedTrial trial = cohortList.get(0);
+        if (trial != null) {
+            table.addCell(Cells.createContent(Cells.createContentNoBorder(new Paragraph().addAll(Arrays.asList(new Text(trial.trialId()).addStyle(
```
kduyvesteyn, line 48: This line is too long but somehow my auto-formatting doesn't capture it. Any idea why?
You could maybe split one object creation on other line anyways to make more readable?

report/src/test/java/com/hartwig/actin/report/pdf/TestReportWriterApplication.java
```diff
@@ -33,7 +33,7 @@ public static void main(String[] args) throws IOException {
 
     @NotNull
     private static Report createTestReport() throws IOException {
-        Report report = TestReportFactory.createProperTestReport();
+        Report report = TestReportFactory.createExhaustiveTestReport();
```
kduyvesteyn, line 36: Maybe we should have different functions for "proper test reports" and "exhaustive test reports". The report that can be downloaded from our github page is the one with "proper" data, which does not necessarily have an event of every type

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
jbartletthmf: I'll go ahead and do this.

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/ActinTrialGeneratorFunctions.java
jbartletthmf: No, I don't know why. I liked it better the way I wrote it before the auto-formatter kicked in. I'll fix it, but it's likely to be overwritten on the next commit.

report/src/test/java/com/hartwig/actin/report/pdf/TestReportWriterApplication.java
jbartletthmf: I'll change this back.

jbartletthmf (commented):
Comments addressed in the latest changeset.

common/src/main/java/com/hartwig/actin/molecular/datamodel/driver/VirusType.java
kduyvesteyn: It is somewhat inconsistent with how I've done this in the past (not sure how many of such cases exist in ACTIN, the majority probably in serve or hmftools), but I get the point. One advantage with explicit "display" function is that you can't remove it without breaking code. With toString this is less clear.

report/src/main/java/com/hartwig/actin/report/pdf/components/Footer.java
kduyvesteyn: Updated the disclaimer

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
kduyvesteyn: If I recall correctly, I added isGain() to simplify the code that you propose now, but maybe should revert that...?

report/src/main/java/com/hartwig/actin/report/pdf/tables/treatment/ActinTrialGeneratorFunctions.java
```diff
@@ -0,0 +1,55 @@
...
+                return Cells.createContentNoBorderDeemphasize(text);
+            } else {
+                return Cells.createContentNoBorder(text);
+            }
+        }).forEach(table::addCell);
+    }
+
+    public static void insertTrialRow(List<EvaluatedCohort> cohortList, Table table, Table trialSubTable) {
+        EvaluatedCohort cohort = cohortList.get(0);
+        if (cohort != null) {
```
kduyvesteyn, line 47: I have not checked this exhaustively, but this looks sketchy upon first glance. Can't cohortList be empty? And why would the list itself contain a null value?

report/src/main/java/com/hartwig/actin/report/pdf/tables/molecular/RecentMolecularSummaryGenerator.java
jbartletthmf: `CopyNumber`s are always created from a type (`FULL_GAIN, PARTIAL_GAIN, LOSS, NONE`), and this operation can never result in a `CopyNumber` with both `isGain` and `isLoss` set to true. In that case, it should be sufficient to just check `isGain()` for amplifications and `isLoss()` for deletions. Do you agree?

kduyvesteyn: In theory someone could add one more enum value that sets both to true, but that would be a pretty huge mistake by itself so maybe fine to assume that won't happen. I am fine either way, if you filter amps just by "isGain == true" then I am quite confident that will never be a problem

PR #13, opened 2023-03-09
ACTIN-27: Filter reported drivers by actionability

jbartletthmf:
In molecular details, reported drivers must have known driver
likelihood or be actionable. Additionally:
- Extract VirusDriverLikelihood UNKNOWN as null.
- Use streams to facilitate filtering and concatenation.
- Add toString method to DriverLikelihood.

### DIFF
--- a/report/src/main/java/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactory.java
+++ b/report/src/main/java/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactory.java
@@ -60,8 +60,8 @@ public class MolecularDriverEntryFactory {
                         fromFusions(drivers.fusions()),
                         fromViruses(drivers.viruses()))
                 .flatMap(Function.identity())
-                .filter(driver -> driver.driverLikelihood() != null || !driver.externalTrials().isEmpty() || !driver.actinTrials().isEmpty()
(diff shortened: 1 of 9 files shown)


report/src/main/java/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactory.java
```diff
@@ -145,57 +135,36 @@ private Set<MolecularDriverEntry> fromDisruptions(@NotNull Set<Disruption> disru
...
+        });
     }
 
     @NotNull
-    private Set<MolecularDriverEntry> fromViruses(@NotNull Set<Virus> viruses) {
-        Set<MolecularDriverEntry> entries = Sets.newHashSet();
-
-        for (Virus virus : viruses) {
+    private Stream<MolecularDriverEntry> fromViruses(@NotNull Set<Virus> viruses) {
+        return viruses.stream().map(virus -> {
```
jbartletthmf, line 158: `fromViruses` is the only case where we weren't filtering using `isReportable()`, so I didn't change that. Do we still want to use this filter for other drivers? Should it be applied to Viruses as well? Should we also filter with `isReportable()` on the summary page for consistency?

PR #1005, opened 2025-02-27
ACTIN-1745: Implement HasOnlyHadTreatmentWithCategoryOfTypes

zoehehartwig:
Proposed logics:

PASS: The patient has had one or more treatments that all are of category X and have at least any type Y
FAIL: No treatments, or 1 or more treatments that don’t contain category X and/or type Y
WARN: The patient has one or more treatments that match category X but no types are configured (e.g. “Chemotherapy”) (so in theory it could be a PASS if types were known).

This is bug fix for LKO151 I-2.

Note that radiotherapies and surgeries can be ignored (this rule is about systemic treatment only)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+        val (nullType, wrongTreatments) = diffTreatments.partition {
+            it.categories().contains(category) && it.types().isEmpty()
+        }
+        return when {
+            wrongTreatments.isNotEmpty() -> {
+                EvaluationFactory.fail("There are treatments of the wrong category or type")
+            }
+            nullType.isNotEmpty() -> {
+                EvaluationFactory.warn("There are treatments with null type")
```
jbartletthmf, line 27: We aren't actually using `null` here and "null" doesn't have a clinical meaning so it would be better to use "unknown" in the message and val names.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val types: Set<TreatmentType>
+) : EvaluationFunction {
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }
+        val (matchingTreatments, diffTreatments) = treatments.partition {
+            it.categories().contains(category) && it.types().intersect(types).isNotEmpty()
+        }
+        val (nullType, wrongTreatments) = diffTreatments.partition {
+            it.categories().contains(category) && it.types().isEmpty()
+        }
```
jbartletthmf, line 21: I'd consider using `groupBy` here and assigning keys true, false, or null based on whether the treatment matches the criteria instead of passing over the treatments twice (and having to evaluate the category each time).

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        assertEvaluation(EvaluationResult.FAIL, function.evaluate(withTreatmentHistory(emptyList())))
+    }
+
+    @Test
+    fun `Should fail if there are treatments of the wrong category`() {
+        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(category = TreatmentCategory.SURGERY)))
+    }
+
+    @Test
+    fun `Should fail if there are treatments of the wrong type`() {
```
jbartletthmf, line 27: It's important that we fail for these cases even if the history includes treatments that do match. Could you ensure that scenario is tested as well?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(types = setOf(DrugType.HER2_ANTIBODY))))
+    }
+
+    @Test
+    fun `Should warn if there are treatments of the null type`() {
```
jbartletthmf, line 32: Should be "of unknown type".

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.hartwig.actin.datamodel.algo.EvaluationResult
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.drugTreatment
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatmentHistoryEntry
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.withTreatmentHistory
+import com.hartwig.actin.datamodel.clinical.treatment.DrugType
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import org.junit.Test
+
+class HasOnlyHadTreatmentWithCategoryOfTypesTest {
+    private val function = HasOnlyHadTreatmentWithCategoryOfTypes(TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.EGFR_INHIBITOR_GEN_3))
```
jbartletthmf, line 14: It wouldn't hurt to define `matchingCategory` and `matchingTypes` to reference here and below.

jbartletthmf (commented):
The function looks OK, but it's not connected to anything! I think there are at minimum 5 files that need to be added/modified for a new rule: the rule itself ✅, its test ✅, the corresponding rule mapper, `EligibilityRule.kt`, and the README.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -13,22 +13,26 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
...
         val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }
-        val (matchingTreatments, diffTreatments) = treatments.partition {
-            it.categories().contains(category) && it.types().intersect(types).isNotEmpty()
-        }
-        val (unknownType, wrongTreatments) = diffTreatments.partition {
-            it.categories().contains(category) && it.types().isEmpty()
+        val treatmentMatch = treatments.groupBy {
+            when {
+                it.categories().contains(category) && it.types().intersect(types).isNotEmpty() -> true
+                it.categories().contains(category) && it.types().isEmpty() -> null
```
jbartletthmf, line 19: You could set `val matchesCategory` before the `when` instead of checking it in two conditions.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -13,22 +13,26 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
...
+                else -> false
+            }
         }
         return when {
-            wrongTreatments.isNotEmpty() -> {
+            !treatmentMatch[false].isNullOrEmpty() -> {
```
jbartletthmf, line 24: `groupBy` will only set the keys when values exist, so you could just use `treatmentMatch.containsKey(false)` or `false in treatmentMatch` here and in the other conditions.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -13,22 +13,26 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
 ) : EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
         val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }
-        val (matchingTreatments, diffTreatments) = treatments.partition {
-            it.categories().contains(category) && it.types().intersect(types).isNotEmpty()
-        }
-        val (unknownType, wrongTreatments) = diffTreatments.partition {
-            it.categories().contains(category) && it.types().isEmpty()
+        val treatmentMatch = treatments.groupBy {
```
jbartletthmf, line 16: `treatmentMatch` isn't a great name since we usually use it to refer to something different (i.e. a `TreatmentMatch`). Maybe `treatmentsByMatchEvaluation` would be better?

jbartletthmf (commented):
Looks good! Just minor comments.

jbartletthmf (approved):

cbruel:
Hi @zoehehartwig, I see you’ve already merged this PR, but there are still some things incorrect. This test   @Test
    fun `Should fail if there are treatments of the wrong category`() {
        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(category = TreatmentCategory.SURGERY, types = emptySet())))
    }still some the messages could still use some improvements. It’s important to keep all our messages consistent in format and clear for doctors reading the reports.

PR #1012, opened 2025-03-03
ACTIN-1752 add source id to trial config and trial identification

andreia-hartwig:
Add source id to trial config and trial identification and use it when generating the LKO URL in the report.


report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/ActinTrialGeneratorFunctions.kt
```diff
@@ -68,11 +68,14 @@ object ActinTrialGeneratorFunctions {
...
 
+
             table.addCell(
                 when (cohort.source) {
-                    TrialSource.LKO -> createContent(Paragraph().addAll(trialLabelText.map { it.addStyle(Styles.urlStyle()) })).setAction(
-                        PdfAction.createURI(cohort.trialId.replace("LKO", "https://longkankeronderzoek.nl/studies/"))
-                    )
+                    TrialSource.LKO -> cohort.sourceId?.let {
+                        createContent(Paragraph().addAll(trialLabelText.map { it.addStyle(Styles.urlStyle()) })).setAction(
+                            PdfAction.createURI(cohort.sourceId.replace("LKO", "https://longkankeronderzoek.nl/studies/"))
```
pauldwolfe, line 76: Should we continue to depend on prefixing "LKO" to the ids? I would expect that in the trial we would just have sourceId = "123" and source = "LKO", then this code would be "https://longkankeronderzoek.nl/studies/{sourceId}".

pauldwolfe (approved):

PR #1013, opened 2025-03-03
ACTIN-1755 Rework display of locations

nedleitch:
There are two complete implementations in a row here which is weird.

Maybe just pay attention to the final state, I thought what I had would be fine but thinking about it, I think it was confusing so I changed it.

Make a prefix row for any trial that has more than one cohort when the locations are to be displayed and put it there instead of in the per-cohort rows. Also fix the bug reported by Charlotte where the location sometimes was not showing at all.


nedleitch:
Screenshots from local PDF generation. The page break is somewhat unfortunate but you can see the first trial has two cohorts and thus a "prefix" row with the "Applies to all cohorts below" message, while the second has only one and thus just shows the location directly:

p1
p2

pauldwolfe (approved):

jbartletthmf (approved):

PR #1015, opened 2025-03-06
Actin 1745 fixes

zoehehartwig:
Edit README
Make evaluation function ignore chemotherapy and radiotherapy treatments
Change return statements to be more specific
Edit test cases


algo/README.md
```diff
@@ -234,7 +234,7 @@ The following criterion algorithms ('rules') are available:
...
+| HAS_HAD_ONLY_CATEGORY_X_TREATMENT_OF_TYPES_Y                                               | Treatment history only contains treatment of category defined in X and of any type defined in Y and nothing else.                                                                                                                                                                                                                                                                                                                                                                                                                                            | Resolves to `WARN` if there are any treatments of the correct category but unknown type. Resolves to `FAIL` if there are no treatments.                                                                                                    |
```
jbartletthmf, line 237: I feel like the README should indicate that this rule only considers systemic treatments. Also, maybe we should consider renaming to something like `HAS_HAD_SYSTEMIC_TREATMENT_ONLY_OF_CATEGORY_X_AND_TYPE_Y`. @cbruel, thoughts?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,7 +12,8 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
     private val types: Set<TreatmentType>
 ) : EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }
+        val ignore = setOf( TreatmentCategory.SURGERY, TreatmentCategory.RADIOTHERAPY)
```
jbartletthmf, line 15: If we're going to use categories, I think it would be better to use `TreatmentCategory.SYSTEMIC_CANCER_TREATMENT_CATEGORIES` as suggested by @cbruel so we don't have to maintain a separate list of categories here. This list also doesn't appear to be comprehensive.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,7 +12,8 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
...
+        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }.filter { it.categories().intersect(ignore).isEmpty() }
         val treatmentsByMatchEvaluation = treatments.groupBy {
```
jbartletthmf, line 17: Could inline here:
```suggestion
        val treatmentsByMatchEvaluation = record.oncologicalHistory.flatMap { it.allTreatments() }
            .filter { it.categories().intersect(ignore).isEmpty() }
            .groupBy {
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,7 +12,8 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
     private val types: Set<TreatmentType>
 ) : EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }
+        val ignore = setOf( TreatmentCategory.SURGERY, TreatmentCategory.RADIOTHERAPY)
+        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }.filter { it.categories().intersect(ignore).isEmpty() }
```
jbartletthmf, line 16: @cbruel Would it be better to use `Treatment.isSystemic` over the categories here? That assumes this is set correctly for each treatment, but might be more resilient to special cases where the specific category is unknown.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -21,7 +21,12 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
 
     @Test
     fun `Should fail if there are treatments of the wrong category`() {
-        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(category = TreatmentCategory.SURGERY)))
+        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(category = TreatmentCategory.HORMONE_THERAPY)))
+    }
+
+    @Test
+    fun `Should pass by ignoring surgery and radiotherapy`() {
+        assertEvaluation(EvaluationResult.PASS, function.evaluate(makeRecord(category = TreatmentCategory.SURGERY)))
```
jbartletthmf, line 29: I think it would help to rename `makeRecord` to e.g. `makeRecordWithMatchingAndAdditionalEntry` so it's clear that this case is passing because of the matching record. A patient that has only had a surgery should also produce a FAIL result...could you add a test for this as well?

algo/README.md
```diff
@@ -234,7 +234,7 @@ The following criterion algorithms ('rules') are available:
...
+| HAS_HAD_SYSTEMIC_TREATMENT_ONLY_OF_CATEGORY_X_AND_TYPE_Y                                               | Treatment history only contains systemic treatment of category defined in X and of any type defined in Y and nothing else.                                                                                                                                                                                                                                                                                                                                                                                                                                   | Resolves to `WARN` if there are any treatments of the correct category but unknown type. Resolves to `FAIL` if there are no treatments.                                                                                                    |
```
jbartletthmf, line 237: Should be "Resolves to `FAIL` if there are no *systemic* treatments."

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,9 +12,9 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
     private val types: Set<TreatmentType>
 ) : EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        val ignore = setOf( TreatmentCategory.SURGERY, TreatmentCategory.RADIOTHERAPY)
-        val treatments = record.oncologicalHistory.flatMap { it.allTreatments() }.filter { it.categories().intersect(ignore).isEmpty() }
-        val treatmentsByMatchEvaluation = treatments.groupBy {
+        val treatmentsByMatchEvaluation = record.oncologicalHistory.flatMap { it.allTreatments() }.filter {
+            it.categories().intersect( TreatmentCategory.SYSTEMIC_CANCER_TREATMENT_CATEGORIES ).isNotEmpty()
```
jbartletthmf, line 16: There shouldn't be spaces around `TreatmentCategory.SYSTEMIC_CANCER_TREATMENT_CATEGORIES`.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -38,9 +40,12 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
 
     @Test
     fun `Should pass by ignoring surgery and radiotherapy`() {
+        val treatmentHistoryEntry = treatmentHistoryEntry(setOf(drugTreatment("test", matchingCategory, matchingTypes)))
+        val surgery = treatmentHistoryEntry(setOf(TreatmentTestFactory.treatment(name = "surgery", isSystemic = false, categories = setOf(matchingCategory), types = setOf(OtherTreatmentType.DEBULKING_SURGERY))))
```
jbartletthmf, line 44: It doesn't make sense that the category here is chemo; should that be `SURGERY`?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -38,9 +40,12 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
 
     @Test
     fun `Should pass by ignoring surgery and radiotherapy`() {
```
jbartletthmf, line 42: Did you mean to add a radiotherapy treatment entry as well?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,9 +12,7 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
     private val types: Set<TreatmentType>
 ) : EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        val treatmentsByMatchEvaluation = record.oncologicalHistory.flatMap { it.allTreatments() }.filter {
-            it.categories().intersect( TreatmentCategory.SYSTEMIC_CANCER_TREATMENT_CATEGORIES ).isNotEmpty()
-        }.groupBy {
+        val treatmentsByMatchEvaluation = record.oncologicalHistory.flatMap { it.allTreatments() }.filter { it.isSystemic }.groupBy {
```
jbartletthmf, line 15: With a long call chain like this, I think it helps to put each function call on its own line:
```suggestion
        val treatmentsByMatchEvaluation = record.oncologicalHistory.flatMap { it.allTreatments() }
            .filter { it.isSystemic }
            .groupBy {
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
cbruel: It does not really make a difference I think which one we use, as both fields come from the `treatment.json`. The `category` and `isSystemic` should always be known (only the type can be unknown). But I like the idea of using `Treatment.isSystemic` here!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentRuleMapper.kt
```diff
@@ -85,7 +85,7 @@ class TreatmentRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
             EligibilityRule.HAS_HAD_OBJECTIVE_CLINICAL_BENEFIT_FOLLOWING_TREATMENT_WITH_ANY_NAME_X to hasHadClinicalBenefitFollowingSomeTreatmentCreator(),
             EligibilityRule.HAS_HAD_OBJECTIVE_CLINICAL_BENEFIT_FOLLOWING_CATEGORY_X_TREATMENT to hasHadClinicalBenefitFollowingTreatmentOfCategoryCreator(),
             EligibilityRule.HAS_HAD_OBJECTIVE_CLINICAL_BENEFIT_FOLLOWING_CATEGORY_X_TREATMENT_OF_TYPES_Y to hasHadClinicalBenefitFollowingTreatmentOfCategoryAndTypesCreator(),
-            EligibilityRule.HAS_HAD_ONLY_CATEGORY_X_TREATMENT_OF_TYPES_Y to hasOnlyHadTreatmentWithCategoryOfTypes(),
+            EligibilityRule.HAS_HAD_SYSTEMIC_TREATMENT_ONLY_OF_CATEGORY_X_AND_TYPE_Y to hasOnlyHadTreatmentWithCategoryOfTypes(),
```
cbruel, line 88: `hasHadSystemicTreatmentOnlyOfCategoryOfTypesCreator()` ?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,30 +12,31 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
...
-        }
         return when {
             false in treatmentsByMatchEvaluation -> {
-                EvaluationFactory.fail("There are treatments of the wrong category or type")
+                EvaluationFactory.fail("Did not only receive $types ${category.display()} treatment")
             }
 
             null in treatmentsByMatchEvaluation -> {
-                EvaluationFactory.warn("There are treatments with unknown type")
+                EvaluationFactory.warn("Undetermined if received ${category.display()} is of type $types")
```
cbruel, line 31: I would change this to `EvaluationFactory.undetermined`, this is more in line with our other rules

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypes.kt
```diff
@@ -12,30 +12,31 @@ class HasOnlyHadTreatmentWithCategoryOfTypes(
...
+            .groupBy {
+                val matchesCategory = it.categories().contains(category)
+                when {
+                    matchesCategory && it.types().intersect(types).isNotEmpty() -> true
+                    matchesCategory && it.types().isEmpty() -> null
+                    else -> false
+                }
             }
-        }
         return when {
```
cbruel, line 25: We should add `val typesList = concatItemsWithOr(types)` and change all `$types` to `$typesList` below in the messages (right now we would get messages like `"Did not only receive [PLATINUM_COMPOUND] chemotherapy treatment"` on the report)

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
         assertEvaluation(EvaluationResult.FAIL, function.evaluate(withTreatmentHistory(emptyList())))
     }
 
+    @Test
+    fun `Should fail if there are no treatments that are systemic`() {
+        assertEvaluation(
+            EvaluationResult.FAIL, function.evaluate(
+                withTreatmentHistory(listOf(treatmentHistoryEntry(setOf(drugTreatment("test", TreatmentCategory.SURGERY, matchingTypes)))))
```
cbruel, line 28: Using `drugTreatment` for a surgery seems weird. You could use something like `treatment("surgery", false, setOf(TreatmentCategory.SURGERY), setOf(OtherTreatmentType.DEBULKING_SURGERY))` here.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
...
+            )
+        )
+    }
+
     @Test
     fun `Should fail if there are treatments of the wrong category`() {
-        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(category = TreatmentCategory.SURGERY)))
+        assertEvaluation(
+            EvaluationResult.FAIL,
+            function.evaluate(makeRecordWithMatchingAndAdditionalEntry(category = TreatmentCategory.HORMONE_THERAPY))
```
cbruel, line 37: You could change this to `makeRecordWithMatchingAndAdditionalEntry(category = TreatmentCategory.HORMONE_THERAPY, types = setOf(ANTI_ANDROGEN))`, since the platinum compound used by default is not a type of hormone therapy.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
cbruel: I now notice you also have a surgery entry below in another test. We could make a `private val` for it

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
...
+        )
+    }
+
+    @Test
+    fun `Should pass by ignoring surgery and radiotherapy`() {
```
cbruel, line 42: I think we can make this test name a bit more clear and in line with other test names, something like `"Should pass when only treatment history entry with correct category and type ignoring non systemic treatments"`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
...
+                )
+            )
+        )
+        val radiotherapy = treatmentHistoryEntry(
+            setOf(
+                TreatmentTestFactory.treatment(
+                    name = "radiotherapy",
+                    isSystemic = false,
+                    categories = setOf(TreatmentCategory.RADIOTHERAPY),
+                    types = setOf(OtherTreatmentType.RADIOFREQUENCY)
```
cbruel, line 60: Radiofrequency is Ablation not Radiotherapy. So would change to `categories = setOf(TreatmentCategory.Ablation)` for correctness.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
...
+            function.evaluate(withTreatmentHistory(listOf(treatmentHistoryEntry, surgery, radiotherapy)))
+        )
     }
 
     @Test
     fun `Should fail if there are treatments of the wrong type`() {
-        assertEvaluation(EvaluationResult.FAIL, function.evaluate(makeRecord(types = setOf(DrugType.HER2_ANTIBODY))))
+        assertEvaluation(
+            EvaluationResult.FAIL,
+            function.evaluate(makeRecordWithMatchingAndAdditionalEntry(types = setOf(DrugType.HER2_ANTIBODY)))
```
cbruel, line 75: Would change to `types = setOf(DrugType.ALKYLATING_AGENT)` (HER2_ANTIBODY is not a chemotherapy type)

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasOnlyHadTreatmentWithCategoryOfTypesTest.kt
```diff
@@ -19,29 +21,77 @@ class HasOnlyHadTreatmentWithCategoryOfTypesTest {
...
+        )
+        withTreatmentHistory(listOf(treatmentHistoryEntry, surgery))
+        assertEvaluation(
+            EvaluationResult.PASS,
+            function.evaluate(withTreatmentHistory(listOf(treatmentHistoryEntry, surgery, radiotherapy)))
+        )
     }
 
     @Test
     fun `Should fail if there are treatments of the wrong type`() {
```
cbruel, line 72: `'Should fail if treatment history with correct category but wrong type'`?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    fun `Should fail if there are no treatments that are systemic`() {
+        assertEvaluation(EvaluationResult.FAIL, function.evaluate(withTreatmentHistory(listOf(surgery))))
+    }
+
+    @Test
+    fun `Should fail if there are treatments of the wrong category`() {
+        assertEvaluation(
+            EvaluationResult.FAIL,
+            function.evaluate(makeRecordWithMatchingAndAdditionalEntry(category = TreatmentCategory.HORMONE_THERAPY))
```
cbruel, line 53: `makeRecordWithMatchingAndAdditionalEntry(category = TreatmentCategory.HORMONE_THERAPY, types = setOf(ANTI_ANDROGEN))`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    fun `Should fail if treatment history with correct category but wrong type`() {
+        assertEvaluation(
+            EvaluationResult.FAIL,
+            function.evaluate(makeRecordWithMatchingAndAdditionalEntry(types = setOf(DrugType.ALKYLATING_AGENT)))
+        )
+    }
+
+    @Test
+    fun `Should pass undetermined if there are treatments of unknown type`() {
```
cbruel, line 75: `'Should evaluate to undetermined..'`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypes.kt
```diff
@@ -0,0 +1,46 @@
...
+            }
+
+        val typesList = concatItemsWithOr(types)
+        return when {
+            false in treatmentsByMatchEvaluation -> {
+                EvaluationFactory.fail("Did not only receive $typesList ${category.display()} treatment")
+            }
+
+            null in treatmentsByMatchEvaluation -> {
+                EvaluationFactory.undetermined("Undetermined if received ${category.display()} is of type $typesList")
```
cbruel, line 34: Sorry for not mentioning this in my previous review. But we should also evaluate to undetermined if there is trial medication with unknown category and type, since this trial medication could be of the category and type we are looking for

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypes.kt
```diff
@@ -0,0 +1,54 @@
...
+                when {
+                    matchesCategory && it.types().intersect(types).isNotEmpty() -> true
+                    matchesCategory && it.types().isEmpty() -> null
+                    else -> false
+                }
+            }
+        val trialMedicationWithUnknownCategoryOrIntent = record.oncologicalHistory
+            .filter { it.isTrial }
+            .flatMap { it.allTreatments() }
+            .filter { it.isSystemic && (it.categories().isEmpty() || it.types().isEmpty()) }
```
cbruel, line 29: Looking into how we handle inputs with empty treatmentName (as we will have for trials with unknown treatment), this will end up as a TreatmentHistoryEntry with `treatments = emptySet()`. So what you are checking here (a treatment history entry with `it.categories().isEmpty()`) will never exist. So we can / should only check here `val trialMedicationWithUnknownTreatment = record.oncologicalHistory.filter { it.isTrial }`. The trial medications with known category but unknown types are already handled above in `treatmentsByMatchEvaluation`.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypesTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun `Should evaluate to undetermined if there are treatments of unknown type`() {
+        assertEvaluation(EvaluationResult.UNDETERMINED, function.evaluate(makeRecordWithMatchingAndAdditionalEntry(types = emptySet())))
+    }
+
+    @Test
+    fun `Should evaluate to undetermined if there is a trial medication of unknown category or type`() {
+        val trialTreatmentHistoryEntry = treatmentHistoryEntry(setOf(drugTreatment("test", matchingCategory, emptySet())), isTrial = true)
+        val matchingTreatmentHistoryEntry = treatmentHistoryEntry(setOf(drugTreatment("test", matchingCategory, matchingTypes)))
+        val patientRecord = withTreatmentHistory(listOf(trialTreatmentHistoryEntry, matchingTreatmentHistoryEntry))
+        assertEvaluation(EvaluationResult.UNDETERMINED, function.evaluate(patientRecord))
```
cbruel, line 84: The undetermined outcome here is actually the `null in treatmentsByMatchEvaluation` outcome not the `trialMedicationWithUnknownCategoryOrIntent.isNotEmpty()`. You should also add a test with `val trialTreatmentHistoryEntry = treatmentHistoryEntry(isTrial = true)`, to test that trials with unknown treatment resolve to undetermined

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypes.kt
cbruel: I would just inline this below btw: `record.oncologicalHistory.any { it.isTrial } -> { EvaluationFactory.undetermined .. }`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentOnlyOfCategoryOfTypes.kt
```diff
@@ -0,0 +1,54 @@
...
+
+            null in treatmentsByMatchEvaluation -> {
+                EvaluationFactory.undetermined("Undetermined if received ${category.display()} is of type $typesList")
+            }
+
+            true in treatmentsByMatchEvaluation -> {
+                EvaluationFactory.pass("Has only had $typesList ${category.display()} treatment")
+            }
+
+            trialMedicationWithUnknownCategoryOrIntent.isNotEmpty() -> {
```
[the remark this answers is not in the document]
jbartletthmf: If that's the case, I think the trial check has to be specific enough to exclude trials that satisfy the matching criteria. For example: `record.oncologicalHistory.any { it.isTrial && it.allTreatments().isEmpty() }`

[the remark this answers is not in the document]
cbruel: Agree with John!

cbruel (approved):
Looks good! Approved (would only make the change John suggested)

algo/README.md
```diff
@@ -173,98 +173,99 @@ The following criterion algorithms ('rules') are available:
...
+| HAS_HAD_ONLY_CATEGORY_X_TREATMENT_OF_TYPES_Y                                               | Treatment history only contains treatment of category defined in X and of any type defined in Y and nothing else.                                                                                                                                                                                                                                                                                                                                                                                                                                            | Resolves to `WARN` if there are any treatments of the correct category but unknown type                                                                                                                                                    |
```
jbartletthmf, line 237: It looks like you accidentally undid your previous README improvements here.

jbartletthmf (approved):
Looks good! 🙂

PR #1016, opened 2025-03-06
ACTIN-1774: Move ACTIN clinical datamodels to datamodel module

jbartletthmf:


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/ingestion/IngestionResult.kt
jbartletthmf, line 1: Using these classes in actin-pipelines could reduce a lot of duplication, but they currently use kotlinx for serialization in that project. Rather than add kotlinx serialization to `datamodel` (which we could do if we want to take a step away from gson), we could opt in to the experimental serialization API and generate an external serializer where we need it (https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#deriving-external-serializer-for-another-kotlin-class-experimental). I haven't tried this yet but could test it out if there are no objections.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/ingestion/IngestionResult.kt
```diff
@@ -0,0 +1,83 @@
...
+    val curationResults: Set<CurationResult>,
+    val questionnaireCurationErrors: Set<QuestionnaireCurationError>,
+    val feedValidationWarnings: Set<FeedValidationWarning>
+)
+
+data class CurationWarning(val patientId: String, val category: CurationCategory, val feedInput: String, val message: String)
+
+data class CurationRequirement(val feedInput: String, val message: String)
+
+data class CurationResult(val categoryName: String, val requirements: List<CurationRequirement>) : Comparable<CurationResult> {
```
jbartletthmf, line 30: Now that the categories can be shared with actin-pipelines, we could probably use the category enum in these fields instead of just the name.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/ingestion/IngestionResult.kt
pauldwolfe: For the scope of this ticket, I like your suggestion.

Otherwise... I'm wondering if we should take some steps soon to just conform on one serializer (probably jackson, but maybe we should revisit latest and greatest kotlin serialization first).

pauldwolfe (approved):

PR #1018, opened 2025-03-09
ACTIN-1766: Support for grouped hotspots from SERVE

kzuberihmf:
Corresponding change to ACTIN-1737 in SERVE. Updates hotspot matching.


common/src/test/kotlin/com/hartwig/actin/molecular/evidence/actionability/VariantEvidenceTest.kt
```diff
@@ -12,7 +13,11 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
 
 private val EVIDENCE_FOR_HOTSPOT =
-    TestServeEvidenceFactory.createEvidenceForHotspot(gene = "gene 1", chromosome = "1", position = 5, ref = "A", alt = "T")
+    TestServeEvidenceFactory.createEvidenceForHotspot(
+        TestServeMolecularFactory.createVariantAnnotation(gene = "gene 1", chromosome = "1", position = 5, ref = "A", alt = "T"),
```
jbartletthmf, line 17: It would help to have a constant for the matching variant annotation to make it more obvious where matches are expected.

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/matching/HotspotMatchingTest.kt
```diff
@@ -1,43 +1,87 @@
...
+    fun `Should return true on matching to actionable hotspot`() {
+        assertThat(HotspotMatching.isMatch(ACTIONABLE_HOTSPOT, MATCHING_CRITERIA)).isTrue()
+        assertThat(HotspotMatching.isMatch(ACTIONABLE_HOTSPOT, MATCHING_CRITERIA2)).isTrue()
     }
 
     @Test
-    fun `Should return false on non-matching gene`() {
-        assertThat(HotspotMatching.isMatch(HOTSPOT, MATCHING_CRITERIA.copy(gene = "gene 2"))).isFalse()
+    fun `Should return false on non-matching gene to actionable hotspot`() {
+        assertThat(HotspotMatching.isMatch(ACTIONABLE_HOTSPOT, MATCHING_CRITERIA.copy(gene = "gene 2"))).isFalse()
```
jbartletthmf, line 65: It would be a little less explicit, but do you think this test class would be easier to read if the known and actionable tests were combined? E.g.
```suggestion
    fun `Should return false on non-matching gene`() {
        listOf(KNOWN_HOTSPOT, ACTIONABLE_HOTSPOT).forEach { hotspot ->
            assertThat(HotspotMatching.isMatch(hotspot, MATCHING_CRITERIA.copy(gene = "gene 2"))).isFalse()
        }
```

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifierTest.kt
```diff
@@ -44,7 +44,40 @@ class ServeVerifierTest {
...
+                        TestServeMolecularFactory.createVariantAnnotation(gene = "gene1"),
+                        TestServeMolecularFactory.createVariantAnnotation(gene = "gene2")
+                    )
+                )
+            )
+        )
+
+        val database = toServeDatabase(TestServeEvidenceFactory.create(), trial)
+        assertThatIllegalStateException().isThrownBy { ServeVerifier.verifyServeDatabase(database) }
     }
```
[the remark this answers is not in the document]
kzuberihmf: Good point, thanks.

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/actionability/VariantEvidenceTest.kt
```diff
@@ -110,6 +117,29 @@ class VariantEvidenceTest {
         )
     }
 
+    @Test
+    fun `Should determine evidence and trials for matching within hotspot group`() {
```
kzuberihmf, line 121: lol I think this test is completely useless. maybe replace with a test for non-match, since we don't seem to have that elsewhere below.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
```diff
@@ -34,4 +40,30 @@ object ServeVerifier {
...
+
+    private fun verifyHotspotsInServeRecord(serveRecord: ServeRecord) {
+        serveRecord.evidences()
+            .flatMap { it.molecularCriterium().hotspots() }
+            .forEach { verifyHotspotGeneConsistency(it) }
+
+        serveRecord.trials()
+            .flatMap { it.anyMolecularCriteria() }
+            .flatMap { it.hotspots() }
+            .forEach { verifyHotspotGeneConsistency(it) }
```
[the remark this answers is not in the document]
kzuberihmf: FWIW gene consistency is an invariant of the data that isn't (yet) modeled in the type system. I think this safety test should be quick rather than exhaustive, an error indicates a bug in SERVE and it shouldn't be the case that all individual variants need separate investigation. (I might be wrong here :smile:)

But that brings up a minor thing that bothers me about both my solution and your suggestion: it seems unnecessary to allocate an intermediate collection of thousands of hotspots to perform this check. I actually foolishly thought that the chained maps and foreach would be lazily evaluated, but checking now the bots tell me otherwise. So I achieved neither an efficient check nor exhaustive evaluation as you suggested 😅. I guess I'll revise.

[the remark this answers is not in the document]
jbartletthmf: That's fair, maybe an exhaustive list isn't necessary here.

For lazy evaluations, you could use sequences (https://typealias.com/guides/when-to-use-sequences/). If you prefer not to allocate an intermediate, you could use `filter` instead of `partition` in the first part of my suggestion and then start the second part by flat-mapping to hotspots from `allMolecularCriteria` again. You could also use `find` or `firstOrNull` in place of `filter` if you just want to find 1 or 0 instances that match the criteria.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
```diff
@@ -46,24 +45,52 @@ object ServeVerifier {
...
+                }
+            }
+            throw IllegalStateException("SERVE record contains invalid hotspots:\n$message")
+        }
     }
 
-    private fun verifyHotspotGeneConsistency(actionableHotspot: ActionableHotspot) {
-        if (actionableHotspot.variants().isEmpty()) {
-            throw IllegalStateException("Hotspot contains no variant annotations: $actionableHotspot")
+    private fun verifyHotspotsInServeRecordMine(serveRecord: ServeRecord) {
```
jbartletthmf, line 72: Work in progress?

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifierTest.kt
```diff
@@ -80,6 +80,15 @@ class ServeVerifierTest {
         assertThatIllegalStateException().isThrownBy { ServeVerifier.verifyServeDatabase(database) }
     }
 
+    @Test
+    fun `Should throw exception for hotspot with not variants`() {
```
jbartletthmf, line 84: s/not/no/

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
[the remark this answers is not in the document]
kzuberihmf: cool thanks

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
```diff
@@ -45,49 +45,24 @@ object ServeVerifier {
     }
 
     private fun verifyHotspotsInServeRecord(serveRecord: ServeRecord) {
-        val allMolecularCriteria = serveRecord.evidences().map { it.molecularCriterium() } +
-                serveRecord.trials().flatMap { it.anyMolecularCriteria() }
+        val allMolecularCriteria = serveRecord.evidences().asSequence().map { it.molecularCriterium() } +
+                serveRecord.trials().asSequence().flatMap { it.anyMolecularCriteria() }
```
jbartletthmf, line 49: A Sequence is like a Stream; you can't consume it twice. You either have to build it once for each `flatMap` + `find`, or call `.toList()` here.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
```diff
@@ -45,49 +45,24 @@ object ServeVerifier {
...
-                    append("Hotspots without variants: $hotspotsWithoutVariants\n")
+                if (hotspotWithoutVariants != null) {
+                    append("Hotspot without variants: $hotspotWithoutVariants\n")
                 }
-                if (inconsistentHotspots.isNotEmpty()) {
-                    append("Hotspot with mismatched genes: $inconsistentHotspots\n")
+                if (inconsistentHotspot != null) {
+                    append("Hotspot with mismatched genes: $inconsistentHotspot\n")
                 }
             }
```
jbartletthmf, line 67: Could be:
```kotlin
        val message = listOfNotNull(
            hotspotWithoutVariants?.let { "Hotspot without variants: $it" },
            inconsistentHotspot?.let { "Hotspot with mismatched genes: $it" }
        ).joinToString("\n")
        if (message.isNotEmpty()) {
```

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
jbartletthmf: Hmm, I guess that's not strictly true for sequences, but it still seems unsafe (https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/-sequence/):
> Sequences can be iterated multiple times, however some sequence implementations might constrain themselves to be iterated only once.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeVerifier.kt
```diff
@@ -45,27 +45,27 @@ object ServeVerifier {
...
-                if (hotspotWithoutVariants != null) {
-                    append("Hotspot without variants: $hotspotWithoutVariants\n")
-                }
-                if (inconsistentHotspot != null) {
-                    append("Hotspot with mismatched genes: $inconsistentHotspot\n")
-                }
-            }
+        val message = listOfNotNull(
+            hotspotWithoutVariants?.let { "Hotspot without variants: $it" },
+            inconsistentHotspot?.let { "Hotspot with inconsistent genes: $it" })
```
jbartletthmf, line 59: The closing ")" should go on the next line.

jbartletthmf (approved):
LGTM 🙂

pom.xml
```diff
@@ -23,7 +23,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <serve.version>7.4.1</serve.version>
+        <serve.version>8.0.0-beta.1</serve.version>
```
kduyvesteyn, line 26: Why is the SERVE release a beta? We have tested this properly, would have assumed it would be a proper release.

kduyvesteyn (approved):
Looks all good to, nice job! 

I was at first suspicious of no changes to `VariantEvidence` but have double-checked that and think it's all good (we still match evidence and trials on a per-hotspot level). 

Minor comment on the SERVE version - I would have released SERVE by now already since we won't make changes anymore anyways. And if we do find an issue we would make `8.0.1`. This PR now seems to suggest SERVE itself is still in beta version while it isn't.

PR #1020, opened 2025-03-11
ACTIN-1681: Also warn for matching ICD extension code only when entry…

spdeleeuw:
… is intolerance or toxicity instance and checking for immunotherapy related AEs

Intended logic:
- If the comorbidity instance is toxicity or intolerance: it's an ICD match if the extension code matches to one of the codes in IMMUNOTHERAPY_DRUG_SET, regardless the main code.
- For all other comorbidity instances there has to be a ICD code main match as well (from DRUG_ALLERGY_SET)
- Undetermined for any comorbidity (including tox/intolerance) with main code from DRUG_ALLERGY_SET and no extension code.
- Fail when no immunotherapy in history
- Fail when no comorbidities in history
- Fail when only comorbidities in history with main code not in DRUG_ALLERGY_SET and no extension code or extension code not in IMMUNOTHERAPY_DRUG_SET.

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
@@ -80,7 +80,7 @@ class IcdModel(
 
     fun <T: Comorbidity> findInstancesMatchingAnyExtensionCode(instances: List<T>, targetExtensionCodes: Set<String>): List<Comorbidity> {
         return instances.filter { entity ->
-            entity.icdCodes.flatMap { codeWithAllParents(it.extensionCode) }.toSet().any(targetExtensionCodes::contains)
+            entity.icdCodes.any { codeWithAllParents(it.extensionCode).any(targetExtensionCodes::contains) }
         }
     }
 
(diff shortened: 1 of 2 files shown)


common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
```diff
@@ -78,6 +78,12 @@ class IcdModel(
         )
     }
 
+    fun <T: Comorbidity> findInstancesMatchingAnyExtensionCode(instances: List<T>, targetExtensionCodes: Set<String>): List<Comorbidity> {
+        return instances.filter { entity ->
+            entity.icdCodes.flatMap { codeWithAllParents(it.extensionCode) }.toSet().any(targetExtensionCodes::contains)
```
jbartletthmf, line 83: I think you could do this without building intermediate collections:
```suggestion
            entity.icdCodes.any { codeWithAllParents(it.extensionCode).any(targetExtensionCodes::contains) }
```

jbartletthmf (approved):

PR #1024, opened 2025-03-12
ACTIN-1788: Remove medication data quality mask for beta testing

spdeleeuw:


spdeleeuw:
@pauldwolfe dataqualitymask change we've discussed
@cbruel changes in handling trial medication we've discussed

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractor.kt
```diff
@@ -23,7 +23,9 @@ class StandardMedicationExtractor(
 
     override fun extract(ehrPatientRecord: ProvidedPatientRecord): ExtractionResult<List<Medication>?> {
         return ehrPatientRecord.medications?.map {
-            val atcClassification = if (!it.isTrial && !it.isSelfCare) {
+            val isTrialMedication =
+                it.isTrial || it.name.contains("(studie)", ignoreCase = true) || it.name.contains("studiemedicatie", ignoreCase = true)
```
spdeleeuw, line 27: I've found two ways to describe study drugs. Either `drug_name (STUDIE)` or `studiemedicatie` (without drug name specification)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractor.kt
```diff
@@ -32,7 +34,7 @@ class StandardMedicationExtractor(
                 }
                 it.atcCode?.let { atcCode -> atcModel.resolveByCode(atcCode, "") }
             } else null
-            val atcNameOrInput = atcClassification?.chemicalSubstance?.name ?: it.name
+            val atcNameOrInput = atcClassification?.chemicalSubstance?.name ?: trimmedName(it.name)
             val atcCode = it.atcCode
             val isAntiCancerMedication = MedicationCategories.isAntiCancerMedication(atcCode)
             val drug = treatmentDatabase.findDrugByAtcName(atcNameOrInput)
```
cbruel, line 40: In line 42: if the drug for a trial medication couldn't be found in the drug database this will currently not result in a CurationWarning since isAntiCancerMedication is not true for trialMedications. We could solve this in multiple ways (adding || isTrialMedication to isAntiCancerMedication or adding something to the if statement in line 42), but in all cases we have to be careful that we of course can't except to find a drug for trial medications with name "studiemedicatie", we can only look up the drug for trial medications with "(studie)" I guess. But would we good to add this.

For MedicationExtractor we could simply add `|| isTrialMedication` to the `isAntiCancerMedication` boolean, because we curate all trial medication names, so they should all be found in the drug database. Could you make that change as well? Thanks!

cbruel (commented):
Looks good! Only 1 comment and 1 thought. And (you were probably already planning to do so but) think would be nice to test the clinical ingestion on nki vm to check the clinical records for some patient.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractor.kt
```diff
@@ -23,7 +23,9 @@ class StandardMedicationExtractor(
 
     override fun extract(ehrPatientRecord: ProvidedPatientRecord): ExtractionResult<List<Medication>?> {
         return ehrPatientRecord.medications?.map {
-            val atcClassification = if (!it.isTrial && !it.isSelfCare) {
+            val isTrialMedication =
+                it.isTrial || it.name.contains("(studie)", ignoreCase = true) || it.name.contains("studiemedicatie", ignoreCase = true)
+            val atcClassification = if (!isTrialMedication && !it.isSelfCare) {
                 if (it.atcCode == null) {
                     logger.error(
                         "Patient '${ehrPatientRecord.patientDetails.hashedId}' had medication '${it.name}' with null atc code, " +
```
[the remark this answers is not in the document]
spdeleeuw: I have checked and in all currently available records atcCode is always filled in except for instances where isSelfCare is true

spdeleeuw:
> Looks good! Only 1 comment and 1 thought. And (you were probably already planning to do so but) think would be nice to test the clinical ingestion on nki vm to check the clinical records for some patient.

@cbruel I did run clinical before adding this PR and it runs all the way through. The only issue is this line in class `MedicationCategories`: `val ANTI_CANCER_ATC_CODES = setOf("L01", "L02", "H01CC", "H01CA", "G03XA")`. In input data the code L01ZZ is given to any drug given as part of a trial. However, this also applies to supportive medication (i.e. non-anticancer). Because of the string matching with L01 it results in warnings like these: "Anti cancer medication OMEPRAZOL with ATC code L01ZZ found which is not present in drug database. Please add the missing drug to drug database". Looking into that now.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractor.kt
spdeleeuw: I don't think we should automatically assign trial medication as antiCancerMedication, see also my comment below. Supportive medication given as part of a trial (along with actual anticancer medication) is also annotated with `(STUDIE)`

spdeleeuw: The warning does occur btw, since the "study" ATC code is L01ZZ for all trial-drug (also non-anticancer). See comment below.

cbruel:
@spdeleeuw Hmm those L01ZZ are weird, that's not even an exisiting ATC code. We should somehow ignore those L01ZZ everywhere

spdeleeuw:
> @spdeleeuw Hmm those L01ZZ are weird, that's not even an exisiting ATC code. We should somehow ignore those L01ZZ everywhere

@cbruel I know. Trial drugs are probably explicitly assigned a non-existing ATC code. I think indeed we should not consider L01ZZ an anticancer drug automatically. Not sure how we'd still make sure that the lookup goes well for actual anticancer drugs

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractorTest.kt
```diff
@@ -160,10 +166,41 @@ class StandardMedicationExtractorTest {
         )
     }
 
+    @Test
+    fun `Should trigger warning when anticancer medication cannot be found in drug database`() {
+        every { qtProlongatingDatabase.annotateWithQTProlongating(any()) } returns QTProlongatingRisk.NONE
```
pauldwolfe, line 171: duplicate lines?

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardMedicationExtractorTest.kt
```diff
@@ -160,10 +166,41 @@ class StandardMedicationExtractorTest {
...
+                "Anti cancer medication or supportive trial medication drug with ATC code L01ZZ found which is not present in drug database. " +
+                        "Please add the missing drug to drug database"
+            )
+        )
+    }
+
+    @Test
+    fun `Should not trigger warning for unspecified trial medication`() {
+        every { qtProlongatingDatabase.annotateWithQTProlongating(any()) } returns QTProlongatingRisk.NONE
+        every { qtProlongatingDatabase.annotateWithQTProlongating(any()) } returns QTProlongatingRisk.NONE
```
pauldwolfe, line 190: here too?

pauldwolfe (approved):

PR #1027, opened 2025-03-13
ACTIN-1784: Only consider non-canonical impacts if there are no reportable canonical impacts in GeneHasVariantInCodon

cbruel:
Similar to the bug we found in [ACTIN-1736](https://hartwigmedical.atlassian.net/browse/ACTIN-1736?atlOrigin=eyJpIjoiOTNmYTdiYjM4ZDBlNDI3Yzk3MzRjNWVjYWMzMjE3ZDYiLCJwIjoiaiJ9), also a bug in this class resulted in unwanted messages ".. in canonical transcript together with .. in non-canonical transcript" on the report.

Tested on the report I found this in, it's fixed now.

[ACTIN-1736]: https://hartwigmedical.atlassian.net/browse/ACTIN-1736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
@@ -7,53 +7,77 @@ import com.hartwig.actin.algo.evaluation.util.Format.concatVariants
 import com.hartwig.actin.algo.evaluation.util.Format.percentage
 import com.hartwig.actin.datamodel.algo.Evaluation
 import com.hartwig.actin.datamodel.molecular.MolecularTest
+import com.hartwig.actin.datamodel.molecular.driver.Variant
 import java.time.LocalDate
 
 class GeneHasVariantInCodon(private val gene: String, private val codons: List<String>, maxTestAge: LocalDate? = null) :
     MolecularEvaluationFunction(maxTestAge) {
 
+    private enum class VariantClassification {
+        CANONICAL_REPORTABLE,
+        CANONICAL_REPORTABLE_SUBCLONAL,
+        CANONICAL_UNREPORTABLE,
+        REPORTABLE_OTHER,
+        NONE
+    }
+
     override fun genes() = listOf(gene)
 
     override fun evaluate(test: MolecularTest): Evaluation {
-        val canonicalReportableVariantMatches: MutableSet<String> = mutableSetOf()
-        val canonicalReportableSubclonalVariantMatches: MutableSet<String> = mutableSetOf()
-        val canonicalUnreportableVariantMatches: MutableSet<String> = mutableSetOf()
-        val canonicalCodonMatches: MutableSet<String> = mutableSetOf()
-        val canonicalReportableSubclonalCodonMatches: MutableSet<String> = mutableSetOf()
-        val reportableOtherVariantMatches: MutableSet<String> = mutableSetOf()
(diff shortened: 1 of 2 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
```diff
@@ -7,53 +7,77 @@ import com.hartwig.actin.algo.evaluation.util.Format.concatVariants
...
-        val canonicalCodonMatches: MutableSet<String> = mutableSetOf()
-        val canonicalReportableSubclonalCodonMatches: MutableSet<String> = mutableSetOf()
-        val reportableOtherVariantMatches: MutableSet<String> = mutableSetOf()
-        val reportableOtherCodonMatches: MutableSet<String> = mutableSetOf()
-        for (variant in test.drivers.variants) {
-            if (variant.gene == gene) {
-                for (codon in codons) {
+        val canonicalCodonMatches = mutableSetOf<String>()
+        val canonicalReportableSubclonalCodonMatches = mutableSetOf<String>()
+        val reportableOtherCodonMatches = mutableSetOf<String>()
```
kduyvesteyn, line 29: I wonder if you even need these variables? 

Can't you do the evaluation within the group-by function instead?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
```diff
@@ -7,53 +7,77 @@ import com.hartwig.actin.algo.evaluation.util.Format.concatVariants
 import com.hartwig.actin.algo.evaluation.util.Format.percentage
 import com.hartwig.actin.datamodel.algo.Evaluation
 import com.hartwig.actin.datamodel.molecular.MolecularTest
+import com.hartwig.actin.datamodel.molecular.driver.Variant
 import java.time.LocalDate
 
 class GeneHasVariantInCodon(private val gene: String, private val codons: List<String>, maxTestAge: LocalDate? = null) :
```
kduyvesteyn, line 13: Unrelated to your change but don't understand why `codons` is a list of strings?

We convert this to ints without checking so the code would crash if they don't contain ints anyways!?

kduyvesteyn (approved):
Looks good to me! Some comments provided for thought

kzuberihmf (approved):
Do you think the unit tests need updating for this case?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
cbruel: Because these aren't ints but strings, we curate as input e.g. "V600". So codons will be something like listOf("V600"). We have a `fun isCodon` in `MolecularInputChecker` checking if we used a correct input, so all codons coming in here are correct codon annotations.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodon.kt
cbruel: Spend an hour trying that :), but concluded that this is the cleanest solution (as far as my kotlin skills reach)

PR #1028, opened 2025-03-13
ACTIN-1704 Write list of proteins related to IHC

nedleitch:
Make a list of all proteins referenced by IHC-related rules in the trial database and push them to a new output file.

Per John's suggestion I have refined the parameter types for those rules I could safely say are IHC-related so we can easily find them later.

The list is written in plain text in the output directory, one protein per line.

Some may have been missed, @spdeleeuw you can let me know if this is the case.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
```diff
@@ -306,8 +311,8 @@ class MolecularRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
 
     private fun proteinHasLimitedExpressionByIHCCreator(): FunctionCreator {
         return { function: EligibilityFunction ->
-            val (expressionLevel, protein) = functionInputResolver().createOneStringOneIntegerInput(function)
-            ProteinHasLimitedExpressionByIHC(protein, expressionLevel)
+            val (expressionLevel, protein) = functionInputResolver().createOneProteinOneIntegerInput(function)
+            ProteinHasLimitedExpressionByIHC(expressionLevel, protein)
```
jbartletthmf, line 315: These are backwards (protein is the first field). They could be flipped around...or is it better to use `with`?
```suggestion
            with (functionInputResolver().createOneProteinOneIntegerInput(function)) {
                ProteinHasLimitedExpressionByIHC(proteinName, integer)
            }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
```diff
@@ -270,33 +275,33 @@ class MolecularRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
...
+            ProteinIsExpressedByIHC(functionInputResolver().createOneProteinInput(function).proteinName)
         }
     }
 
     private fun proteinHasExactExpressionByIHCCreator(): FunctionCreator {
         return { function: EligibilityFunction ->
-            val (expressionLevel, protein) = functionInputResolver().createOneStringOneIntegerInput(function)
-            ProteinHasExactExpressionByIHC(protein, expressionLevel)
+            val (expressionLevel, protein) = functionInputResolver().createOneProteinOneIntegerInput(function)
+            ProteinHasExactExpressionByIHC(expressionLevel, protein)
```
jbartletthmf, line 291: backwards

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
```diff
@@ -270,33 +275,33 @@ class MolecularRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
...
         }
     }
 
     private fun proteinHasSufficientExpressionByIHCCreator(): FunctionCreator {
         return { function: EligibilityFunction ->
-            val (expressionLevel, protein) = functionInputResolver().createOneStringOneIntegerInput(function)
-            ProteinHasSufficientExpressionByIHC(protein, expressionLevel)
+            val (expressionLevel, protein) = functionInputResolver().createOneProteinOneIntegerInput(function)
+            ProteinHasSufficientExpressionByIHC(expressionLevel, protein)
```
jbartletthmf, line 298: backwards

trial/src/main/kotlin/com/hartwig/actin/trial/TrialCreatorApplication.kt
```diff
@@ -81,8 +80,10 @@ class TrialCreatorApplication(private val config: TrialCreatorConfig) {
         val outputDirectory = config.outputDirectory
         when (result) {
             is Either.Right -> {
-                LOGGER.info("Writing {} trials to {}", result.value.size, outputDirectory)
+                LOGGER.info("Writing {} trials to [{}]", result.value.size, outputDirectory)
                 TrialJson.write(result.value, outputDirectory)
+                LOGGER.info("Writing list of proteins referenced in inclusion criteria to [{}]", outputDirectory)
+                ProteinList(outputDirectory).writeListOfIhcProteins(result.value)
```
jbartletthmf, line 86: Our convention is to interpolate variables in the string, e.g. `"Writing ${result.value.size} trials"`.

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
```diff
@@ -68,8 +68,8 @@ class TrialIngestion(private val eligibilityFactory: EligibilityFactory) {
                 ).right()
             else UnmappableTrial(trialId = trialState.trialId, trialErrors, unmappableCohorts).left()
         }
-        return if (trialsAndUnmappableTrial.any { it.isLeft }) trialsAndUnmappableTrial.mapNotNull { it.leftOrNull() }
-            .left() else trialsAndUnmappableTrial.mapNotNull { it.getOrNull() }.right()
+        return if (trialsAndUnmappableTrials.any { it.isLeft }) trialsAndUnmappableTrials.mapNotNull { it.leftOrNull() }
+            .left() else trialsAndUnmappableTrials.mapNotNull { it.getOrNull() }.right()
```
jbartletthmf, line 72: I think this could be:
```suggestion
        val (errors, trials) = trialsAndUnmappableTrials.partitionAndJoin()
        return if (errors.isNotEmpty()) errors.left() else trials.right()
```

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
```diff
@@ -0,0 +1,14 @@
+package com.hartwig.actin.trial
+
+import com.hartwig.actin.datamodel.trial.FunctionInput
+import com.hartwig.actin.datamodel.trial.Trial
+import java.nio.file.Files
+import java.nio.file.Path
+
+class ProteinList(val outputDirectory: String) {
+    fun writeListOfIhcProteins(trials: List<Trial>) {
+        var functions = trials.flatMap { it.generalEligibility }.map { it.function }
```
jbartletthmf, line 10: Should be `val`? Also, I think you need to recurse through the composite functions in order to find all the leaves. The `EligibilityRuleUsageEvaluator` does most of what you need (see `extractRules`).

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
```diff
@@ -0,0 +1,14 @@
...
+            .filter { it.rule.input in (listOf(FunctionInput.ONE_PROTEIN, FunctionInput.ONE_PROTEIN_ONE_INTEGER)) }
+        Files.write(Path.of(outputDirectory, "ihc_proteins.list"), functions.map { it.parameters.first().toString() })
```
jbartletthmf, line 12: I think it would be better for the function to return a list of proteins to facilitate testing. The writing itself could be done in the application.

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
```diff
@@ -0,0 +1,14 @@
+package com.hartwig.actin.trial
+
+import com.hartwig.actin.datamodel.trial.FunctionInput
+import com.hartwig.actin.datamodel.trial.Trial
+import java.nio.file.Files
+import java.nio.file.Path
+
+class ProteinList(val outputDirectory: String) {
+    fun writeListOfIhcProteins(trials: List<Trial>) {
```
jbartletthmf, line 9: I don't love the naming here. Maybe it should go in the `EligibilityRuleUsageEvaluator`?

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
```diff
@@ -0,0 +1,14 @@
...
+        var functions = trials.flatMap { it.generalEligibility }.map { it.function }
+            .filter { it.rule.input in (listOf(FunctionInput.ONE_PROTEIN, FunctionInput.ONE_PROTEIN_ONE_INTEGER)) }
```
spdeleeuw, line 11: I like the approach, but unfortunately we're missing the rules with predefined proteins (currently being HER2 and PD-L1):

```
HER2_STATUS_IS_POSITIVE(FunctionInput.NONE),
    PD_L1_SCORE_OF_AT_LEAST_X(FunctionInput.ONE_INTEGER),
    PD_L1_SCORE_OF_AT_MOST_X(FunctionInput.ONE_INTEGER),
    PD_L1_SCORE_CPS_OF_AT_LEAST_X(FunctionInput.ONE_INTEGER),
    PD_L1_SCORE_CPS_OF_AT_MOST_X(FunctionInput.ONE_INTEGER),
    PD_L1_SCORE_TPS_OF_AT_MOST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_SCORE_TAP_OF_AT_MOST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_SCORE_TPS_OF_AT_LEAST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_SCORE_IC_OF_AT_LEAST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_SCORE_TC_OF_AT_LEAST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_SCORE_TAP_OF_AT_LEAST_X(FunctionInput.ONE_DOUBLE),
    PD_L1_STATUS_MUST_BE_AVAILABLE(FunctionInput.NONE),
```

We could consider adding an `isIHC` property to EligibilityRule (`val isIHC: Boolean = false`), though that doesn't feel like the best solution.

trial/src/main/kotlin/com/hartwig/actin/trial/TrialCreatorApplication.kt
nedleitch: Was following this class' convention but will fix :)

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
nedleitch: Indeed. I saw that class and it was tempting to slot this change in there but that felt a bit opportunistic. Happy to move it there.

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
nedleitch: That was indeed where I started with the implementation as a whole. I don't particularly love it however as it feels a bit like trying to shoehorn this functionality into an established object hierarchy.

I have a couple of ideas that I'm trying out locally, stay tuned.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
nedleitch: Thank-you, not sure how that happened.

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
nedleitch: Good pickup, thanks.

trial/src/main/kotlin/com/hartwig/actin/trial/ProteinList.kt
nedleitch: I've reworked it to continue to use the existing code to fetch the attributes from the parameterised rules, but to use an annotation-driven setup for the "hardcoded" ones. This avoids the pitfalls of most of the other approaches.

Not sure what we think of adding the `algo` module to the classpath of the `trial` module, had to be done unless we want to make further changes or come up with some other solution.

jbartletthmf: I don't think trials should have to depend on all of algo just for this. It would be better to indicate the protein with a nullable val on the EligibilityRule, similar to Simon's suggestion.

trial/src/test/kotlin/com/hartwig/actin/trial/EligibilityRuleUsageEvaluatorTest.kt
```diff
@@ -20,4 +20,10 @@ class EligibilityRuleUsageEvaluatorTest {
         val unusedRulesToKeep = EligibilityRule.values().toSet() - expectedUnusedRule - expectedUsedRules
         assertThat(EligibilityRuleUsageEvaluator.evaluate(trials, unusedRulesToKeep)).containsExactly(expectedUnusedRule)
     }
+
+    @Test
+    fun `Should generate list of IHC-adjacent proteins based on parameters and IHC metadata`() {
+        val trials = listOf(TestTrialFactory.createTrialWithIhcRules())
```
jbartletthmf, line 26: Is this the only place this function is used? If so, I think we could make it private and move it here instead of making it part of the `TestTrialFactory` in datamodel.

pom.xml
```diff
@@ -43,6 +43,8 @@
         <apache.log4j.version>2.18.0</apache.log4j.version>
         <slf4j-reload4j.version>1.7.36</slf4j-reload4j.version>
 
+        <reflections.version>0.10.2</reflections.version>
```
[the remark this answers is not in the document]
nedleitch: Thank-you.

trial/src/test/kotlin/com/hartwig/actin/trial/EligibilityRuleUsageEvaluatorTest.kt
nedleitch: I was just checking to see if you were going to say what I thought you would say if I pushed this :)

Indeed it should have been here all along, copied it in-place and forgot to move it.

trial/src/main/kotlin/com/hartwig/actin/trial/EligibilityRuleUsageEvaluator.kt
```diff
@@ -19,30 +20,41 @@ object EligibilityRuleUsageEvaluator {
...
     }
 
-    private tailrec fun extractRules(
-        functions: List<EligibilityFunction>, accumulated: List<EligibilityRule> = emptyList()
+    fun extractIhcProteins(trials: List<Trial>): Set<String> {
+        val allFunctions = collectFunctions(trials.flatMap { it.generalEligibility }.map { it.function })
+        val parameterised =
+            allFunctions.filter { it.rule.input in (listOf(FunctionInput.ONE_PROTEIN, FunctionInput.ONE_PROTEIN_ONE_INTEGER)) }
+                .map { it.parameters.first().toString() }.toSet()
+        return (parameterised + allFunctions.mapNotNull { it.rule.ihcProtein }).toSet()
```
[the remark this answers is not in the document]
nedleitch: Indeed, I did not realise Kotlin would guarantee a set is returned regardless of the type of the second collection.

Nonetheless I will change this slightly to remove the first `toSet()` and keep the second as I prefer that form, it is less surprising.

Thanks.

jbartletthmf (approved):

PR #1029, opened 2025-03-13
ACTIN-1673: Implement logic for CURRENTLY_GETS_CHEMORADIOTHERAPY_OF_TYPE_X_CHEMOTHERAPY_AND_AT_LEAST_Y_CYCLES

zoehehartwig:
Hi, I still need some details on:
when it would resolve to undetermined and 
if these treatments need to be ongoing (since this is named currently and not has had)


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest {
+    @Test
+    fun `Fail if there is no treatments`() {
+        val record = TreatmentTestFactory.withTreatmentHistory(emptyList())
+        assert(EvaluationResult.FAIL, DrugType.ALK_INHIBITOR, 10, record)
+    }
+
+    @Test
+    fun `Pass if there is a match` () {
```
jbartletthmf, line 22: Our convention is for all test names to start with `Should`.

clinical/src/test/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapterTest.kt
```diff
@@ -33,7 +33,7 @@ import org.junit.Test
 private const val PATIENT = "ACTN01029999"
 private val EXPECTED_CLINICAL_RECORD = "${resourceOnClasspath("clinical_record")}/$PATIENT.clinical.json"
 
-class ClinicalIngestionFeedAdapterTest {
+class S {
```
jbartletthmf, line 36: Accident?

clinical/src/test/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapterTest.kt
```diff
@@ -170,6 +170,11 @@ class ClinicalIngestionFeedAdapterTest {
             )
     }
 
+    @Test
+    fun `Should contain validation warning if there is no questionnaire` () {
+
+    }
```
jbartletthmf, line 176: This looks related to another ticket. Sometimes it helps to focus on one task at a time 🙂

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,30 @@
...
+            it.categories().intersect(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY)).isNotEmpty() &&
+                    it.isOfType(type) == true &&
+                    it.treatmentHistoryDetails?.cycles != null &&
+                    it.treatmentHistoryDetails?.cycles!! > minCycles
+        }
+
+        return when {
+            matchingTreatments.isNotEmpty() -> EvaluationFactory.pass("Patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
+            else -> EvaluationFactory.fail("No chemo radio therapy with $type with at least $minCycles is currently received")
+            //else -> EvaluationFactory.undetermined("Undetermined if patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles" )
```
jbartletthmf, line 27: I suspect we'll want to return undetermined when cycles, category, or type is unknown.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,30 @@
...
+) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+
+        val matchingTreatments = record.oncologicalHistory.filter {
+            it.categories().intersect(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY)).isNotEmpty() &&
+                    it.isOfType(type) == true &&
+                    it.treatmentHistoryDetails?.cycles != null &&
+                    it.treatmentHistoryDetails?.cycles!! > minCycles
+        }
```
jbartletthmf, line 22: I don't think we use the `isOngoing` field (it's hard to rely on this since treatments could finish after the record is updated) but we could check that the treatment end date is either unknown or after the reference date.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,30 @@
...
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(
+    private val type: TreatmentType,
+    private val minCycles: Int
+) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+
+        val matchingTreatments = record.oncologicalHistory.filter {
+            it.categories().intersect(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY)).isNotEmpty() &&
```
jbartletthmf, line 18: This will allow all treatments that are chemo AND/OR radiotherapy. I think you should check that both are present here, e.g. `it.categories().containsAll(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY))`.

spdeleeuw (commented):
Hi Zoe, as mentioned in Slack as well: 
We can assume that if it's the most recent line and does not have an end date before the current date (or no end date at all) that the patient is still using it (i.e. ongoing).

All undetermined cases would basically be the cases where you do not have all required data to do a clear evaluation. For instance, when cycles is null, or when you have multiple entries with start and end date null (it's not expected that a patient uses two treatment regimens at the same time, so without any dates you wouldn't know which of those is currently being used and which was used before).
Or when you know it's treatmentCategory Chemotherapy but you don't know the chemotherapy drugType.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+                    false
+                }
+                else -> {
+                    null
+                }
+            }
+        }
+
+        return when {
+            treatmentMatches.isEmpty() -> EvaluationFactory.fail("No treatments received")
```
spdeleeuw, line 38: In an ideal message, the corresponding rule you are evaluating is immediately clear to the reader. I'd suggest always describing what you've evaluated as much as possible, without ending up with a too long message (in this case something like the patient doesn't currently get chemoradiotherapy with type X chemo (something like you did in the else branch).

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+            true in treatmentMatches -> EvaluationFactory.pass("Patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
+            null in treatmentMatches -> EvaluationFactory.undetermined("Undetermined if patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
+            else -> EvaluationFactory.fail("No chemo radio therapy with $type with at least $minCycles is currently received")
```
spdeleeuw, line 41: chemoradiotherapy

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    fun enoughCyclesAndOngoingTreatment(treatmentHistoryDetails: TreatmentHistoryDetails?): Boolean?{
```
spdeleeuw, line 45: function could be private and missing space between ? and {

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+        if (treatmentHistoryDetails == null)
+            return null
+
+        val cycles = treatmentHistoryDetails.cycles
+        if(cycles == null)
```
spdeleeuw, line 50: missing space between if and (

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+            return null
+        else if(cycles < minCycles)
```
spdeleeuw, line 52: missing space after if

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+            return false
```
spdeleeuw, line 53: Alternative for rows 46-53:
```
 val cycles = treatmentHistoryDetails?.cycles ?: return null
 if (cycles < minCycles) return false
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+        val stopMonth = treatmentHistoryDetails.stopMonth
+
+        return when {
+            stopYear == null -> null
+            now.year < stopYear -> true
+            now.year > stopYear -> false
+            stopMonth == null -> null
+            now.monthValue <= stopMonth -> true
+            else -> false
+        }
```
spdeleeuw, line 65: I think having no stop date (i.e. stopYear and stopMonth null) should be considered as ongoing. With the current implementation stopdate null always leads to undetermined.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentType
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryDetails
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(
+    private val type: TreatmentType,
+    private val minCycles: Int
+) : EvaluationFunction {
+
+    private val now = java.time.LocalDate.now()
```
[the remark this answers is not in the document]
jbartletthmf: Correct!

@zoehehartwig, we create a date provider when we start the `TreatmentMatcherApplication` according to the arguments passed in from the command line: `com/hartwig/actin/algo/TreatmentMatcherApplication.kt:62`:
```kotlin
        val referenceDateProvider = ReferenceDateProviderFactory.create(patient, config.runHistorically)
```
This allows us to show results relative to the patient's registration date instead of the current date, which can be useful when re-evaluating old patients.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -6,25 +6,62 @@ import com.hartwig.actin.datamodel.PatientRecord
...
+        val stopMonth = treatmentHistoryDetails.stopMonth
+
+        return when {
+            stopYear == null -> null
+            now.year < stopYear -> true
+            now.year > stopYear -> false
+            stopMonth == null -> null
+            now.monthValue <= stopMonth -> true
+            else -> false
         }
```
jbartletthmf, line 65: I try to avoid putting multiple `return` statements in a single code block because it can make code harder to follow. It's not that different from using `break` or even `goto`. There are also some useful date comparison functions you can use here:
```suggestion
        return treatmentHistoryDetails?.cycles?.let { cycles ->
            val appearsOngoing = with (treatmentHistoryDetails) {
                DateComparison.isAfterDate(referenceDate, stopYear, stopMonth) != false
            }
            cycles >= minCycles && appearsOngoing
        }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.RadiotherapyType
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentType
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryDetails
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
+import org.junit.Test
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest {
+    @Test
+    fun `Fail if there is no treatments`() {
```
jbartletthmf, line 18: `Should fail if there are no treatments`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        val record = TreatmentTestFactory.withTreatmentHistory(emptyList())
+        assert(EvaluationResult.FAIL, DrugType.ALK_INHIBITOR, 10, record)
+    }
+
+    @Test
+    fun `Should pass if there is a match` () {
```
jbartletthmf, line 24: `Should pass if there is a chemotherapy of matching type with sufficient cycles`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,67 @@
...
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(
+    private val type: TreatmentType,
+    private val minCycles: Int
+) : EvaluationFunction {
+
+    private val now = java.time.LocalDate.now()
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val treatmentMatches = record.oncologicalHistory.groupBy {
```
jbartletthmf, line 19: I'm not sure how best to consider whether another line is started after a treatment with unknown stop date. One option would be to first find the maximum start year/month of all treatment entries, and then pass that in to `enoughCyclesAndOngoingTreatment` below. Kotlin has some useful functions for this, like https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/max-of-or-null.html.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -13,55 +15,47 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
             }
         }
 
         return when {
-            treatmentMatches.isEmpty() -> EvaluationFactory.fail("No treatments received")
+            treatmentMatches.isEmpty() -> EvaluationFactory.fail("The patient doesn't currently get chemoradiotherapy with $type chemo")
             true in treatmentMatches -> EvaluationFactory.pass("Patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
             null in treatmentMatches -> EvaluationFactory.undetermined("Undetermined if patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
-            else -> EvaluationFactory.fail("No chemo radio therapy with $type with at least $minCycles is currently received")
+            else -> EvaluationFactory.fail("No chemoradio therapy with $type with at least $minCycles is currently received")
```
jbartletthmf, line 44: Should be "chemoradiotherapy'

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -13,55 +15,47 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
         }
     }
 
-    fun enoughCyclesAndOngoingTreatment(treatmentHistoryDetails: TreatmentHistoryDetails?): Boolean?{
-        if (treatmentHistoryDetails == null)
-            return null
+    private fun enoughCyclesAndOngoingTreatment(treatmentHistoryDetails: TreatmentHistoryDetails?, record: PatientRecord): Boolean? {
+        val referenceDateProvider = ReferenceDateProviderFactory.create(record, true).date()
```
jbartletthmf, line 49: Within this class, we don't know if the user is running against current or historic patients (nor should we). Instead, we should pass in `referenceDate` to its constructor in the `TreatmentRuleMapper`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -13,55 +15,47 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
-        val stopMonth = treatmentHistoryDetails.stopMonth
 
-        return when {
-            stopYear == null -> null
-            now.year < stopYear -> true
-            now.year > stopYear -> false
-            stopMonth == null -> null
-            now.monthValue <= stopMonth -> true
-            else -> false
+        return treatmentHistoryDetails.cycles?.let { cycles ->
```
jbartletthmf, line 54: If you use a null-safe call here (`treatmentHistoryDetails?.cycles`), you could drop the extra null check for `treatmentHistoryDetails` above.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
jbartletthmf: This hasn't been addressed yet. Could we use the latest start year/month from the treatment history to potentially eliminate some previous treatments that presumably aren't ongoing? e.g.
```kotlin
val latestStart = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
```
Then, when evaluating `appearsOngoing`, we can add:
```kotlin
DateComparison.isAfterDate(latestStart, startYear, startMonth) != false
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val type: TreatmentType,
+    private val minCycles: Int
+) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val treatmentMatches = record.oncologicalHistory.groupBy {
+            when {
+                (it.categories().containsAll(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY)) &&
+                        it.isOfType(type) == true &&
+                        enoughCyclesAndOngoingTreatment(it.treatmentHistoryDetails, record) == true) -> {
```
jbartletthmf, line 23: You could define a `val` for `enoughCyclesAndOngoingTreatment` since it's referenced in 2 places. Same for `categories`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -8,17 +8,18 @@ import com.hartwig.actin.datamodel.algo.Evaluation
...
 
 class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(
     private val type: TreatmentType,
     private val minCycles: Int,
-    private val referenceDate: java.time.LocalDate
+    private val referenceDate: LocalDate
 ) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
         val treatmentMatches = record.oncologicalHistory.groupBy {
```
jbartletthmf, line 20: The latest start should be calculated once and then referenced below in `enoughCyclesAndOngoingTreatment` instead of being recalculated each time.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,64 @@
...
+            else -> EvaluationFactory.fail("No chemoradiotherapy with $type with at least $minCycles is currently received")
+        }
+    }
+
+    private fun enoughCyclesAndOngoingTreatment(treatmentHistoryDetails: TreatmentHistoryDetails?, record: PatientRecord): Boolean? {
+        return treatmentHistoryDetails?.cycles?.let { cycles ->
+            val appearsOngoing = with(treatmentHistoryDetails) {
+                val startYear = record.oncologicalHistory.maxOfOrNull { it.startYear ?: 0 }
+                val startMonth = record.oncologicalHistory.maxOfOrNull { it.startMonth ?: 1 }
+                val latestStart = LocalDate.of(startYear ?: 0, startMonth ?: 0, 1)
```
jbartletthmf, line 56: We don't want the max overall month, we want the max year-month. This should look more like `val latestStart = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }` as I proposed earlier.

zoehehartwig:
Should I update the readme?

jbartletthmf:
You'll need to address the conflicts before you can merge:
```
git checkout master
git pull
git checkout ACTIN-1673
git merge master
```
then resolve conflicts and commit + push.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,65 @@
...
+            else -> EvaluationFactory.fail("No chemoradiotherapy with $type with at least $minCycles is currently received")
+        }
+    }
+
+    private fun enoughCyclesAndOngoingTreatment(treatmentHistoryEntry: TreatmentHistoryEntry, latestStart: LocalDate?): Boolean? {
+        val treatmentHistoryDetails = treatmentHistoryEntry.treatmentHistoryDetails
+        return treatmentHistoryDetails?.cycles?.let { cycles ->
+            val appearsOngoing = with(treatmentHistoryDetails) {
+
+                DateComparison.isAfterDate(referenceDate, stopYear, stopMonth) != false &&
```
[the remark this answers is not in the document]
jbartletthmf: No, if the treatment stops after the reference date, we know it's ongoing.

algo/README.md
```diff
@@ -175,6 +175,7 @@ The following criterion algorithms ('rules') are available:
...
+| CURRENTLY_GETS_CHEMORADIOTHERAPY_OF_TYPE_X_CHEMOTHERAPY_AND_AT_LEAST_Y_CYCLES              | Patient currently receives chemotherapy and radiotherapy of type x and of at least a Y cycles                                                                                                                                                                                                                                                                                                                                                                                                                                                                      | Resolves to `UNDETERMINED` if any data is unknown but present data does not explicitly `FAIL`. Checks if the treatment is currently expired and if there are more recent treatments to determine if it is currently being administered.                           |
```
jbartletthmf, line 178: > Patient currently receives chemotherapy and radiotherapy of type x and of at least a Y cycles

This needs some proofreading. Also, the "pass" criteria in the README should do more than just restate the name of the rule. In this case, it would be helpful to indicate how we determine which treatments are current.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val type: TreatmentType,
+    private val minCycles: Int,
+    private val referenceDate: LocalDate
+) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val treatmentMatches = record.oncologicalHistory.groupBy {
+            val matchingCategories = it.categories().containsAll(setOf(TreatmentCategory.CHEMOTHERAPY, TreatmentCategory.RADIOTHERAPY))
+
+            val latestStart = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
```
jbartletthmf, line 23: The key selector function passed into [groupBy](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/group-by.html) is applied to each element, so it's inefficient to do any calculations in this block that don't depend on the `TreatmentHistoryEntry` being classified (`it`). This line needs to be pulled outside of the `groupBy` block so it's only evaluated once.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,65 @@
...
+                (!matchingCategories &&
+                        it.categories().isNotEmpty() ||
+                        it.isOfType(type) == false ||
+                        enoughCyclesAndOngoingTreatment == false) -> {
+                    false
+                }
+
+                else -> {
+                    null
+                }
```
jbartletthmf, line 42: The `when` conditions don't need parentheses, and we could drop the braces around the simple return values if they fit on the same line:
```suggestion
                matchingCategories && it.isOfType(type) == true && enoughCyclesAndOngoingTreatment == true -> true

                !matchingCategories && it.categories().isNotEmpty() || it.isOfType(type) == false || enoughCyclesAndOngoingTreatment == false -> false

                else -> null
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -0,0 +1,65 @@
...
+            treatmentMatches.isEmpty() -> EvaluationFactory.fail("The patient doesn't currently get chemoradiotherapy with $type chemo")
+            true in treatmentMatches -> EvaluationFactory.pass("Patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
+            null in treatmentMatches -> EvaluationFactory.undetermined("Undetermined if patient is currently getting chemoradiotherapy with $type chemotherapy and at least $minCycles cycles")
+            else -> EvaluationFactory.fail("No chemoradiotherapy with $type with at least $minCycles is currently received")
+        }
+    }
+
+    private fun enoughCyclesAndOngoingTreatment(treatmentHistoryEntry: TreatmentHistoryEntry, latestStart: LocalDate?): Boolean? {
+        val treatmentHistoryDetails = treatmentHistoryEntry.treatmentHistoryDetails
+        return treatmentHistoryDetails?.cycles?.let { cycles ->
```
jbartletthmf, line 56: Could be inlined:
```suggestion
        return treatmentHistoryEntry.treatmentHistoryDetails?.cycles?.let { cycles ->
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.Radiotherapy
+import com.hartwig.actin.datamodel.clinical.treatment.RadiotherapyType
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentType
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryDetails
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
+import java.time.LocalDate
+import org.junit.Test
+
+class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest {
```
jbartletthmf, line 17: We're missing a number of test cases here. How can we verify that all scenarios are handled correctly? For example:

1. matching treatment with insufficient cycles
2. matching treatment with future end date
3. matching treatment with unknown end date with other treatment line starting later
4. matching treatment with unknown end date with other treatment line with unknown start date
5. matching category with unknown type
6. ...

jbartletthmf (changes requested):

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
zoehehartwig: I think the 2nd one is already handled by the default PASS test case

algo/README.md
```diff
@@ -175,7 +175,7 @@ The following criterion algorithms ('rules') are available:
...
+| CURRENTLY_GETS_CHEMORADIOTHERAPY_OF_TYPE_X_CHEMOTHERAPY_AND_AT_LEAST_Y_CYCLES              | Patient has a treatment history entry with categories of both chemotherapy and radiotherapy of type x and of at least a Y cycles that is ongoing as of the reference date. Checks whether the treatment is ongoing by if the treatment is past the reference date and if there is no reference date, then if there are more recent treatments to determine if it is currently being administered.                                                                                                                                                                  | Resolves to `UNDETERMINED` if any data is unknown but present data does not explicitly `FAIL`.                                                                                                                                                                    |
```
jbartletthmf, line 178: "of at least ~a~ Y cycles" -> "of at least Y cycles"

Instead of "Checks whether the treatment is ongoing..." you could say, "Excludes treatments that appear to be stopped before the evaluation date based on stop date or the presence of later treatments."

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -55,9 +44,11 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
         return treatmentHistoryDetails?.cycles?.let { cycles ->
             val appearsOngoing = with(treatmentHistoryDetails) {
-
-                DateComparison.isAfterDate(referenceDate, stopYear, stopMonth) != false &&
-                        (latestStart?.let { DateComparison.isAfterDate(latestStart, treatmentHistoryEntry.startYear, treatmentHistoryEntry.startYear) }) != false
+                val treatmentNotStopped = DateComparison.isAfterDate(referenceDate, stopYear, stopMonth)
+                treatmentNotStopped == true ||
+                    (treatmentNotStopped == null && (latestStart?.let {
+                        DateComparison.isAfterDate(latestStart, treatmentHistoryEntry.startYear, treatmentHistoryEntry.startMonth)
+                    }) != false)
```
jbartletthmf, line 51: You can drop the parentheses around `(latestStart?.let {})`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -17,29 +17,18 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
 ) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
+        val latestStart = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
```
jbartletthmf, line 20: Strictly speaking, I think we need to exclude any dates that are greater than the reference date.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
         assertResultForPatient(EvaluationResult.UNDETERMINED, DrugType.ALK_INHIBITOR, 1, record)
     }
 
+    @Test
+    fun `Should FAIL for matching treatment with insufficient cycles`() {
```
jbartletthmf, line 63: We don't usually capitalize the expected results in test names.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+        val matchingTreatment = TreatmentHistoryEntry(
+            treatments = setOf(
+                TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.ALK_INHIBITOR)),
+                Radiotherapy("Radiotherapy", radioType = RadiotherapyType.CYBERKNIFE)
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2030, cycles = 10)
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, DrugType.ALK_INHIBITOR, 20, record)
```
jbartletthmf, line 72: I think it would be more intuitive if you held the `minCycles` consistent, at least for most tests, and varied the details of the patients being tested. For example, you could define `private const val MIN_CYCLES = 10` and then create patients with e.g. `cycles = MIN_CYCLES + 1`. This would allow you to drop the `minCycles` parameter on the helper function.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
         assertResultForPatient(EvaluationResult.UNDETERMINED, DrugType.ALK_INHIBITOR, 1, record)
     }
 
+    @Test
+    fun `Should FAIL for matching treatment with insufficient cycles`() {
+        val matchingTreatment = TreatmentHistoryEntry(
+            treatments = setOf(
+                TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.ALK_INHIBITOR)),
+                Radiotherapy("Radiotherapy", radioType = RadiotherapyType.CYBERKNIFE)
```
jbartletthmf, line 67: Is the `radioType` relevant for any of these tests? I could see an argument for `Should fail if there is only radiotherapy`, but the required type is supposed to be a drug type (`OF_TYPE_X_CHEMOTHERAPY`) so it doesn't really apply there either. It should suffice to leave it out in most cases. You could also define a `private val radiotherapy` to define it in one place.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2030, cycles = 10)
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, DrugType.ALK_INHIBITOR, 20, record)
+    }
+
+    @Test
+    fun `Should FAIL if the end date is null but there is a newer treatment`() {
```
jbartletthmf, line 76: `Should fail`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+            treatments = setOf(TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.ABLATION)),
+            treatmentHistoryDetails = TreatmentHistoryDetails(cycles = 10),
+            startYear = 2022
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment, newerTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, DrugType.ALK_INHIBITOR, 1, record)
+    }
+
+    @Test
+    fun `Should PASS if the end date is null but there is another treatment with an unknown start date`() {
```
jbartletthmf, line 95: `Should pass`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+            treatments = setOf(TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.ABLATION)),
+            treatmentHistoryDetails = TreatmentHistoryDetails(cycles = 10),
+            startYear = null
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment, newerTreatment))
+        assertResultForPatient(EvaluationResult.PASS, DrugType.ALK_INHIBITOR, 1, record)
+    }
+
+    @Test
+    fun `Should be UNDETERMINED if there is a chemotherapy of unknown type with sufficient cycles`() {
```
jbartletthmf, line 114: `Should be undetermined`

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+                Radiotherapy("Radiotherapy", radioType = RadiotherapyType.CYBERKNIFE)
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(cycles = 10),
+            startYear = 2020
+        )
+        val newerTreatment = TreatmentHistoryEntry(
+            treatments = setOf(TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.ABLATION)),
+            treatmentHistoryDetails = TreatmentHistoryDetails(cycles = 10),
+            startYear = null
+        )
```
jbartletthmf, line 108: You could consider using `TreatmentTestFactory.treatmentHistoryEntry` here. Either way, `startYear` is null by default and doesn't have to be provided.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment, newerTreatment))
+        assertResultForPatient(EvaluationResult.PASS, DrugType.ALK_INHIBITOR, 1, record)
+    }
+
+    @Test
+    fun `Should be UNDETERMINED if there is a chemotherapy of unknown type with sufficient cycles`() {
+        val matchingTreatment = TreatmentHistoryEntry(
+            treatments = setOf(
+                TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.ALK_INHIBITOR)),
```
jbartletthmf, line 117: This appears to be a chemotherapy of known type...

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -59,6 +59,70 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2030, cycles = 10)
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.UNDETERMINED, DrugType.ALK_INHIBITOR, 1, record)
+    }
+
     private fun assertResultForPatient(evaluationResult: EvaluationResult, type: TreatmentType, minCycles: Int, record: PatientRecord) {
         val evaluation = CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(
             type, minCycles, LocalDate.of(1900, 1, 1)
```
jbartletthmf, line 128: Rather than using an unrealistically old date, you could set a constant for e.g. `REFERENCE_YEAR` and use it here, then use e.g. ``REFERENCE_YEAR + 1` above when you want to use a later date.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,132 @@
...
+                Radiotherapy("Radiotherapy", radioType = RadiotherapyType.CYBERKNIFE)
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2030, cycles = 10)
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, RadiotherapyType.CYBERKNIFE, 1, record)
+    }
+
+    @Test
+    fun `Should be undetermined if there are no explicit fails`() {
```
jbartletthmf, line 50: Can this name be made more descriptive? e.g. it appears that the chemotherapy type is unknown.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
jbartletthmf: I wasn't intending that to be an exhaustive list, I was just naming some scenarios off the top of my head. Can you think of any others?

6. matching treatment with unknown cycles
7. matching treatment with earlier end date
8. matching category with wrong type
9. ...

algo/README.md
```diff
@@ -175,7 +175,7 @@ The following criterion algorithms ('rules') are available:
...
+| CURRENTLY_GETS_CHEMORADIOTHERAPY_OF_TYPE_X_CHEMOTHERAPY_AND_AT_LEAST_Y_CYCLES              | Patient has a treatment history entry with categories of both chemotherapy and radiotherapy of type x and of at least y cycles that is ongoing as of the reference date. Excludes treatments that appear to be stopped before the evaluation date based on stop date or the presence of later treatments.                                                                                                                                                                                                                                                          | Resolves to `UNDETERMINED` if any data is unknown but present data does not explicitly `FAIL`.                                                                                                                                                                    |
```
jbartletthmf, line 178: The convention in the README is to use uppercase X and Y.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -17,7 +17,14 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val latestStart = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
+        val latestStartAll = record.oncologicalHistory.maxOfOrNull { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
+        val latestStart = if (latestStartAll?.let { DateComparison.isAfterDate(it, referenceDate.year, referenceDate.monthValue) } == true){
+            referenceDate
+        }
+        else{
+            latestStartAll
+        }
```
jbartletthmf, line 26: I don't see why we'd want to set the latest start date to the reference date. How about:
```suggestion
        val latestStartAll = record.oncologicalHistory.map { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
            .sortedDescending()
            .firstOrNull { !it.isAfter(referenceDate) }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -15,10 +15,11 @@ import java.time.LocalDate
 import org.junit.Test
 
 class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest {
+    private val MIN_CYCLES = 5
     @Test
```
jbartletthmf, line 19: There should be a blank line before the first test.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -48,7 +51,7 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
     }
 
     @Test
-    fun `Should be undetermined if there are no explicit fails`() {
+    fun `Should be undetermined if there are no explicit fails but the types is undetermined`() {
```
jbartletthmf, line 54: What does it mean that "there are no explicit fails"? To me, this should be named something like `Should be undetermined for current chemoradiotherapy with matching cycles and unknown chemo type`.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -124,9 +126,67 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+    }
+
+    @Test
+    fun `Should fail for matching treatment with an end date before the reference date`() {
+        val matchingTreatment = TreatmentHistoryEntry(
+            treatments = setOf(
+                TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.ALK_INHIBITOR)),
+                RADIOTHERAPY
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2022, cycles = MIN_CYCLES)
```
jbartletthmf, line 136: It would help if you define these years with respect to `REFERENCE_YEAR` so we can easily see how these entries compare.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -124,9 +126,67 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
...
+    fun `Should fail if the category matches but the type is wrong`() {
+        val matchingTreatment = TreatmentHistoryEntry(
+            treatments = setOf(
+                TreatmentTestFactory.drugTreatment("Alk Inhibitor", TreatmentCategory.CHEMOTHERAPY, setOf(DrugType.ALK_INHIBITOR)),
+                RADIOTHERAPY
+            ),
+            treatmentHistoryDetails = TreatmentHistoryDetails(stopYear = 2030, cycles = MIN_CYCLES)
+        )
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, DrugType.ABL_INHIBITOR, record)
```
jbartletthmf, line 152: I think it would be easier to follow if the function criteria were kept consistent (i.e. `ALK_INHIBITOR`) and the patient/treatment details varied instead. You could also define a `MATCHING_TYPE` constant to make this even more obvious.

jbartletthmf (approved):

spdeleeuw (approved):

PR #1032, opened 2025-03-14
ACTIN-1743: Add implementation for IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X - Osimertinib

cbruel:
@spdeleeuw Logic used is from https://www.nvmo.org/wp-content/uploads/2025/03/KNT_lijst_NSCLC_v3-2025-02-03.pdf


algo/README.md
```diff
@@ -176,7 +176,7 @@ The following criterion algorithms ('rules') are available:
...
+| IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X                                                       | CRC patients are evaluated using CrcDecisionTree and will resolve to `UNDETERMINED` if patient is eligible for treatment X. Osimertinib is evaluated for NSCLC patients and will `PASS` in case EGFR exon19 deletion, EGFR L858R mutation, EGFR activating mutation (except for EGFR exon20 insertion) or EGFR T790M with prior TKI gen 1/2 treatment. `WARN` in case the patient has already received treatment X. Will resolve to `UNDETERMINED` in all other cases.                                                                                       |                                                                                                                                                                                                                                                                   |
```
spdeleeuw, line 179: Maybe you could specify that the WARN in case already administered does not apply to osimertinib?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -32,6 +52,22 @@ class IsEligibleForOnLabelTreatment(
                 }
             }
 
+            isNSCLC && treatmentNameToEvaluationFunctionsForNSCLC.containsKey(treatmentDisplay) -> {
+                when (evaluate(record, treatmentNameToEvaluationFunctionsForNSCLC[treatmentDisplay]!!).result) {
```
jbartletthmf, line 56: Ultimately, I think this kind of logic is a better fit in the `StandardOfCareEvaluator`. We're building up a model of NSCLC standard of care but it's hidden in the evaluation functions instead of defined in one place. I don't think we have a full NSCLC SOC model yet, but we could add something like `StandardOfCareEvaluator.isEligibleForOnLabelTreatment(record, treatment): Evaluation?` that could check if we've defined enough of the SOC model to return an evaluation.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -32,6 +52,22 @@ class IsEligibleForOnLabelTreatment(
...
+
+                    else -> {
+                        EvaluationFactory.fail("Not eligible for on-label treatment $treatmentDisplay")
+                    }
+                }
+            }
+
             record.oncologicalHistory.flatMap { it.allTreatments() }.any { it.name.equals(treatment.name, ignoreCase = true) } -> {
                 EvaluationFactory.warn(
                     "Patient might be ineligible for on-label $treatmentDisplay since this treatment was already administered"
```
jbartletthmf, line 73: Should this warning still be shown for NSCLC patients that have already received the treatment? Should the SOC model for that case consider past treatments?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
jbartletthmf: That change may be non-trivial though as the SOC code is pretty CRC-specific at this time.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -47,4 +83,39 @@ class IsEligibleForOnLabelTreatment(
...
+                ),
+                And(
+                    listOf(
+                        GeneHasVariantWithProteinImpact("EGFR", setOf("T790M"), maxTestAge),
+                        HasHadSomeTreatmentsWithCategoryOfTypes(
+                            TreatmentCategory.TARGETED_THERAPY,
+                            setOf(DrugType.TYROSINE_KINASE_INHIBITOR_GEN_1, DrugType.TYROSINE_KINASE_INHIBITOR_GEN_2),
+                            1
+                        ),
+                        HasHadLimitedSystemicTreatments(1)
```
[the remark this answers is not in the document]
cbruel: @spdeleeuw I added this based on "Indien een 1ste of 2de generatie TKI is gegeven in de 1ste lijn en er een EGFR exon 20 T790M resistentie mutatie gevonden wordt, kan in de 2de lijn behandeld worden met osimertinib". I also find this is other sources like ZorgInstituut. If a pt first got chemo and then EGFR TKI, isn't the patient considered 3th line? Or do you mean this could be considered 2nd line? Then it would we better to leave HasHadLimitedSystemicTreatments(1) out.

But I think I will ask Korneel to ask Joop tomorrow what mutations/conditions he considers eligible for osimertinib. To be sure we implement this rule correctly?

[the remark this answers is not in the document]
spdeleeuw: A key concern is the eligibility for osimertinib in patients who received neoadjuvant/adjuvant chemotherapy and surgery for lower-stage disease, but later progressed to stage IV. It's challenging to distinguish between curative and metastatic treatment lines, making `HasHadLimitedSystemicTreatments(1)` problematic. However, in practice, patients with two prior metastatic lines would likely have already received osimertinib, suggesting we can omit this condition without generating false positives.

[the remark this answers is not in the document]
spdeleeuw: Apart from this, good idea to ask Joop tomorrow.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
cbruel: Agree ultimately this should be in `StandardOfCareEvaluator`, but as you already say, this doesn't feel a suitable place to add it right now, so therefore added it here

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -47,4 +83,39 @@ class IsEligibleForOnLabelTreatment(
...
+        val evaluation = evaluationFunctions.evaluate(record)
+        return evaluation.copy(
+            inclusionMolecularEvents = emptySet(),
+            exclusionMolecularEvents = emptySet(),
+            isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
+        )
+    }
+
+    private val treatmentNameToEvaluationFunctionsForNSCLC: Map<String, EvaluationFunction> = mapOf(
+        "Osimertinib" to Or(
```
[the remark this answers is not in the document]
cbruel: I tried to keep this implementation as general as possible, so we could add more treatments in the future. But agree on your other comment, that if we add more treatments for NSCLC this should ideally be evaluated in the `StandardOfCareEvaluator`. I'm not really sure though if and when we expect this to grow. 

Personally I still like the more general approach (Map<String, EvaluationFunction>) and seems a bit weird to me to make it really specific only for Osimertinib

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -32,6 +52,22 @@ class IsEligibleForOnLabelTreatment(
...
+                    }
+
+                    EvaluationResult.WARN -> {
+                        EvaluationFactory.undetermined("Undetermined if patient is eligible for on-label treatment $treatmentDisplay")
+                    }
+
+                    else -> {
+                        EvaluationFactory.fail("Not eligible for on-label treatment $treatmentDisplay")
+                    }
+                }
```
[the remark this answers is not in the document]
cbruel: Good point though I think this will make the messages very long. Especially out of these molecular functions we can get some long warn messages. I don't think it's really needed to have these in this messages

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
cbruel: Sorry I do not fully understand this comment. But maybe this answers it: my addition to this rule should only change the evaluation for NSCLC patients when evaluating Osimertinib. Evaluating eligibility for other treatments for a NSCLC patient, should stay the same

jbartletthmf: Should an NSCLC patient who has already received Osimertinib be eligible for Osimertinib? As it stands, they'll only be evaluated with the SOC criteria below, which don't seem to consider the presence of Osimertinib in the treatment history.

cbruel: Ah thanks, I understand your point. @spdeleeuw I think John is correct, and it would be better to move the warn block above the block I added, so a patient who already received osimertinib will evaluate to warn. What do you think?

cbruel: I think this would still work for the criterium we want to use this for “The patient receives osimertinib 80 mg QD as part of their standard treatment plan.” -> OR(AND(IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X[Osimertinib], NOT(HAS_HAD_TREATMENT_NAME_X[Osimertinib])), HAS_HAD_TREATMENT_NAME_X_WITHIN_Y_WEEKS[Osimertinib, 2]).

Since if a patient received osimertinib recently this will result in: OR(AND(WARN, FAIL), PASS) -> PASS
And for a patient who got osimertinib 3 months ago this will result in: OR(AND(WARN, FAIL), FAIL) -> FAIL

jbartletthmf: I agree that captures the requirement, but should `NOT(HAS_HAD_TREATMENT_NAME_X[Osimertinib])` be considered part of the eligibility requirements for Osimertinib? Otherwise, won't we have to check this everywhere we check eligibility?

For CRC SOC, we consider exclusions based on previous treatments (e.g. `com.hartwig.actin.algo.soc.TreatmentCandidateDatabase#eligibleIfDrugsNotInRecentHistoryOrWithPD`) and we warn for most other treatments here in `IsEligibleForOnLabelTreatment`.

cbruel: Have added this. Thanks for thinking along!

jbartletthmf (approved):

PR #1033, opened 2025-03-14
ACTIN-1794: Add Allergic or hypersensitivity conditions ICD block to HAS_INTOLERANCE_RELATED_TO_STUDY_MEDICATION

cbruel:


cbruel:
@spdeleeuw Very good point, thanks!

PR #1034, opened 2025-03-17
ACTIN-1775: Clarify insufficient molecular data message

cbruel:
This is the one we recently talked about in slack. I played around with a few solutions, but came to the conclusion that we already have a solution for this (the `noMolecularRecordEvaluation()`), we just hadn't implemented it yet for these functions. Tested on real patient, and looks good.


kzuberihmf (approved):
Paul's the export on the intentions of the molecular override functions. I was a bit worried about adding behaviour that depends on if the upstream source of the molecular data is from orange or not, but I don't see issues with that here. LGTM.

pauldwolfe (approved):
Looks good

PR #1035, opened 2025-03-18
ACTIN-1787: Use CurationCategory enum in clinical ingestion datamodel

jbartletthmf:
Exclude ClinicalRecord from IngestionResult and refactor.


clinical/src/main/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapter.kt
```diff
@@ -1,27 +1,32 @@
...
 
 import com.hartwig.actin.clinical.curation.CurationDatabaseContext
+import com.hartwig.actin.clinical.curation.extraction.CurationExtractionEvaluation
 import com.hartwig.actin.clinical.feed.ClinicalFeedIngestion
 import com.hartwig.actin.clinical.sort.ClinicalRecordComparator
+import com.hartwig.actin.datamodel.clinical.ClinicalRecord
 import com.hartwig.actin.datamodel.clinical.ingestion.IngestionResult
+import com.hartwig.actin.datamodel.clinical.ingestion.PatientIngestionResult
+
+typealias IngestionResultWithClinicalRecords = Pair<IngestionResult, List<ClinicalRecord>>
```
andreia-hartwig, line 11: IngestionResultWithClinicalRecords is only used once in ClinicalIngestionFeedAdapter.run so wondering if it would not be simpler to use the Pair directly in the run function.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/StandardDataIngestion.kt
```diff
@@ -63,79 +64,82 @@ class StandardDataIngestion(
...
-            val sequencingTests = sequencingTestExtractor.extract(ehrPatientRecord)
+    override fun ingest(): List<Triple<ClinicalRecord, PatientIngestionResult, CurationExtractionEvaluation>> {
+        return Files.list(Paths.get(directory)).filter { it.name.endsWith("json") }
+            .map(::recordAndEvaluationFromPath)
+            .map { (record, evaluation) ->
+                Triple(
+                    record,
+                    PatientIngestionResult(
+                        record.patientId,
+                        if (evaluation.warnings.isEmpty()) PatientIngestionStatus.PASS else PatientIngestionStatus.WARN_CURATION_REQUIRED,
```
andreia-hartwig, line 75: I wonder why only PASS or WARN_CURATION_REQUIRED are taken into account here since WARN_CURATION_REQUIRED is also a possible value.

andreia-hartwig (approved):

clinical/src/main/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapter.kt
jbartletthmf: We could, the alias doesn't do much. I think Paul previously suggested using an alias for pairs returned by public functions for clarity.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/StandardDataIngestion.kt
jbartletthmf: The other possible status is `WARN_NO_QUESTIONNAIRE`, which doesn't apply here because standard ingestion doesn't use a questionnaire. The EMC code uses all statuses.

clinical/src/main/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapter.kt
pauldwolfe: Did I? Actually I'm fine with just the pairs in this case.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/StandardDataIngestion.kt
pauldwolfe: I think we should really just have PASS and WARN (and maybe FAIL?) as status rather than these WARN+reason. If you do consider that change keep in mind Zoe is working on a ticket to send NO_QUESTIONNAIRE message to slack.

pauldwolfe (approved):

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/StandardDataIngestion.kt
jbartletthmf: I agree, and this ties in with the discussion being had here: https://github.com/hartwigmedical/actin-pipelines/pull/203#discussion_r2003067852.

PR #1036, opened 2025-03-18
ACTIN-1797: Do not show SOC options based on CUPPA result of insufficient quality WGS sample

kduyvesteyn:
I have verified that for ACTN01020436 this leads to desired change

If either of you could review that would be appreciated! (both is overkill I think)

### DIFF
--- a/report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
+++ b/report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
@@ -106,14 +106,8 @@ class TumorOriginInterpreterTest {
 
     @Test
     fun `Should display at most three predictions in summary`() {
-        assertThat(
-            withPredictions(
-                0.4,
-                0.12,
-                0.15,
-                0.25
-            ).generateSummaryString()
-        ).isEqualTo("Inconclusive (type 1 40%, type 4 25%, type 3 15%)")
+        assertThat(withPredictions(0.4, 0.12, 0.15, 0.25).generateSummaryString())
+            .isEqualTo("Inconclusive (type 1 40%, type 4 25%, type 3 15%)")
     }
 
     @Test
@@ -122,7 +116,7 @@ class TumorOriginInterpreterTest {
     }
 
     @Test
-    fun `Should return 'unknown' predicted tumor origin when inconclusive with insufficient quality and purity`() {
+    fun `Should return 'unknown' predicted tumor origin when inconclusive with insufficient quality`() {
         assertThat(insufficientQualityInconclusiveInterpreter.generateSummaryString()).isEqualTo("Unknown")
     }
 
--- a/report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
+++ b/report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
@@ -24,21 +24,36 @@ class TumorOriginInterpreterTest {
             featureClassifier = 0.0102
         ),
     )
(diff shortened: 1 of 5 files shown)


report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
```diff
@@ -71,48 +86,49 @@ class TumorOriginInterpreterTest {
...
-        assertThat(withPredictions(0.4, 0.12, 0.15, 0.25).generateSummaryString(true))
-            .isEqualTo("Inconclusive (type 1 40%, type 4 25%, type 3 15%)")
+        assertThat(
+            withPredictions(
+                0.4,
+                0.12,
+                0.15,
+                0.25
+            ).generateSummaryString()
+        ).isEqualTo("Inconclusive (type 1 40%, type 4 25%, type 3 15%)")
```
spdeleeuw, line 116: Probably auto-formatting at work here, but maybe leave the `withPredictions(` on one line for readability?

report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
```diff
@@ -71,48 +86,49 @@ class TumorOriginInterpreterTest {
...
 
     @Test
     fun `Should return 'unknown' predicted tumor origin when conclusive with insufficient quality`() {
-        assertThat(conclusiveInterpreter.generateSummaryString(hasSufficientQuality = false))
-            .isEqualTo("Unknown")
+        assertThat(insufficientQualityConclusiveInterpreter.generateSummaryString()).isEqualTo("Unknown")
     }
 
     @Test
     fun `Should return 'unknown' predicted tumor origin when inconclusive with insufficient quality and purity`() {
```
spdeleeuw, line 125: Why does the test title contain quality AND purity while the one above only contains quality? Maybe change for consistency? Since only the word conclusive/inconclusive should be the difference here

report/src/test/kotlin/com/hartwig/actin/report/interpretation/TumorOriginInterpreterTest.kt
```diff
@@ -71,48 +86,49 @@ class TumorOriginInterpreterTest {
...
-        assertThat(inconclusiveInterpreter.generateSummaryString(hasSufficientQuality = false))
-            .isEqualTo("Unknown")
+        assertThat(insufficientQualityInconclusiveInterpreter.generateSummaryString()).isEqualTo("Unknown")
     }
 
     @Test
     fun `Should return 'unknown' predicted tumor origin when there is no prediction in molecular record`() {
-        assertThat(TumorOriginInterpreter(null).generateSummaryString(true)).isEqualTo("Unknown")
+        assertThat(nullInterpreter.generateSummaryString()).isEqualTo("Unknown")
```
spdeleeuw, line 131: Some redundancy here? See above:
```
@Test
    fun `Should return empty list for display when predicted tumor origin is null`() {
        assertThat(TumorOriginInterpreter(null).topPredictionsToDisplay()).isEmpty()
        assertThat(nullInterpreter.topPredictionsToDisplay()).isEmpty()
    }
```
Do we need these both?

kduyvesteyn: I don't think there is redundancy? One test tests the summary string upon null data and the other one checks the top predictions to display (unless I missed something!)

spdeleeuw: My bad, you're right

PR #1037, opened 2025-03-18
ACTIN-1783: Review Evaluation datamodel and EvaluationFactory usage

spdeleeuw:
See JIRA ticket

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
@@ -8,31 +8,26 @@ import com.hartwig.actin.datamodel.algo.EvaluationResult
 class WarnIf(private val function: EvaluationFunction) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val evaluation: Evaluation = function.evaluate(record)
-        if (evaluation.result == EvaluationResult.PASS) {
-            return Evaluation(
-                result = EvaluationResult.WARN,
-                recoverable = evaluation.recoverable,
-                inclusionMolecularEvents = emptySet(),
-                exclusionMolecularEvents = emptySet(),
-                warnMessages = evaluation.passMessages,
-                isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
-            )
-        } else if (evaluation.result == EvaluationResult.WARN) {
-            return evaluation.copy(
-                inclusionMolecularEvents = emptySet(),
-                exclusionMolecularEvents = emptySet(),
-                isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
-            )
-        }
+        val evaluation = function.evaluate(record)
+        return when (evaluation.result) {
+            EvaluationResult.PASS -> {
+                Evaluation(
+                    result = EvaluationResult.WARN,
+                    recoverable = evaluation.recoverable,
+                    warnMessages = evaluation.passMessages
+                )
+            }
 
-        return Evaluation(
-            result = EvaluationResult.PASS,
-            recoverable = evaluation.recoverable,
-            inclusionMolecularEvents = emptySet(),
-            exclusionMolecularEvents = emptySet(),
-            passMessages = (evaluation.passMessages + evaluation.warnMessages + evaluation.undeterminedMessages + evaluation.failMessages),
-            isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
-        )
+            EvaluationResult.WARN -> evaluation.copy(inclusionMolecularEvents = emptySet(), exclusionMolecularEvents = emptySet())
+
+            else -> {
+                Evaluation(
+                    result = EvaluationResult.PASS,
+                    recoverable = evaluation.recoverable,
+                    passMessages = (evaluation.passMessages + evaluation.warnMessages + evaluation.undeterminedMessages + evaluation.failMessages),
+                    isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
+                )
+            }
+        }
     }
 }
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
@@ -16,7 +16,7 @@ class HasMaximumWHOStatus(private val maximumWHO: Int) : EvaluationFunction {
 
             who <= maximumWHO -> EvaluationFactory.pass("WHO $who is within requested max WHO $maximumWHO")
 
-            who - maximumWHO == 1 -> EvaluationFactory.recoverableFail("WHO $who exceeds max WHO $maximumWHO")
+            who - maximumWHO == 1 -> EvaluationFactory.warn("WHO $who exceeds max WHO $maximumWHO")
 
             else -> EvaluationFactory.fail("WHO $who exceeds max WHO $maximumWHO")
         }
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
@@ -16,7 +16,7 @@ class HasMaximumWHOStatus(private val maximumWHO: Int) : EvaluationFunction {
 
             who <= maximumWHO -> EvaluationFactory.pass("WHO $who is within requested max WHO $maximumWHO")
 
-            who - maximumWHO == 1 -> EvaluationFactory.warn("WHO $who exceeds max WHO $maximumWHO")
+            who - maximumWHO == 1 -> EvaluationFactory.recoverableFail("WHO $who exceeds max WHO $maximumWHO")
 
             else -> EvaluationFactory.fail("WHO $who exceeds max WHO $maximumWHO")
         }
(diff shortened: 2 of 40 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasExhaustedSOCTreatments.kt
```diff
@@ -55,12 +55,8 @@ class HasExhaustedSOCTreatments(
                 }
             }
 
-            record.oncologicalHistory.isEmpty() -> {
-                EvaluationFactory.undetermined("Exhaustion of SOC undetermined (no prior cancer treatment)")
-            }
-
             else -> {
-                EvaluationFactory.notEvaluated("Assumed that SOC is exhausted (had prior cancer treatment)")
+                EvaluationFactory.undetermined("Exhaustion of SOC undetermined")
```
cbruel, line 59: This change from `notEvaluated` -> `undetermined` will lead to quite some extra messages on the report (for non-CRC and non-NSCLC patients)? We once implemented it this way (https://hartwigmedical.atlassian.net/browse/ACTIN-115) to prevent these messages on the report

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/priortumor/HasActiveSecondMalignancy.kt
```diff
@@ -12,7 +12,7 @@ class HasActiveSecondMalignancy: EvaluationFunction {
         return if (record.priorSecondPrimaries.any { it.status == TumorStatus.ACTIVE }) {
             EvaluationFactory.pass("Presence of active second malignancy")
         } else if (record.priorSecondPrimaries.any { it.status == TumorStatus.EXPECTATIVE }) {
-            EvaluationFactory.warn("Presence of expected second malignancy")
+            EvaluationFactory.warn("Presence second malignancy with expectative status")
```
cbruel, line 15: `Presence of second malignancy with expectative status`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/othercondition/HasSevereConcomitantIllness.kt
```diff
@@ -14,7 +14,7 @@ class HasSevereConcomitantIllness: EvaluationFunction {
             return EvaluationFactory.warn("Potential severe concomitant illnesses (WHO $whoStatus)")
         }
         return if (whoStatus == 5) {
-            EvaluationFactory.pass("Has severe concomitant illness (WHO 5)")
+            EvaluationFactory.pass("WHO 5")
```
cbruel, line 17: I wouldn't change this message to only `WHO 5`, I think this is a too vague? Would just keep it as is?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
```diff
@@ -16,7 +16,7 @@ class HasMaximumWHOStatus(private val maximumWHO: Int) : EvaluationFunction {
 
             who <= maximumWHO -> EvaluationFactory.pass("WHO $who is within requested max WHO $maximumWHO")
 
-            who - maximumWHO == 1 -> EvaluationFactory.recoverableFail("WHO $who exceeds max WHO $maximumWHO")
```
cbruel, line 19: I think `recoverableFail` is more suitable here since it's actually a fail just looking at the data (WHO is too high) but we say it's recoverable since it could maybe change?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasWHOStatus.kt
```diff
@@ -20,7 +20,7 @@ class HasWHOStatus(private val requiredWHO: Int) : EvaluationFunction {
             }
 
             abs(who - requiredWHO) == 1 -> {
-                EvaluationFactory.recoverableFail("WHO status is $who but should be $requiredWHO")
```
cbruel, line 23: Same here, I like `recoverableFail` better here.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
```diff
@@ -8,31 +8,26 @@ import com.hartwig.actin.datamodel.algo.EvaluationResult
...
-                isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
-            )
-        }
+        val evaluation = function.evaluate(record)
+        return when (evaluation.result) {
+            EvaluationResult.PASS -> {
+                Evaluation(
+                    result = EvaluationResult.WARN,
+                    recoverable = evaluation.recoverable,
+                    warnMessages = evaluation.passMessages
```
cbruel, line 17: Why did you remove `inclusionMolecularEvents = emptySet(), exclusionMolecularEvents = emptySet() and isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation` here?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
```diff
@@ -8,31 +8,26 @@ import com.hartwig.actin.datamodel.algo.EvaluationResult
...
+                )
+            }
+
+            EvaluationResult.WARN -> evaluation.copy(inclusionMolecularEvents = emptySet(), exclusionMolecularEvents = emptySet())
```
cbruel, line 21: Why did you remove `isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation` here?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
```diff
@@ -8,31 +8,26 @@ import com.hartwig.actin.datamodel.algo.EvaluationResult
...
-            exclusionMolecularEvents = emptySet(),
-            passMessages = (evaluation.passMessages + evaluation.warnMessages + evaluation.undeterminedMessages + evaluation.failMessages),
-            isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
-        )
+            else -> {
+                Evaluation(
+                    result = EvaluationResult.PASS,
+                    recoverable = evaluation.recoverable,
+                    passMessages = (evaluation.passMessages + evaluation.warnMessages + evaluation.undeterminedMessages + evaluation.failMessages),
+                    isMissingMolecularResultForEvaluation = evaluation.isMissingMolecularResultForEvaluation
```
cbruel, line 28: Why did you remove `inclusionMolecularEvents = emptySet(), exclusionMolecularEvents = emptySet()` here?

cbruel (commented):
Looks good, agree with everything you wrote down in the jira ticket! A few small comments. 

I had the impression though with what we wrote down in the Q1 planning (reviewing/revamping the evaluation datamodel) we were going to make bigger changes to the datamodel. So I think it would be good to discuss with Nina / Korneel, how they envisioned this Q task and if it is now 'completed' :) (What you did in this ticket all looks good, it's just not what I had in mind)
I mean things like:
- WARN_IF(AND(..)) currently leads to vague messages (e.g. in case WARN_IF(AND(NOT_EVALUATED, PASS)) only the PASS message is shown on the report, which doesn't make clear what we want to warn for). A solution would be to be able to pass a message in the WARN_IF, which should be shown on the report, so we can tune the message for that specific WARN_IF case.
- For OR(..) you could say it's valid for OR(UNDETERMINED, WARN) to only show the UNDETERMINED message (you don't have to show all messages, since only 1 has to be true), but for AND(..) it would actually maybe make sense to show all messages since all these criteria should be true (currently we drop messages in And)
- Create a whole new datamodel for EvaluationResult where it could actually maybe be a boolean instead of an enum (PASS: true, WARN: true, FAIL: false, UNDETERMINED: null, NOT_EVALUATED: null). With an extra boolean isReliable, where WARN's have a lower reliability than PASS. And keeping the isRecoverable boolean

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/othercondition/HasSevereConcomitantIllness.kt
spdeleeuw: WHO 5 means the patient is dead. So this evaluation doesn't make sense at all and will never be reached (I kept it in to remain support for all possible WHO-statuses). I removed the "Has severe concomitant illness" part because that seems silly to leave in.

cbruel: Ahhh oops sorry makes total sense now. Thanks for catching this and agree with your change

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
spdeleeuw: Because for the first two you mention that's already the default argument in the constructor.  And pass itself doesn't take `isMissingMolecularResultForEvaluation` as an argument so this will always be false (the default). Fine by me to put that last one back though, but not a fan of leaving the default arguments in as explicit arguments.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
spdeleeuw: Because AFAIK with `.copy` that field is automatically populated with the original value (thus redundant to keep that in)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/composite/WarnIf.kt
spdeleeuw: Because for the first two you mention that's already the default argument in the constructor.

spdeleeuw:
@cbruel @ your comments above: my assumption was that we'd review the current implementations. I wasn't aware of suggestions like you state above. I'll check and follow-up if necessary.

PR #1038, opened 2025-03-18
ACTIN-1798: Add HAS_GENE_X_TO_CENTROMERE_OF_CHROMOSOME_Y_RATIO_OF_AT_LEAST_Z_BY_FISH

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasSufficientGeneToCentromereRatioByFish.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.hartwig.actin.algo.evaluation.EvaluationFactory
+import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+
+class HasSufficientGeneToCentromereRatioByFish(private val gene: String, private val chromosome: Int) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        return EvaluationFactory.undetermined(
+            "$gene to centromere of chromosome $chromosome ratio by FISH undetermined",
```
spdeleeuw, line 12: Not sure if I like this implementation. We rarely receive FISH data (and if we did we wouldn't receive the actual data needed here) so this will result in this trial always ending up in the "additional tests needed" table, except when there is an actual MET ex14sk or MET act mut.

PR #1042, opened 2025-03-19
ACTIN-1804 Make the tested genes more dynamic and use "no mutation found"

pauldwolfe:
No mutations will be a combination of the implied no mutation (difference between provided list and what is actually
in the results), and test results explicitly marked as no mutation.

I would suggest for the future we remove the tested genes list from the ProvidedMolecularTest model and make it
purely dynamic based on test results. That will remove the duplication in finding the implied no mutation, and avoid any
possible discrepancies between the tested gene list and what is actually specified in the results.

### DIFF
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
@@ -1,20 +1,20 @@
 package com.hartwig.actin.clinical.feed.standard.extraction
 
 import com.hartwig.actin.clinical.ExtractionResult
-import com.hartwig.actin.datamodel.clinical.ingestion.CurationCategory
 import com.hartwig.actin.clinical.curation.CurationDatabase
 import com.hartwig.actin.clinical.curation.CurationResponse
 import com.hartwig.actin.clinical.curation.config.SequencingTestConfig
 import com.hartwig.actin.clinical.curation.extraction.CurationExtractionEvaluation
-import com.hartwig.actin.datamodel.clinical.provided.ProvidedMolecularTest
-import com.hartwig.actin.datamodel.clinical.provided.ProvidedMolecularTestResult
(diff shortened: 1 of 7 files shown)


clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
```diff
@@ -159,7 +159,7 @@ class StandardPriorIHCTestExtractorTest {
     }
 
     @Test
-    fun `Should extract and curate IHC lines from molecular test list (new)`() {
+    fun `Should extract and curate IHC lines from molecular test ihc result`() {
```
spdeleeuw, line 162: Should we also add a test for ignoring IHC input using `PATIENT | freeText` ?

spdeleeuw (approved):
Looks good to me

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
```diff
@@ -61,6 +63,14 @@ class StandardPriorSequencingTestExtractor(val curation: CurationDatabase<Sequen
         }
     }
 
+    private fun noMutations(providedTestedGenes: Set<String>, allResults: Set<ProvidedMolecularTestResult>): Set<String> {
+        val impliedNoMutations =
+            providedTestedGenes - (allResults.map { it.gene } + allResults.map { it.deletedGene } + allResults.map { it.fusionGeneUp } + allResults.map { it.fusionGeneDown } + allResults.map { it.amplifiedGene }).filterNotNull()
```
cbruel, line 68: I don't think `allResults.map { it.gene }` is correct here since this contains noMutationsFound == true, false and null? I think this should also be the noMutationsFound != true ones?

pauldwolfe: indeed, nice catch!

pauldwolfe:
Seem that the PR was merged

PR #1045, opened 2025-03-24
ACTIN-1814: Prevent empty IHC molecular test instance from showing up…

spdeleeuw:
… in molecular summary and details

This fixes the issue where ALK or ROS1 IHC results are displayed in both the IHC results section as well as empty IHC "molecular test" instances (`IHC (Date unknown) No mutations found`) in molecular summary/details.

Background on why ALK/ROS1 IHC is considered a panel molecular test -> https://github.com/hartwigmedical/actin/pull/817/files

Works well in beta-testing.

### DIFF
--- a/report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
+++ b/report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
@@ -1,7 +1,7 @@
 package com.hartwig.actin.report.pdf.chapters
 
+import com.hartwig.actin.datamodel.molecular.ExperimentType
 import com.hartwig.actin.datamodel.molecular.MolecularRecord
-import com.hartwig.actin.datamodel.molecular.PanelRecord
 import com.hartwig.actin.report.datamodel.Report
 import com.hartwig.actin.report.interpretation.InterpretedCohort
 import com.hartwig.actin.report.interpretation.InterpretedCohortFactory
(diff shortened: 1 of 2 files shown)


report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
```diff
@@ -74,7 +74,7 @@ class MolecularDetailsChapter(
         } ?: orangeMolecularTable.addCell(Cells.createContent("No OncoAct WGS and/or Hartwig NGS panel performed"))
         document.add(orangeMolecularTable)
 
-        val externalPanelResults = report.patientRecord.molecularHistory.molecularTests.filterIsInstance<PanelRecord>()
+        val externalPanelResults = report.patientRecord.molecularHistory.molecularTests.filter { it.experimentType == ExperimentType.PANEL }
```
jbartletthmf, line 77: Thanks for the context.

Your changes make sense given that we have IHC tests in our panel records, but would it be better to represent these with a new type if they're truly different? If they're meant to be equivalent (and the `experimentType` is the proper way to distinguish between records), should we rename `PanelRecord` to something more general?

jbartletthmf (approved):

kzuberihmf (approved):
FWIW, Nina's harmonization of molecular interpretation proposal suggests reworking this to remove IHC from the clinical part of the model. So I assume further improvements when that's put into action.

PR #1048, opened 2025-03-25
ACTIN-1817: Prevent StandardPriorIHCTestExtractor from creating curat…

spdeleeuw:
…ion warnings for all molecular tests with non-null freeText

Fixes a bug caused in https://github.com/hartwigmedical/actin/pull/1042/files#diff-c370ea1349fd88bd4346bcf9a473ace630fd7c6da86ed1bf97e5a4ca58bda3c8

Intended logic:
- Being able to ignore an IHC entry from input feed using either it's ihcResult or freeText
- Never creating an IHC missing curation warning for entries with ihcResult null
- Do not create a curation warning for freeText if it's corresponding ihcResult is curated


cbruel:
@spdeleeuw We could close this one I think? Because Paul already merged the other ACTIN-1817 PR

PR #1050, opened 2025-03-26
ACTIN-1817 Ignore curation warnings from free text for IHC

pauldwolfe:
Alternative to https://github.com/hartwigmedical/actin/pull/1048

Use an empty extraction evaluation when combining the free text curation with the others.

### DIFF
--- a/clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
+++ b/clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
@@ -71,12 +71,7 @@ class StandardPriorIHCTestExtractorTest {
 
     @Test
     fun `Should ignore lines if ignored in curation`() {
-        every { molecularTestCuration.find(IHC_LINE) } returns setOf(
-            IHCTestConfig(
-                input = IHC_LINE,
-                ignore = true
-            )
-        )
+        returnIgnoreFromCurationDB(IHC_LINE)
         val result =
             extractor.extract(EHR_PATIENT_RECORD_WITH_PATHOLOGY)
         assertThat(result.extracted).isEmpty()
@@ -159,7 +154,7 @@ class StandardPriorIHCTestExtractorTest {
     }
 
     @Test
-    fun `Should extract and curate IHC lines from molecular test ihc result and free text`() {
+    fun `Should extract and curate IHC lines from molecular test ihc result`() {
         every { molecularTestCuration.find(IHC_LINE) } returns setOf(
             IHCTestConfig(
                 input = IHC_LINE,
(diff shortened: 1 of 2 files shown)


clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
```diff
@@ -185,6 +194,13 @@ class StandardPriorIHCTestExtractorTest {
...
+        assertThat(result.evaluation.warnings).containsExactly(
+            CurationWarning(
+                patientId = HASHED_ID_IN_BASE64,
+                category = CurationCategory.MOLECULAR_TEST_IHC,
+                feedInput = uncuratedInput,
+                message = "Could not find molecular test config for input 'uncurated'"
+            )
+        )
     }
 }
```
spdeleeuw, line 206: Maybe add this test from my PR ?
```
@Test
    fun `Should ignore lines if ignored in curation using free text`() {
        val ihcResultWithFreeText = ProvidedMolecularTestResult(ihcResult = IHC_LINE, freeText = "some text")
        val record = EHR_PATIENT_RECORD.copy(
            molecularTests = listOf(
                ProvidedMolecularTest(
                    test = "test",
                    results = setOf(ihcResultWithFreeText)
                )
            )
        )

        every { molecularTestCuration.find("some text") } returns setOf(
            IHCTestConfig(
                input = "some text",
                ignore = true
            )
        )

        val result = extractor.extract(record)
        assertThat(result.extracted).isEmpty()
        assertThat(result.evaluation.warnings).isEmpty()
    }
```

spdeleeuw (approved):
Looks good to me! One suggestion

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
pauldwolfe: I've added a similar test, which checks that you can ignore an entry via the ihcResult. Ignoring the free text is not necessary as we discard the curation warnings.

pauldwolfe:
@cbruel @ninajacobs After discussing the missing test with @spdeleeuw we decided that it would be better to ignore erroneous IHC results like we had with patient 88 with a combined "patient | ihc string" curation input.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorIHCTestExtractorTest.kt
pauldwolfe: disregard my comment here, as I'd forgotten the original intent was to be able to ignore on freeText when the ihcResult is erroneous for the patient, but not for future patients.

spdeleeuw (approved):

PR #1052, opened 2025-03-26
ACTIN-1785 Remove split of IHC results. 

andreia-hartwig:


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -41,7 +41,7 @@ object QuestionnaireExtraction {
             otherOncologicalHistory = toList(value(lines, mapping[QuestionnaireKey.OTHER_ONCOLOGICAL_HISTORY])),
             secondaryPrimaries = secondaryPrimaries(lines, mapping[QuestionnaireKey.SECONDARY_PRIMARY]),
             nonOncologicalHistory = toList(value(lines, mapping[QuestionnaireKey.NON_ONCOLOGICAL_HISTORY])),
-            ihcTestResults = toList(value(lines, mapping[QuestionnaireKey.IHC_TEST_RESULTS])),
+            ihcTestResults = value(lines, mapping[QuestionnaireKey.IHC_TEST_RESULTS])?.let { listOf(cleanAndTrim(it)) },
```
jbartletthmf, line 44: What is this change going to do for past curations? We have hundreds - are you planning to rewrite them without breaks for separators?

Also, is this a step towards eliminating the mapping/correction of raw questionnaire entries (which I believe was added in part to remove problematic punctuation)?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -120,6 +120,10 @@ object QuestionnaireExtraction {
         return trimmed
     }
 
+    private fun cleanAndTrim(value: String): String {
```
jbartletthmf, line 123: It would help to name this distinctly from `cleanAndTrim` since it does something different. How about `trimWhitespaceAndCommas`?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -120,6 +120,10 @@ object QuestionnaireExtraction {
...
+        return value.trim { it <= ' ' || it == ',' }
```
jbartletthmf, line 124: Do we expect some inputs to start/end with a comma? This should ideally be covered in testing.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
andreia-hartwig: @jbartletthmf All past curations will have to be adjusted. It will be quite a bit of work, this is something I want to discuss with @cbruel tomorrow. 
This is being done because the current split on comma can be error prone and lead to loss of information. 

See full description of the problem in ACTIN-1785.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
andreia-hartwig: @jbartletthmf When multiple lines are used to represent the ICH results in the questionnaire we might have the result starting with a comma. When splitting on comma this is handled automatically, as empty lines are dropped, but without the split we need to handle it. We could just keep the comma but I though this would be cleaner and make curation easier to read. 

I will cover this with a test.

jbartletthmf (approved):

PR #1053, opened 2025-03-26
ACTIN-1821: Review trial datamodel and improve trial matching chapter display

spdeleeuw:
See JIRA ticket for intended sorting/filtering logic: https://hartwigmedical.atlassian.net/jira/software/projects/ACTIN/boards/45?selectedIssue=ACTIN-1821
I've kept ACTIN and External trials separate in the datamodel, but combined them into the same table, using the original ACTIN table generators.

Couple of things I couldn't fix:
- `Last row is not completed. Table bottom border may collapse as you do not expect it` warning. This was partly due to the fact that external trials only had 3 items in the subtable (cohort, actin event, sites), leaving the warning column empty. Fixing this issue still leaves 2 warnings in (for test trials KRAS-TRIAL-DE and EGFR-BE).
- `The table continues on the next page` sometimes being printed over text in `Cohort` column. -> I tried using setPaddingBottom on the table or setPaddingTop on the footer, but that didn't do much.
- Cell content being duplicated when it doesn't fix on one page -> on first + second page. This is frequently combined with the "table continues" footer issue.

Can one of the reviewers help?

NB. There's a beta branch ready (1.733.0-beta.9) if you'd like to see reports with real patient data.


system/src/test/kotlin/com/hartwig/actin/system/regression/ReportRegressionTest.kt
```diff
@@ -65,7 +65,9 @@ class ReportRegressionTest(private val exampleName: String) {
             )
         )
 
-        assertThat(logLevelRecorder.levelRecorded(Level.WARN) || logLevelRecorder.levelRecorded(Level.ERROR))
+//        assertThat(logLevelRecorder.levelRecorded(Level.WARN) || logLevelRecorder.levelRecorded(Level.ERROR))
+//            .withFailMessage("There are errors or warnings in the logs").isFalse()
+        assertThat(logLevelRecorder.levelRecorded(Level.ERROR))
```
spdeleeuw, line 70: Need to solve this warning: `[WARN ] - Last row is not completed. Table bottom border may collapse as you do not expect it`

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/EligibleTrialTableGenerator.kt
```diff
@@ -0,0 +1,186 @@
...
+                it.isPotentiallyEligible && it.isOpen && it.isMissingMolecularResultForEvaluation!!
+            }
+            val recruitingAndEligibleTrials = recruitingAndEligibleCohorts.map(InterpretedCohort::trialId).distinct()
+            val cohortFromTrialsText = when {
+                recruitingAndEligibleCohorts.isNotEmpty() -> "(${
+                    formatCountWithLabel(
+                        recruitingAndEligibleCohorts.size,
+                        "cohort"
+                    )
+                }" + " from ${formatCountWithLabel(recruitingAndEligibleTrials.size, "trial")})"
```
jbartletthmf, line 120: I think it would be better to define vals for `numCohorts` and `numTrials` so the string can be written on one line. The `+` isn't doing anything here unless it's being used to continue a string on the next line.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/ExternalTrialFunctions.kt
```diff
@@ -26,10 +18,10 @@ object EligibleExternalTrialGeneratorFunctions {
         } else {
             val hospitals = homeCountries.first().hospitalsPerCity.flatMap { it.value }
             val cities = homeCountries.first().hospitalsPerCity.keys
-            val hospitalsString = if (hospitals.size > 10) {
+            val hospitalsString = if (hospitals.size > 3) {
```
jbartletthmf, line 21: It would be better to use a constant in all these places.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/IneligibleTrialTableGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+    private val allowDeEmphasis: Boolean
+) : TrialTableGenerator {
+
+    override fun title(): String {
+        return title
+    }
+
+    override fun contents(): Table {
+        val table = Tables.createFixedWidthCols(trialColWidth, subTableWidths.sum())
+        val headers = sequenceOf("Trial", "Cohort", "Molecular", "Sites", "Ineligibility reasons".takeIf { includeIneligibilityReasonCol })
```
jbartletthmf, line 32: could use `listOfNotNull`

system/src/test/kotlin/com/hartwig/actin/system/regression/ReportRegressionTest.kt
```diff
@@ -14,7 +14,7 @@ import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
 import org.junit.runners.Parameterized.Parameters
 import java.time.LocalDate
-import java.util.Locale
+import java.util.*
```
jbartletthmf, line 17: star import

system/src/test/kotlin/com/hartwig/actin/system/regression/ReportRegressionTest.kt
jbartletthmf: Can you narrow it down to a single table by commenting them out until the warning goes away?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/EligibleTrialTableGenerator.kt
```diff
@@ -22,23 +22,22 @@ class EligibleTrialsGenerator(
...
+        val subTableWidths = listOfNotNull(cohortColWidth, molecularEventColWidth, locationColWidth, (checksColWidth ?: 0f)).toFloatArray()
+        val table = Tables.createFixedWidthCols(trialColWidth, subTableWidths.sum())
+        val headers = sequenceOf("Trial", "Cohort", "Molecular", "Sites", "Warnings".takeIf { cohorts.isNotEmpty() }).filterNotNull()
         if (cohorts.isNotEmpty() || externalTrials.isNotEmpty()) {
-            table.addHeaderCell(Cells.createContentNoBorder(Cells.createHeader("Trial")))
-            val headerSubTable = Tables.createFixedWidthCols(*widths)
-            sequenceOf("Cohort", "Molecular", "Sites", "Warnings").map(Cells::createHeader).forEach(headerSubTable::addHeaderCell)
+            table.addHeaderCell(Cells.createContentNoBorder(Cells.createHeader(headers.first())))
+            val headerSubTable = Tables.createFixedWidthCols(*subTableWidths)
+            headers.drop(1).map(Cells::createHeader).forEach(headerSubTable::addHeaderCell)
```
jbartletthmf, line 40: If `headers` has its size depending on the presence of cohorts, I would expect `subTableWidths` to do the same. Otherwise, the number of elements won't match the number of columns here.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/EligibleTrialTableGenerator.kt
```diff
@@ -117,11 +136,17 @@ class EligibleTrialsGenerator(
...
+        ): EligibleTrialTableGenerator {
             val unavailableAndEligible = cohorts.filter { trial: InterpretedCohort -> trial.isPotentiallyEligible && !trial.isOpen }
             val title =
                 "Trials and cohorts that are potentially eligible, but are closed (${unavailableAndEligible.size})"
-            return create(unavailableAndEligible, emptySet(), requestingSource, null, title, contentWidth, null)
+            return create(unavailableAndEligible, emptySet(), requestingSource, null, title, contentWidth, null, false)
+        }
+
+        fun forFilteredTrials(trials: Set<ExternalTrialSummary>, homeCountry: Country, contentWidth: Float
+        ): EligibleTrialTableGenerator {
```
jbartletthmf, line 147: Unusual line break; if this doesn't fit on one line it should be:
```suggestion
        fun forFilteredTrials(
            trials: Set<ExternalTrialSummary>, homeCountry: Country, contentWidth: Float
        ): EligibleTrialTableGenerator {
```

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
```diff
@@ -27,36 +29,46 @@ object TrialGeneratorFunctions {
...
+            val content = if (contentList.size < tableWidths.size) contentList + "" else contentList
+            content.map(subContentFunction).forEach(trialSubTable::addCell)
+            table.addCell(createContent(wrapSubTable(trialSubTable)))
         }
     }
 
+    private fun externalTrialLocation(trial: ExternalTrialSummary, homeCountry: Country?): String {
+        return homeCountry?.let {
+            val hospitalsToCities = hospitalsAndCitiesInCountry(trial, it)
+            if (homeCountry == Country.NETHERLANDS) hospitalsToCities.first else hospitalsToCities.second
```
jbartletthmf, line 68:
```suggestion
            val (hospitals, cities) = hospitalsAndCitiesInCountry(trial, it)
            if (homeCountry == Country.NETHERLANDS) hospitals else cities
```

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/ExternalTrialFunctions.kt
```diff
@@ -26,10 +18,10 @@ object EligibleExternalTrialGeneratorFunctions {
...
                 MANY_PLEASE_CHECK_LINK
             } else hospitals.joinToString { it.name }
-            val citiesString = if (cities.size > 8) {
+            val citiesString = if (cities.size > 3) {
                 MANY_PLEASE_CHECK_LINK
             } else cities.joinToString { it }
             Pair(hospitalsString, citiesString)
```
jbartletthmf, line 27: Alternatively:
```kotlin
            val (hospitals, cities) = homeCountries.first().hospitalsPerCity
                .let { listOf(it.keys, it.values.flatten().map(Hospital::name)) }
                .map { if (it.size > MAX_TO_DISPLAY) MANY_PLEASE_CHECK_LINK else it.joinToString() }
            hospitals to cities
```

jbartletthmf (commented):
I found the changes much easier to follow when I filtered to the commits after some of the files were renamed: https://github.com/hartwigmedical/actin/pull/1053/files/0ab4e8b8f5d1ecfabe470aecfa201c0f40dfd57b..2d086989211c10422c0ad8f28593ed64430f6ac9#diff-54d9ee7ef64f55dd3571dcf8c19c0bb06a0d2c6cf61575debacbc1ed82084a61R19-R27

system/src/test/kotlin/com/hartwig/actin/system/regression/ReportRegressionTest.kt
nedleitch: Based on my reading of our source and the source for our version of `itext` I think this block is causing the warning:

```
        // check if the last row is incomplete
        if (tableModel.isComplete() && !tableModel.isEmpty()) {
            CellRenderer[] lastRow = rows.get(rows.size() - 1);
            int lastInRow = lastRow.length - 1;
            while (lastInRow >= 0 && null == lastRow[lastInRow]) {
                lastInRow--;
            }
            if (lastInRow < 0 || lastRow.length != lastInRow + (int) lastRow[lastInRow].getPropertyAsInteger(Property.COLSPAN)) {
                Logger logger = LoggerFactory.getLogger(TableRenderer.class);
                logger.warn(IoLogMessageConstant.LAST_ROW_IS_NOT_COMPLETE);
            }
        }
```

There is only one other block which both references the `LAST_ROW_IS_NOT_COMPLETE` constant and isn't a test. So I think the library is trying to tell you that either a cell or a colspan is missing.

You should be able to isolate this pretty quickly I would think with some reasonably-chosen breakpoints. I'd start by suspecting that bit that @jbartletthmf called out above where the columns might not match the headers.

Looked it at it for only a few minutes, you could add some breakpoints in the iText class `RootElement`, I think most rendering calls go through the `createAndAddRendererSubTree` method there.

nedleitch (changes requested):

system/src/test/kotlin/com/hartwig/actin/system/regression/ReportRegressionTest.kt
spdeleeuw: Thanks @nedleitch and @jbartletthmf! There were two issues:
- localExternalTrialGenerator in TrialMatchingChapter -> since localTrials == true in this function, checksColWidth was not null in `EligibleTrialTableGenerator`. Fixed this now with an extra `if (cohorts.isEmpty)` check.
- the declaration of subTableWidths (in `EligibleTrialTableGenerator`) contained `checksColWidth ?: 0f` -> this meant that checksColWidth wasn't null anymore and thus part of the list when it should've been filtered.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/EligibleTrialTableGenerator.kt
```diff
@@ -0,0 +1,186 @@
...
+            requestingSource: TrialSource?,
+            homeCountry: Country? = null,
+            width: Float,
+            localTrials: Boolean = true
+        ): EligibleTrialTableGenerator {
+            val recruitingAndEligibleCohorts = TrialsProvider.filterCohortsAvailable(cohorts)
+            val recruitingAndEligibleTrials = recruitingAndEligibleCohorts.map(InterpretedCohort::trialId).distinct()
+            val cohortFromTrialsText = if (recruitingAndEligibleCohorts.isNotEmpty() || externalTrials.isNotEmpty()) {
+                "(${formatCountWithLabel(recruitingAndEligibleCohorts.size + externalTrials.size, "cohort")} " +
+                        "from ${formatCountWithLabel(recruitingAndEligibleTrials.size + externalTrials.size, "trial")})"
```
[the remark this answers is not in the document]
spdeleeuw: The helper function `formatCountWithLabel` already adds an `s` if needed

report/src/main/kotlin/com/hartwig/actin/report/pdf/ReportContentProvider.kt
```diff
@@ -161,47 +149,55 @@ class ReportContentProvider(private val report: Report, private val enableExtend
...
-                        cohorts, report.requestingHospital, contentWidth, enableExtendedMode
-                    ).takeIf {
-                        report.config.includeIneligibleTrialsInSummary
-                    })
+            }) + trialTableGenerators
     }
 
-    private fun getGeneratorsForSource(
-        cohorts: List<InterpretedCohort>,
+    private fun getGenerators(
```
kduyvesteyn, line 155: Minor but would rename to `createGenerators` as your function actually does something. "Get" typically implies a trivial lookup

report/src/main/kotlin/com/hartwig/actin/report/pdf/ReportContentProvider.kt
```diff
@@ -161,47 +149,55 @@ class ReportContentProvider(private val report: Report, private val enableExtend
...
         requestingSource: TrialSource?,
-        source: TrialSource? = requestingSource,
-        contentWidth: Float,
-        includeLocation: Boolean = false,
-    ): List<EligibleActinTrialsGenerator?> {
-        val openCohortsWithSlotsGenerator = EligibleActinTrialsGenerator.forOpenCohorts(
-            cohorts, source?.description, contentWidth, slotsAvailable = true, includeLocation = includeLocation
+        contentWidth: Float
+    ): List<TrialTableGenerator?> {
+        val homeCountry = report.config.countryOfReference
```
kduyvesteyn, line 161: Wouldn't define this val here. Related, `homeCountry` should be renamed to `countryOfReference` everywhere but can be considered a side project

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/TrialMatchingChapter.kt
```diff
@@ -36,62 +36,62 @@ class TrialMatchingChapter(
 
     private fun addTrialMatchingOverview(document: Document) {
         val table = Tables.createSingleColWithWidth(contentWidth())
-        addGenerators(createGenerators(), table, false)
+        addGenerators(getGenerators(), table, false)
```
kduyvesteyn, line 39: Same, would rename to `createGenerators`

kduyvesteyn (commented):
Just a bunch of minor comments. Want to test more on actual patients + test data

kduyvesteyn:
@spdeleeuw Tables wrapping poorly over multiple pages usually has to do with `keepTogether` setting on cells/rows/tables. If you have an actual example then I'd take that and start experimenting with that setting on various levels, maybe the code isn't enforced to keep together certain parts that shouldn't be wrapped over multiple pages.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/ExternalTrialFunctions.kt
```diff
@@ -16,21 +18,16 @@ object ExternalTrialFunctions {
...
         }
     }
 
     fun countryNamesWithCities(externalTrial: ExternalTrialSummary): String {
         return externalTrial.countries.joinToString { country ->
-            val cities = if (country.hospitalsPerCity.keys.size > 8) {
+            val cities = if (country.hospitalsPerCity.keys.size > MAX_TO_DISPLAY) {
                 MANY_PLEASE_CHECK_LINK
             } else {
                 country.hospitalsPerCity.keys.joinToString(", ")
```
jbartletthmf, line 33: You could consider refactoring with a helper function for displaying collections like this.

jbartletthmf (approved):
Looks good! Just minor comments.

kduyvesteyn (approved):
Looks good! Nice job!

A few thoughts when inspecting real patients, but could be considered out of scope for this ticket:
 - ">3 locations - please check link", to me it feels "3+" would be nicer than ">3" but no super strong opinion.
 - Maybe we should make the "applies to all cohorts below" italic. E.g for a trial that has "part 1" and "part 2" as cohorts it looks a bit weird. Maybe italic would make it stand out more that it is not an actual cohort?
 - The sites can be quite lengthy (esp e.g. DRUP running), should be consider condensing to e.g. "3+ hospitals" or something? Or show the source hospital + "and  3+ others")? Related, the "sites" take up a lot of space in the ineligible table and are likely never relevant. I think it would be cleaner to remove that column for ineligible trials (see DRUP as the best example of how much space is wasted).

ninajacobs:
Nice work Simon!

I also had a look at a report, some (small) thoughts:
- I'd prefer 'Trials in NL' over 'Trials in The Netherlands', for some reason it doesn't read smoothly to me
- I'd mention "Trials in NL" explicitly in the second table, it is not readily clear that this is about NL
- I would remove "open cohorts with no slots available in grey" if nothing is grey (only confusing IMO)
- For sites, I'd propose the following:
o If sites does not contain the requesting hospital, then mention all of them
o If sites contains requesting hospital & total sites is 1 or 2, mention the requesting hospital first (then the other, if applicable)
o if sites contains requesting hospital & total sites is 3+, mention the requesting hospital plus e.g. "2 other sites", or "3 other sites", so if NKI requests, then show "NKI-AvL, 2 other sites", to save up space and don't show irrelevant locations if the trial already runs locally?

- Title should move to next page if the only thing shown on previous page is title itself (but can be done in separate ticket, somewhat recurring issue)

ninajacobs:
- in Ineligible table, we need more room for 'ineligibility reasons' (but can be separate ticket)
- International table: shall we replace "- please check" by ", see" to reduce text?

spdeleeuw:
@ninajacobs 
- "open cohorts with no slots available in grey" is only shown when there are cohorts without slots (if you use the latest beta release to test).
- I've implemented the locations as Korneel suggested -> max 3, regardless if one of those is the requesting site, but with showing the requesting site "+ x others" when the requesting site is one of them
- NL instead of The Netherlands -> Then we'd need to change the display of Country.NETHERLANDS to NL. That would make it inconsistent with other countries. Fine by me though but that's something to keep in mind.
- Sticking the title to the table -> that's not trivial and takes quite some changes in how we currently build tables. I've discussed this in the past with @jbartletthmf and decided to not implement that at the time. I think that's not in scope of this PR and would prefer to keep that for a separate ticket.

PR #1054, opened 2025-03-26
ACTIN-1818: Extract latest non-empty questionnaire when available

zoehehartwig:
I added a check in FeedRecord to get the latest questionnaire that doesn't have an empty text field and check before adding a validation warning that the registration date is within the last year. Does this sound good? I know there was some debate on what to do about the unintended missing questionnaire warnings.


jbartletthmf:
@zoehehartwig can you update the PR title to match convention?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
```diff
@@ -187,14 +188,14 @@ class EmcClinicalFeedIngestor(
         val ingestionStatus =
             if (questionnaire == null || curationResults.isNotEmpty()) PatientIngestionStatus.WARN else PatientIngestionStatus.PASS
 
-        val validationWarnings = if (questionnaire == null) {
-            feedRecord.validationWarnings + FeedValidationWarning(patientId, "No Questionnaire found")
+        val validationWarnings = if (questionnaire == null && record.patient.registrationDate.isAfter(LocalDate.now().minusMonths(1))) {
```
jbartletthmf, line 191: This task (https://hartwigmedical.atlassian.net/browse/ACTIN-1818) is about selecting a non-empty questionnaire when present, even if it isn't the latest. The discussion about filtering missing-questionnaire warnings for old patients hasn't been captured in any Jira ticket to my knowledge.

When it comes to filtering the warnings, I think it would be better to keep the existing behaviour in ACTIN, i.e. include a warning for every patient that's missing a questionnaire. That will ensure ACTIN behaves deterministically and predictably, even for historic patients.

IMO, the filtering should be applied in actin-pipelines where we decide what to include in the Slack message. Since we don't currently read in the patient record there, we may have to add the registration date to the PatientIngestionResult if that's going to be our criteria for filtering. This change requires a little more design and potentially some cross-repo coordination, so I think it should be handled in a separate ticket. @andreia-hartwig, thoughts?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/FeedRecord.kt
```diff
@@ -47,7 +47,7 @@ data class FeedRecord(
                 medicationEntries,
                 intoleranceEntries,
                 allBodyWeightEntries.distinct(),
-                questionnaireEntries.maxByOrNull(QuestionnaireEntry::authored),
+                questionnaireEntries.filter { it.text.isNotEmpty() }.maxByOrNull(QuestionnaireEntry::authored),
```
jbartletthmf, line 50: I think Andreia explained to you why this filtering is incorrect. In [the ticket](https://hartwigmedical.atlassian.net/browse/ACTIN-1818), we suggested including all questionnaire entries here:
> Since proper questionnaire validation is non-trivial, we suggest propagating all questionnaire entries to the extraction stage and using the latest valid questionnaire entry at that time.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestorTest.kt
```diff
@@ -3,19 +3,25 @@ package com.hartwig.actin.clinical.feed.emc
...
 import com.hartwig.actin.datamodel.clinical.Comorbidity
+import com.hartwig.actin.datamodel.clinical.Gender
 import com.hartwig.actin.datamodel.clinical.TumorDetails
 import io.mockk.every
 import io.mockk.mockk
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import java.time.LocalDate
 
 class EmcClinicalFeedIngestorTest {
```
jbartletthmf, line 17: I expect to see a new test that verifies that a newer, empty questionnaire is discarded in favour of an older, populated one.

jbartletthmf:
@zoehehartwig We don't typically invite more than two reviewers unless the goal is knowledge-sharing for a change with broader impact, and in those cases it's good to clearly designate who should review the code critically and who was invited just as an FYI.

Since @andreia-hartwig and I have already weighed in here, I think @ninajacobs and @spdeleeuw can sit this PR out unless they have a particular interest in these changes.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
andreia-hartwig: @jbartletthmf I agree that the filtering missing-questionnaire warnings should be handled in a separate ticket - I've created ACTIN-1825 for this purpose. 
The solution needs to be clear and handling it in the pipeline may be a good idea, specially we want to connect this to the presence of WGS or not.

jbartletthmf: Thanks Andreia!

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -215,14 +215,30 @@ class QuestionnaireExtractionTest {
...
+                authored = LocalDate.now()
+            )
+        val oldValidEntry = TestQuestionnaireFactory.createTestQuestionnaireEntry().copy(
+            authored = LocalDate.now().minusDays(1),
```
jbartletthmf, line 234: You could set `val today = LocalDate.now()` so you don't have to call this function each time.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -215,14 +215,30 @@ class QuestionnaireExtractionTest {
...
+            text = TestQuestionnaireFactory.createTestQuestionnaireValueV1_7().replace("\n", "\\n")
+        )
+        val extractedEntry = QuestionnaireExtraction.extract(listOf(newInvalidEntry, newEmptyEntry, oldValidEntry))
+        assertThat(extractedEntry.first).isNotNull
+        assertThat(extractedEntry.first!!.hasActiveBrainLesions!!).isEqualTo(true)
```
jbartletthmf, line 239: I think it would be clearer to assign both sides of the pair and assert on each, e.g.:
```suggestion
        val (extractedEntry, errors) = QuestionnaireExtraction.extract(listOf(newInvalidEntry, newEmptyEntry, oldValidEntry))
        assertThat(extractedEntry).isNotNull
        assertThat(extractedEntry!!.hasActiveBrainLesions!!).isEqualTo(true)
        assertThat(errors).isEmpty
```

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,9 +11,20 @@ object QuestionnaireExtraction {
...
-    fun extract(entry: QuestionnaireEntry?): Pair<Questionnaire?, List<QuestionnaireCurationError>> {
+    fun extract(entryList: List<QuestionnaireEntry>): Pair<Questionnaire?, List<QuestionnaireCurationError>> {
+        val extracted = entryList.sortedByDescending(QuestionnaireEntry::authored).firstNotNullOfOrNull { it ->
+            extractQuestionnaire(it)
+        }
+        return if (extracted != null) {
+            extracted
+        } else {
+            null to emptyList()
+        }
```
jbartletthmf, line 22: You can use the elvis operator to handle nulls specifically. I would also use `asSequence` for lazy evaluation (https://typealias.com/guides/when-to-use-sequences/):
```suggestion
        return entryList.asSequence().sortedByDescending(QuestionnaireEntry::authored)
            .firstNotNullOfOrNull(::extractQuestionnaire)
            ?: null to emptyList()
```

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -56,7 +67,43 @@ object QuestionnaireExtraction {
...
+                            hasBrainLesions,
+                            hasActiveBrainLesions,
+                            hasCnsLesions,
+                            hasActiveCnsLesions,
+                            hasBoneLesions.curated,
+                            hasLiverLesions.curated,
+                            whoStatus.curated,
+                            infectionStatus.curated,
+                            ecg.curated
+                        ).all { it == null }
```
jbartletthmf, line 100: I think it would help to put this in a helper function in this class or in a public function on the `Questionnaire`, e.g. `Questionnaire.isEmpty()` (or `Questionnaire.isNotEmpty()` for the negation).

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,9 +11,20 @@ object QuestionnaireExtraction {
...
+        } else {
+            null to emptyList()
+        }
+    }
+
+    fun extractQuestionnaire(entry: QuestionnaireEntry?): Pair<Questionnaire, List<QuestionnaireCurationError>>? {
         if (entry == null || !isActualQuestionnaire(entry)) {
-            return null to emptyList()
+            return null
         }
```
jbartletthmf, line 28: The more idiomatic kotlin way to do this would be:
```kotlin
    fun extractQuestionnaire(entry: QuestionnaireEntry?): Pair<Questionnaire, List<QuestionnaireCurationError>>? {
        return entry?.takeIf(::isActualQuestionnaire)?.let { validEntry ->
            // questionnaire extraction and validation
        }
    }
```

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -56,7 +67,43 @@ object QuestionnaireExtraction {
...
+                            hasCnsLesions,
+                            hasActiveCnsLesions,
+                            hasBoneLesions.curated,
+                            hasLiverLesions.curated,
+                            whoStatus.curated,
+                            infectionStatus.curated,
+                            ecg.curated
+                        ).all { it == null }
+            }) {
+            null
```
jbartletthmf, line 102: Since failed extractions usually result in a `(curated, errors)` pair like `(null, listOf(errorMessage))`, I think there's a real risk that we skip over some newer questionnaires with errors with this approach (assuming we want to be aware of those). I think it would be better to compile the full set of errors first, i.e.
```kotlin
val errors = listOf(hasBoneLesions, hasLiverLesions, hasMeasurableDisease, whoStatus, infectionStatus, ecg, stage).flatMap(ValidatedQuestionnaireCuration::errors)
```
And then return
```kotlin
Pair(questionnaire, errors).takeIf { questionnaire.isNotEmpty() || errors.isNotEmpty() }
```

clinical/src/test/kotlin/com/hartwig/actin/clinical/ClinicalIngestionFeedAdapterTest.kt
```diff
@@ -90,7 +90,7 @@ class ClinicalIngestionFeedAdapterTest {
 
     @Test
     fun `Should run ingestion from proper curation and feed files, read from filesystem`() {
-        assertQuestionnaireInFeedIsOfLatestVersion()
+        //assertQuestionnaireInFeedIsOfLatestVersion()
```
jbartletthmf, line 93: Don't forget to uncomment these tests 🙂

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,9 +11,20 @@ object QuestionnaireExtraction {
...
+            extractQuestionnaire(it)
+        }
+        return if (extracted != null) {
+            extracted
+        } else {
+            null to emptyList()
+        }
+    }
+
+    fun extractQuestionnaire(entry: QuestionnaireEntry?): Pair<Questionnaire, List<QuestionnaireCurationError>>? {
```
jbartletthmf, line 25: Can this function be made private?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,52 +11,66 @@ object QuestionnaireExtraction {
...
+                ecg = ecg.curated,
+                complications = toList(value(lines, mapping[QuestionnaireKey.COMPLICATIONS])),
+            )
+            val errors = listOf(hasBoneLesions, hasLiverLesions, hasMeasurableDisease, whoStatus, infectionStatus, ecg, stage).flatMap(
+                ValidatedQuestionnaireCuration<*>::errors
+            )
+
+            when {
+                questionnaire.isEmpty() && errors.isEmpty() -> null
+                questionnaire.isEmpty() -> null to errors
```
andreia-hartwig, line 70: is "null to errors " really a valid response here?  If a questionnaire is empty, the errors would be about missing data. This will return a pair with the 1st element null, that will be propagated further in the flow, and instead we should take the next questionnaire that is not empty.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,52 +11,66 @@ object QuestionnaireExtraction {
...
+                hasBoneLesions = hasBoneLesions.curated,
+                hasLiverLesions = hasLiverLesions.curated,
+                otherLesions = otherLesions(entry, lines, mapping),
+                whoStatus = whoStatus.curated,
+                unresolvedToxicities = toList(value(lines, mapping[QuestionnaireKey.UNRESOLVED_TOXICITIES])),
+                infectionStatus = infectionStatus.curated,
+                ecg = ecg.curated,
+                complications = toList(value(lines, mapping[QuestionnaireKey.COMPLICATIONS])),
+            )
+            val errors = listOf(hasBoneLesions, hasLiverLesions, hasMeasurableDisease, whoStatus, infectionStatus, ecg, stage).flatMap(
```
andreia-hartwig, line 64: what about brainLesionData and cnsLesionData ?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
jbartletthmf: @andreia-hartwig isn't it better to propagate errors to the user instead of silently skipping them? Otherwise, they could add a new questionnaire with new data and some errors and we would never see it. This was my basis for requesting this change: https://github.com/hartwigmedical/actin/pull/1054#discussion_r2021269684

jbartletthmf: I'm assuming that if Q2 is truly empty, then `errors` should be empty as well and we'll return null: `questionnaire.isEmpty() && errors.isEmpty() -> null`. Otherwise, we would see extraction errors for all empty questionnaires. Often, an extraction error results in a null extracted value, so I worry we'd hide errors if we returned just null for cases where the questionnaire is empty and the errors are not.

andreia-hartwig: @jbartletthmf If errors are always empty when the questionnaire is truly empty then there is no issue. However, since questionnaire.isEmpty() returns true when the questionnaire is truly empty, doesn't that mean it will never really return in the second condition?

jbartletthmf: Looking at https://github.com/hartwigmedical/actin/blob/87b6dc56d98cf044090f3ab7c92b7ef3683231da/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireCuration.kt#L38-L45
and similar functions, we seem to return `Pair(null, errors)` for many cases where there's an extraction error. I think it's possible to produce an empty `Questionnaire` with non-empty errors, though it may be unlikely.

andreia-hartwig: @jbartletthmf This is actually an interesting point, in this case, having an empty questionnaire doesn't actually mean the questionnaire was empty in the feed, it just means none of the validations passed and so all values are null. I wonder if we need to distinguish between those to scenarios - empty in feed vs empty due to failed validations. 

Nevertheless, even in this case, shouldn't we try to see if the next questionnaire on the list of Questionnaire Entries is valid?

jbartletthmf: Going by the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I think a new questionnaire entry with errors should never be skipped over without warning. Otherwise, how would we know that we received one?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/Questionnaire.kt
```diff
@@ -30,4 +30,36 @@ data class Questionnaire(
...
+                    hasBrainLesions,
+                    hasActiveBrainLesions,
+                    hasCnsLesions,
+                    hasActiveCnsLesions,
+                    hasBoneLesions,
+                    hasLiverLesions,
+                    whoStatus,
+                    infectionStatus,
+                    ecg
+                ).all { it == null }
```
jbartletthmf, line 63: This long, multiline expression is a little tough to follow. I think it would be better to define 3 boolean vals, e.g. `allStringsEmpty`, `allListsEmpty`, and `allBooleanAndObjectFieldsNull`, and then return an expression of all of them.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtraction.kt
```diff
@@ -11,52 +11,66 @@ object QuestionnaireExtraction {
...
+                otherLesions = otherLesions(entry, lines, mapping),
+                whoStatus = whoStatus.curated,
+                unresolvedToxicities = toList(value(lines, mapping[QuestionnaireKey.UNRESOLVED_TOXICITIES])),
+                infectionStatus = infectionStatus.curated,
+                ecg = ecg.curated,
+                complications = toList(value(lines, mapping[QuestionnaireKey.COMPLICATIONS])),
+            )
+            val errors = listOf(brainLesionData, cnsLesionData, hasBoneLesions, hasLiverLesions, hasMeasurableDisease, whoStatus, infectionStatus, ecg, stage).flatMap(
+                ValidatedQuestionnaireCuration<*>::errors
+            )
```
jbartletthmf, line 66: I would move `flatMap` to the next line:
```suggestion
            val errors = listOf(brainLesionData, cnsLesionData, hasBoneLesions, hasLiverLesions, hasMeasurableDisease, whoStatus, infectionStatus, ecg, stage)
                .flatMap(ValidatedQuestionnaireCuration<*>::errors)
```

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -8,6 +11,10 @@ import org.junit.Test
 import java.time.LocalDate
 
 class QuestionnaireExtractionTest {
+    private val today = LocalDate.now()
+    private val model: FeedModel = TestFeedFactory.createProperTestFeedModel()
+    private val feedRecord: FeedRecord = model.read().single()
```
jbartletthmf, line 16: Could be inlined:
```suggestion
    private val feedRecord: FeedRecord = TestFeedFactory.createProperTestFeedModel().read().single()
```

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -215,14 +222,41 @@ class QuestionnaireExtractionTest {
...
+        assertThat(extractedEntry!!.hasActiveBrainLesions!!).isEqualTo(true)
+        assertThat(errors).isEmpty()
+    }
+
+    @Test
+    fun `Should be able to determine latest questionnaire`() {
+        val (latest, _) = QuestionnaireExtraction.extract(feedRecord.questionnaireEntries.map {
+            val text = it.text
+            it.copy(text = text.replace("\n", "\\n"))
+        })
```
jbartletthmf, line 255: I think it's cleaner to use a named value instead of `it`:
```suggestion
        val (latest, _) = QuestionnaireExtraction.extract(feedRecord.questionnaireEntries.map { entry ->
            entry.copy(text = entry.text.replace("\n", "\\n"))
        })
```
Also, I think this test would be better if we asserted that the returned errors are empty. Otherwise, it might allow some unexpected behaviour.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/Questionnaire.kt
```diff
@@ -30,4 +30,37 @@ data class Questionnaire(
...
```
jbartletthmf, line 63: Most of these lines appear to be over-indented. Also, it would help readability to have a blank line between these multi-line statements.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -215,14 +220,42 @@ class QuestionnaireExtractionTest {
...
+        val newInvalidEntry = entryWithText("Does not exist").copy(authored = today)
+        val newEmptyEntry =
+            entryWithText("\"ACTIN Questionnaire V1.5\\nImportant: The information in these fields will be automatically extracted from the EHR as part of the ACTIN project. Please make sure that these fields never contain non-anonymized data!\\n\\nRelevant patient history\\nTreatment history current tumor: \\nOther oncological history (e.g. radiotherapy, surgery): \\nSecondary primary: \\n- Last date of active treatment: \\nNon-oncological history: \\n\\nTumor details\\nPrimary tumor location: \\nPrimary tumor type: \\nBiopsy location: \\nStage: \\nCNS lesions: \\n-Active: \\nBrain lesions: \\n-Active: \\nBone lesions: \\nLiver lesions: \\nOther lesions (e.g. lymph node, pulmonal):\\nMeasurable disease: \\n\\nPrevious Molecular tests\\n- IHC test results: \\n- PD L1 test results:\\n\\nClinical details\\nWHO status: \\nUnresolved toxicities grade => 2: \\nSignificant current infection: \\nSignificant aberration on latest ECG: \\nCancer-related complications (e.g. pleural effusion):\\n\\n\"").copy(
```
jbartletthmf, line 235: You should be able to use triple quotes (`"""`) to define a multiline string here as in `TestQuestionnaireFactory` instead of having a very long line.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/questionnaire/QuestionnaireExtractionTest.kt
```diff
@@ -215,14 +220,42 @@ class QuestionnaireExtractionTest {
...
+                authored = today
+            )
+        val oldValidEntry = TestQuestionnaireFactory.createTestQuestionnaireEntry().copy(
+            authored = today.minusDays(1),
+            text = TestQuestionnaireFactory.createTestQuestionnaireValueV1_7().replace("\n", "\\n")
+        )
+        val (extractedEntry, errors) = QuestionnaireExtraction.extract(listOf(newInvalidEntry, newEmptyEntry, oldValidEntry))
+        assertThat(extractedEntry).isNotNull
+        assertThat(extractedEntry!!.hasActiveBrainLesions!!).isEqualTo(true)
```
jbartletthmf, line 244: Could we use safe calls here? I'm not sure if you can make this assertion on a nullable, but if you can I don't think there's a reason to use `!!`.
```suggestion
        assertThat(extractedEntry?.hasActiveBrainLesions).isEqualTo(true)
```

jbartletthmf (approved):

PR #1056, opened 2025-03-27
ACTIN-1827 Updated iText dependency to 1.9.0. 

ArnoHMF:
Required some small changes because API has changed.  Use Italic and Bold Italic fonts as in 9.1.0 the method setBold and setItalic are not supported anymore


jbartletthmf (approved):
The updated reports look good 👍

PR #1058, opened 2025-03-27
ACTIN-1670: Implement HasTnmTScore using tumor stage

zoehehartwig:
Please note in a future commit, I will change the class to take in a TNM T score instead of a string. I will also add TNM T scores into the data model and update the function input. 
Also @spdeleeuw , the chart you gave me doesn't mention tumor stages IIC, IIID, and IVC so I assumed they work the same as their previous stages but I commented out those sections for now
Screenshot 2025-03-19 at 09 23 01


spdeleeuw:
@zoehehartwig Lung cancer doesn't have stages IIC, IIID or IVC. That's why they aren't part of the chart.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,52 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+        val allT = T1 + T2 + TnmT.T3 + TnmT.T4
+        val M1 = setOf(TnmT.M1, TnmT.M1A, TnmT.M1B, TnmT.M1C)
+        val M1AB = setOf(TnmT.M1, TnmT.M1A, TnmT.M1B)
+        val M1C = setOf(TnmT.M1, TnmT.M1C)
+
+        val stageMap = mapOf(
+            TumorStage.I to T1 + T2A,
+            TumorStage.IA to T1,
+            TumorStage.IB to T2A,
+            TumorStage.II to T123,
```
spdeleeuw, line 31: I assume it's still a WIP and you will change this to T1 + T2 + T3?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,52 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+        )
+
+        val scoreTnmT = TnmT.valueOf(score.uppercase())
+
+        val stage = record.tumor.stage
+        return when {
+            stageMap[stage]?.contains(TnmT.M1) == true -> EvaluationFactory.undetermined("Cancer is metastatic. Primary tumor stage is unknown")
+            stageMap[stage]?.contains(scoreTnmT) == true -> EvaluationFactory.pass("Tumor is of stage $score with potential T scores of ${stageMap[stage]}")
+            else -> EvaluationFactory.fail("Tumor is of stage $score")
+        }
```
spdeleeuw, line 53: I'd suggest evaluating as follows:
- Allowing multiple stages to match as argument to this class
- If there's a full match -> all derived tnmT's match to any of the requested T's -> pass
- If there's a partial match -> some derived tnmT's match to any of the requested T's, some of the derived tnmT's do not match -> undetermined
- Tumor stage is IV -> undetermined
- No match (none of the derived tnmT's match the requested T's -> fail

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
zoehehartwig: t123 is defined above as
private val t123 = t1 + t2 + TnmT.T3

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
zoehehartwig: I finished. Give it a look

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,48 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+    private val M1C = setOf(TnmT.M1, TnmT.M1C)
+
+    private val stageMap = mapOf(
+        TumorStage.I to t1 + t2A,
+        TumorStage.IA to t1,
+        TumorStage.IB to t2A,
+        TumorStage.II to t123,
+        TumorStage.IIA to t1 + t2B,
+        TumorStage.IIB to t123,
+        //TumorStage.IIC to T123,
```
spdeleeuw, line 28: You can drop the commented ones here and below?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/TnmT.kt
```diff
@@ -0,0 +1,24 @@
...
+    T1(null),
+    T1A(T1),
+    T1B(T1),
+    T1C(T1),
+    T2(null),
+    T2A(T2),
+    T2B(T2),
+    T3(null),
+    T4(null),
+    M1(null),
```
spdeleeuw, line 16: Strictly speaking the M is not part of the T in **T**NM anymore (it's the M). Maybe it's cleaner to drop the M's here and maybe map the stage IV's to e.g. null? What do you think @jbartletthmf ?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,48 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+    )
 
-class HasTnmTScore(private val score: String): EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        return EvaluationFactory.undetermined("Undetermined if tumor is TNM $score")
+        val stage = record.tumor.stage
+        val possibleTnmTs = stageMap[stage]?: emptySet()
+
+        return when {
+            possibleTnmTs.contains(TnmT.M1) -> EvaluationFactory.undetermined("Cancer is metastatic. Primary tumor stage is unknown")
```
spdeleeuw, line 45: Suggestion for the message: "Undetermined if tumor is TNM T-classification`{insert requested T or T's}`"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,48 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+            possibleTnmTs.containsAll(scores) -> EvaluationFactory.pass("Tumor could be of stages $scores with potential T scores of $possibleTnmTs")
```
spdeleeuw, line 46: Suggestion: "Tumor is of TNM T-classification {requested T's concatenated with `or`}"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,48 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+            possibleTnmTs.intersect(scores).isNotEmpty() -> EvaluationFactory.undetermined("Tumor could be of $possibleTnmTs but can't be of ${scores.subtract(possibleTnmTs)}")
```
spdeleeuw, line 47: Suggestion: "Undetermined if TNM T-classification is {requested T's concatenated with `or`} - derived T's based on tumor stage {derived T's concatenated with `or`}.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -4,9 +4,48 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+            else -> EvaluationFactory.fail("Tumor is not of stage $scores")
```
spdeleeuw, line 48: You can use `Format.concatWithCommaAndOr` to make sure the display of this iterable is easy to read.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScoreTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    fun `Should fail if the score is not possible with the tumor stage`(){
+        assertEvaluation(
+            EvaluationResult.FAIL, function(setOf(TnmT.T2A) , TumorTestFactory.withTumorStage(TumorStage.IIA))
+        )
+    }
+}
+
+fun function(scores: Set<TnmT>, record: PatientRecord) = HasTnmTScore(scores).evaluate(record)
```
spdeleeuw, line 33: This should be private. And usually we declare it as a private val -> `private val function = HasTnmTScore(scores)` -> `function.evaluate(...)`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
zoehehartwig: I don't think that would be accurate since many tumor stages map to multiple T scores.

zoehehartwig: For example, IIIA and IIIB map to all 4 stages

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
 import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.algo.evaluation.util.Format
 import com.hartwig.actin.datamodel.PatientRecord
 import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.TumorStage
+import com.hartwig.actin.datamodel.clinical.TnmT
+
+class HasTnmTScore(private val scores: Set<TnmT>) : EvaluationFunction {
+    private val t1 = setOf(TnmT.T1, TnmT.T1A, TnmT.T1B, TnmT.T1C)
+    private val t2 = setOf(TnmT.T2, TnmT.T2A, TnmT.T2B)
```
jbartletthmf, line 13: Could be assigned as `t2A + t2B`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/TumorRuleMapper.kt
```diff
@@ -171,8 +171,8 @@ class TumorRuleMapper(resources: RuleMappingResources) : RuleMapper(resources) {
 
     private fun hasSpecificTnmTScoreCreator(): FunctionCreator {
         return { function: EligibilityFunction ->
-            val score = functionInputResolver().createOneStringInput(function)
-            HasTnmTScore(score)
+            val scores = functionInputResolver().createManyTnmTInput(function)
+            HasTnmTScore(scores)
```
jbartletthmf, line 175: Should we use `DerivedTumorStageEvaluationFunction` here to handle cases where the specific tumor stage is not known? Actually, it might be even better to write the function to work with derived stages as well (e.g. flat-map each derived stage to all possible T-scores) in order to correctly interpret the full range of possible scores.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
-class HasTnmTScore(private val score: String): EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        return EvaluationFactory.undetermined("Undetermined if tumor is TNM $score")
+        val stage = record.tumor.stage
+        val possibleTnmTs = stageMap[stage] ?: emptySet()
+
+        return when {
+            stage == TumorStage.IV || stage == TumorStage.IVA || stage == TumorStage.IVB || stage == TumorStage.IVC ->
+                EvaluationFactory.undetermined("Cancer is metastatic. Undetermined if tumor is TNM T-classification ${show(scores)}")
+            possibleTnmTs.containsAll(scores) -> EvaluationFactory.pass("Tumor could be T scores of ${show(possibleTnmTs)}")
```
jbartletthmf, line 39: Isn't this backwards? From Simon:
> If there's a full match -> all derived tnmT's match to any of the requested T's -> pass

Here you're checking that all of the requested Ts are contained in the derived Ts, but some derived Ts might be outside the requested set. I think this should be `scores.containsAll(possibleTnmTs)`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
 
 import com.hartwig.actin.algo.evaluation.EvaluationFactory
 import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.algo.evaluation.util.Format
 import com.hartwig.actin.datamodel.PatientRecord
 import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.TumorStage
+import com.hartwig.actin.datamodel.clinical.TnmT
+
+class HasTnmTScore(private val scores: Set<TnmT>) : EvaluationFunction {
```
jbartletthmf, line 11: Calling this e.g. `targetTnmTs`, `tnmTsToFind`, etc. might make the code clearer.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScoreTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    fun `Should be undetermined if the tumor is TNM M`(){
+        assertEvaluation(
+            EvaluationResult.UNDETERMINED, function(setOf(TnmT.T1), TumorTestFactory.withTumorStage(TumorStage.IVA))
+        )
+    }
+
+    @Test
+    fun `Should pass if the score matches the tumor stage`(){
+        assertEvaluation(
+            EvaluationResult.PASS, function(setOf(TnmT.T4) , TumorTestFactory.withTumorStage(TumorStage.IIIA))
```
jbartletthmf, line 21: extra space before comma

common/src/main/kotlin/com/hartwig/actin/trial/input/FunctionInputResolver.kt
```diff
@@ -929,6 +940,18 @@ class FunctionInputResolver(
         }
     }
 
+    private fun toTnmTs(input: Any): Set<TnmT>{
+        return toStringList(input).map(::toTnmT).toSet()
+    }
+
+    private fun toTnmT(tnmT: String): TnmT {
+        try {
+            return TnmT.valueOf(tnmT.uppercase(Locale.getDefault()))
```
jbartletthmf, line 949: We usually omit the `Locale` parameter here. Also, is `uppercase()` even needed? We should expect users to provide these strings in uppercase already.

common/src/main/kotlin/com/hartwig/actin/trial/input/FunctionInputResolver.kt
```diff
@@ -929,6 +940,18 @@ class FunctionInputResolver(
...
+        } catch (e: Exception) {
+            throw IllegalStateException("TNM T not found: $tnmT")
+        }
```
jbartletthmf, line 952: We don't need to catch and re-throw here; the exception from `valueOf` will do (all exceptions in this class are caught at a higher level).

common/src/main/kotlin/com/hartwig/actin/trial/input/FunctionInputResolver.kt
```diff
@@ -929,6 +940,18 @@ class FunctionInputResolver(
         }
     }
 
+    private fun toTnmTs(input: Any): Set<TnmT>{
+        return toStringList(input).map(::toTnmT).toSet()
```
jbartletthmf, line 944: I think you can do this without `toTnmT`:
```suggestion
        return toStringList(input).map(TnmT::valueOf).toSet()
```

common/src/test/kotlin/com/hartwig/actin/trial/input/FunctionInputResolverTest.kt
```diff
@@ -1064,6 +1065,17 @@ class FunctionInputResolverTest {
         assertThat(resolver.hasValidInputs(create(rule, listOf("FGFR2", "FGFR2b", 1)))!!).isFalse
     }
 
+    @Test
+    fun `Should resolve functions with one TNM T as input`() {
```
jbartletthmf, line 1069: Inaccurate test name

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/TnmT.kt
```diff
@@ -0,0 +1,21 @@
+package com.hartwig.actin.datamodel.clinical
+
+import com.hartwig.actin.datamodel.Displayable
+
+enum class TnmT(val category: TnmT?) : Displayable {
```
jbartletthmf, line 5: Do we need `category`? It doesn't seem to be used.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
+                EvaluationFactory.undetermined("Cancer is metastatic. Undetermined if tumor is TNM T-classification ${show(scores)}")
+            possibleTnmTs.containsAll(scores) -> EvaluationFactory.pass("Tumor could be T scores of ${show(possibleTnmTs)}")
+            possibleTnmTs.intersect(scores).isNotEmpty() ->
+                EvaluationFactory.undetermined("Undetermined if TNM T-classification is of ${show(scores)}- derived T's based on tumor stage are ${show(possibleTnmTs)}")
+            else -> EvaluationFactory.fail("Tumor is not of stage ${show(scores)}")
+        }
+    }
+
+    private fun show(set: Set<TnmT>): String{
+        return Format.concatWithCommaAndOr(set.map { it.toString() })
```
jbartletthmf, line 47: Since we have a dedicated `display()` function, we should use that here to ensure any special formatting gets applied. The `Format` object also has some `concatItems...` functions that are meant to work with `Displayable` collections, so we would just have to add a `concatItemsWithCommaAndOr` using the same patterns to simplify the call here.

Otherwise, if `display()` isn't going to be used, there's no need to implement the `Displayable` interface.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
+            stage == TumorStage.IV || stage == TumorStage.IVA || stage == TumorStage.IVB || stage == TumorStage.IVC ->
+                EvaluationFactory.undetermined("Cancer is metastatic. Undetermined if tumor is TNM T-classification ${show(scores)}")
+            possibleTnmTs.containsAll(scores) -> EvaluationFactory.pass("Tumor could be T scores of ${show(possibleTnmTs)}")
+            possibleTnmTs.intersect(scores).isNotEmpty() ->
+                EvaluationFactory.undetermined("Undetermined if TNM T-classification is of ${show(scores)}- derived T's based on tumor stage are ${show(possibleTnmTs)}")
+            else -> EvaluationFactory.fail("Tumor is not of stage ${show(scores)}")
+        }
+    }
+
+    private fun show(set: Set<TnmT>): String{
```
jbartletthmf, line 46: How about `tnmTs` instead of `set`? Unless you meant to generalize this to a `Set<Displayable>` or a `Set<String>` (which doesn't seem necessary here)?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
zoehehartwig: I implemented this but I could see a potential problem of if the user inputs 2A for IB but it would be undetermined instead of pass since they'd need to input 2A and 2

spdeleeuw: Good catch, @jbartletthmf! And I'm not entirely sure what you mean here @zoehehartwig, but I think within the class we should consider the user input valid.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScoreTest.kt
```diff
@@ -30,18 +30,34 @@ class HasTnmTScoreTest {
     }
 
     @Test
-    fun `Should pass if the stage contains all the scores as possibilities`() {
+    fun `Should pass if the targets contains all possible TnmTs of the target`() {
```
jbartletthmf, line 33: "of the patient"?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScoreTest.kt
```diff
@@ -30,18 +30,34 @@ class HasTnmTScoreTest {
...
         assertEvaluation(
-            EvaluationResult.PASS, function(setOf(TnmT.T1A, TnmT.T4, TnmT.T2B) , TumorTestFactory.withTumorStage(TumorStage.IIIA))
+            EvaluationResult.PASS, function(setOf(TnmT.T2, TnmT.T4, TnmT.T2A) , TumorTestFactory.withTumorStage(TumorStage.IB))
         )
     }
 
     @Test
-    fun `Should be undetermined if only some of the scores are possible with the stage`() {
+    fun `Should be undetermined if only some of the stages are possible with the targets`() {
```
jbartletthmf, line 40: This wording isn't clear to me. How about "if only some of the possible TnmTs exist in the target set"?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScoreTest.kt
```diff
@@ -30,18 +30,34 @@ class HasTnmTScoreTest {
...
+    @Test
+    fun `Should use derived stages if stage is null`() {
+        assertEvaluation(
+            EvaluationResult.UNDETERMINED,
+            function(setOf(TnmT.T1A, TnmT.T4, TnmT.T2B), TumorTestFactory.withTumorStageAndDerivedStages(null, setOf(TumorStage.IIB)))
+        )
+    }
+
+    @Test
+    fun `Should return true if all target TnmTs are listed between possible TnmTs of the derived stage`() {
```
jbartletthmf, line 55: What does "Should return true" mean in this context? Were you trying to design a test case that would result in a PASS?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -30,20 +29,20 @@ class HasTnmTScore(private val scores: Set<TnmT>) : EvaluationFunction {
     )
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val stage = record.tumor.stage
-        val possibleTnmTs = stageMap[stage] ?: emptySet()
+        val stages = record.tumor.stage?.let { setOf(it) }?: record.tumor.derivedStages
```
jbartletthmf, line 32: What happens if `stage` and `derivedStages` on `record.tumor` are both null? I don't see this captured by a test. What should happen in that case?

common/src/main/kotlin/com/hartwig/actin/trial/input/FunctionInputResolver.kt
```diff
@@ -423,6 +423,11 @@ class FunctionInputResolver(
                     return true
                 }
 
+                FunctionInput.ONE_PROTEIN_ONE_STRING -> {
+                    createOneProteinOneStringInput(function)
+                    return true
+                }
```
jbartletthmf, line 429: This change and the corresponding test appear unrelated to the rest of the changes. Do you need to:
```
git merge master
git push
```
?
Otherwise it looks good.

zoehehartwig: That was there before. Look in the files changed. I think it's just because I rearranged some stuff

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
+    )
 
-class HasTnmTScore(private val score: String): EvaluationFunction {
     override fun evaluate(record: PatientRecord): Evaluation {
-        return EvaluationFactory.undetermined("Undetermined if tumor is TNM $score")
+        val stages = record.tumor.stage?.let { setOf(it) }?: record.tumor.derivedStages
+        val possibleTnmTs = stages?.mapNotNull { stageMap[it] }?.flatten()?.toSet() ?: emptySet()
+
+        return when {
+            stages.isNullOrEmpty() -> EvaluationFactory.undetermined("No tumor stage  or derived tumor stage found. Tnm T scores not determined.")
```
spdeleeuw, line 36: Double space between `stage` and `or`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTnmTScore.kt
```diff
@@ -2,11 +2,48 @@ package com.hartwig.actin.algo.evaluation.tumor
...
+            setOf(TumorStage.IV, TumorStage.IVA, TumorStage.IVB, TumorStage.IVC).containsAll(stages) ->
+                EvaluationFactory.undetermined("Cancer is metastatic. Undetermined if tumor is TNM T-classification ${show(targetTnmTs)}")
+            targetTnmTs.containsAll(possibleTnmTs) -> EvaluationFactory.pass("Tumor could be T scores of ${show(possibleTnmTs)}")
```
spdeleeuw, line 39: I think we are certain that the derived T matches one of the requested at this point, right? So maybe change to `Tumor has TNT T-classification ${show(possibleTnmTs)}` ?

spdeleeuw (approved):
Thanks, Zoe! Looks good to me. 2 minor comments

cbruel:
@zoehehartwig Hey! Randomly came across this PR because I was having some merge conflicts. A heads up that you should still change the README (it still says "T.B.I") (it's important to keep the README up to date). If you only make changes to the read me, you can just push this in master (and if unconfident, discuss with me or Simon first what the readme text should be). Thanks for updating the readme!

PR #1059, opened 2025-03-27
ACTIN-1831: Add external configuration of panel gene lists 

pauldwolfe:
Replace the hardcoded list of genes with a TSV file configuration.

The TSV file will have two column, a name (regex) and a gene. The
name unfortunately needs to be duplicated for each gene but that's
just the limitation of using TSV.

### DIFF
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
@@ -1,0 +1,28 @@
+package com.hartwig.actin.clinical.feed.standard
+
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import java.io.File
+
+class PanelGeneList(private val panelRegexToGenes: Map<Regex, List<String>>) {
+
+    operator fun get(input: String): Set<String> {
+        return panelRegexToGenes
+            .filter { (regex, _) -> regex.containsMatchIn(input.lowercase()) }
+            .flatMap { (_, value) -> value }
+            .toSet()
+    }
+
+    companion object {
+        fun create(panelGeneListTsvPath: String): PanelGeneList {
+
+            class PanelGeneEntry(val testNameRegex: String, val gene: String)
+
+            val entries = CsvMapper().apply { registerModule(KotlinModule.Builder().build()) }.readerFor(PanelGeneEntry::class.java)
+                .with(CsvSchema.emptySchema().withHeader().withColumnSeparator('\t')).readValues<PanelGeneEntry>(File(panelGeneListTsvPath))
+            return PanelGeneList(entries.readAll().groupBy { it.testNameRegex }.mapKeys { Regex(it.key) }
+                .mapValues { it.value.map { e -> e.gene } })
+        }
+    }
+}
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
@@ -7,22 +7,18 @@ import java.io.File
 
 class PanelGeneList(private val panelRegexToGenes: Map<Regex, List<String>>) {
 
-    operator fun get(input: String): Set<String> {
-        return panelRegexToGenes
-            .filter { (regex, _) -> regex.containsMatchIn(input.lowercase()) }
-            .flatMap { (_, value) -> value }
-            .toSet()
+    fun matchGenesForTest(testName: String): Set<String> {
+        return panelRegexToGenes.filterKeys { it.containsMatchIn(testName.lowercase()) }.values.flatten().toSet()
     }
 
     companion object {
         fun create(panelGeneListTsvPath: String): PanelGeneList {
 
-            class PanelGeneEntry(val testNameRegex: String, val gene: String)
+            data class PanelGeneEntry(val testNameRegex: String, val gene: String)
 
             val entries = CsvMapper().apply { registerModule(KotlinModule.Builder().build()) }.readerFor(PanelGeneEntry::class.java)
                 .with(CsvSchema.emptySchema().withHeader().withColumnSeparator('\t')).readValues<PanelGeneEntry>(File(panelGeneListTsvPath))
-            return PanelGeneList(entries.readAll().groupBy { it.testNameRegex }.mapKeys { Regex(it.key) }
-                .mapValues { it.value.map { e -> e.gene } })
+            return PanelGeneList(entries.readAll().groupBy(PanelGeneEntry::testNameRegex, PanelGeneEntry::gene).mapKeys { Regex(it.key) })
         }
     }
 }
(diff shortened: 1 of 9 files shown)


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    companion object {
+        fun create(panelGeneListTsvPath: String): PanelGeneList {
+
+            class PanelGeneEntry(val testNameRegex: String, val gene: String)
+
+            val entries = CsvMapper().apply { registerModule(KotlinModule.Builder().build()) }.readerFor(PanelGeneEntry::class.java)
+                .with(CsvSchema.emptySchema().withHeader().withColumnSeparator('\t')).readValues<PanelGeneEntry>(File(panelGeneListTsvPath))
+            return PanelGeneList(entries.readAll().groupBy { it.testNameRegex }.mapKeys { Regex(it.key) }
+                .mapValues { it.value.map { e -> e.gene } })
```
jbartletthmf, line 25: You could pass in a value selector to `groupBy` here:
```suggestion
            return PanelGeneList(entries.readAll().groupBy(PanelGeneEntry::testNameRegex, PanelGeneEntry::gene).mapKeys { Regex(it.key) })
```

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneListTest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class PanelGeneListTest {
+    
+    @Test
+    fun `Should read from panel gene list TSV and match gene lists on test name regex`() {
+        val geneList = PanelGeneList.create(ResourceLocator.resourceOnClasspath("panel_gene_list/panel_gene_list.tsv"))
+        val oncoPanel = geneList["HMF oncopanel "]
+        assertThat(oncoPanel).containsExactly("ABCB1", "ABL1")
+        val avlPanelGenes = geneList["an ngs test"]
+        assertThat(avlPanelGenes).containsExactly("EGFR")
```
nedleitch, line 15: Might be interesting to put something here that makes sure we don't get unexpected results if the input has a pipe character, I'm sure it's working but maybe worth doing anyway?

Also, the English words "leavings" and "angst" would both match this regex I think - would it be worth padding (as in making sure `ngs` only appears as a token unto itself) or anchoring for that one case?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import java.io.File
+
+class PanelGeneList(private val panelRegexToGenes: Map<Regex, List<String>>) {
+
+    operator fun get(input: String): Set<String> {
+        return panelRegexToGenes
+            .filter { (regex, _) -> regex.containsMatchIn(input.lowercase()) }
+            .flatMap { (_, value) -> value }
```
jbartletthmf, line 13: or `.flatMap { it.value }`, or `.values.flatten()`.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
```diff
@@ -0,0 +1,28 @@
+package com.hartwig.actin.clinical.feed.standard
+
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import java.io.File
+
+class PanelGeneList(private val panelRegexToGenes: Map<Regex, List<String>>) {
+
+    operator fun get(input: String): Set<String> {
```
jbartletthmf, line 10: The operator fun is cool here, but I find it a little surprising that what looks like a map lookup is actually a regex evaluation than returns a non-null set for every query. Would it be better to name this function descriptively like `listGenesForPanel`?

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneListTest.kt
jbartletthmf: Good questions. Another option to increase specificity would be to make the regex matching in the code case-sensitive, and then write case-insensitive regexes where appropriate. That might help if we need to match "NGS" but not "rings".

pauldwolfe: This is just a test, you can put any regex you want into the file and anchor as appropriate. In practice these aren't really relevant edge cases.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
pauldwolfe: yup, AI generated :)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
pauldwolfe: you're no fun @jbartletthmf ;)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
jbartletthmf: Eventually I'll train the LLMs to anticipate my PR feedback 🙃

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
jbartletthmf: The best kind of fun is the self-documenting kind 🤓

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/PanelGeneList.kt
```diff
@@ -0,0 +1,24 @@
+package com.hartwig.actin.clinical.feed.standard
+
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import java.io.File
+
+class PanelGeneList(private val panelRegexToGenes: Map<Regex, List<String>>) {
+
+    fun matchGenesForTest(testName: String): Set<String> {
```
jbartletthmf, line 10: I don't love the naming still. Why does the function refer to "Test" while the class refers to "Panel"? In general, I think it's good to minimize our use of "test" to describe anything that doesn't pertain to our test suites. I also think that `match` doesn't obviously convey what's being returned here compared to `find`, `list`, `filter`, `provide`, etc.

jbartletthmf (approved):
I still have some thoughts on naming, but it's good to go.

PR #1063, opened 2025-04-01
ACTIN-1825: Add registration date so that actin pipeline can filter out missing questionnaire warnings for old patients

zoehehartwig:
This is to prevent old patients from showing the missing questionnaire warning. I simply modified the feedValidationWarning to have an optional registration date that is passed in if there is a missing questionnaire warning. The rest of the changes are in actin pipeline

### DIFF
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
@@ -108,7 +108,7 @@ class EmcClinicalFeedIngestor(
 
             Triple(
                 record,
-                ingestionResult(record.patientId, questionnaire, patientEvaluation, questionnaireCurationErrors, feedRecord, record.patient.registrationDate),
+                ingestionResult(record.patientId, record.patient.registrationDate, questionnaire, patientEvaluation, questionnaireCurationErrors, feedRecord),
                 patientEvaluation
             )
         }
@@ -178,11 +178,11 @@ class EmcClinicalFeedIngestor(
 
     private fun ingestionResult(
         patientId: String,
+        registrationDate: LocalDate,
         questionnaire: Questionnaire?,
         patientEvaluation: CurationExtractionEvaluation,
         questionnaireCurationErrors: List<QuestionnaireCurationError>,
         feedRecord: FeedRecord,
-        registrationDate: LocalDate,
     ): PatientIngestionResult {
         val curationResults = curationResultsFromWarnings(patientEvaluation.warnings)
 
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
@@ -182,7 +182,7 @@ class EmcClinicalFeedIngestor(
         questionnaire: Questionnaire?,
         patientEvaluation: CurationExtractionEvaluation,
         questionnaireCurationErrors: List<QuestionnaireCurationError>,
-        feedRecord: FeedRecord,
+        feedRecord: FeedRecord
     ): PatientIngestionResult {
         val curationResults = curationResultsFromWarnings(patientEvaluation.warnings)
 
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
@@ -39,6 +39,7 @@ import com.hartwig.actin.datamodel.clinical.VitalFunctionCategory.HEART_RATE
 import com.hartwig.actin.datamodel.clinical.VitalFunctionCategory.NON_INVASIVE_BLOOD_PRESSURE
 import com.hartwig.actin.datamodel.clinical.VitalFunctionCategory.SPO2
 import com.hartwig.actin.datamodel.clinical.ingestion.FeedValidationWarning
+import com.hartwig.actin.datamodel.clinical.ingestion.Warnings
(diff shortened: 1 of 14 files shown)


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/EmcClinicalFeedIngestor.kt
```diff
@@ -180,15 +181,16 @@ class EmcClinicalFeedIngestor(
...
 
         val ingestionStatus =
             if (questionnaire == null || curationResults.isNotEmpty()) PatientIngestionStatus.WARN else PatientIngestionStatus.PASS
 
         val validationWarnings = if (questionnaire == null) {
-            feedRecord.validationWarnings + FeedValidationWarning(patientId, "No Questionnaire found")
+            feedRecord.validationWarnings + FeedValidationWarning(patientId, "No Questionnaire found", registrationDate)
         } else {
             feedRecord.validationWarnings
         }
```
jbartletthmf, line 196: I think it would make more sense to put the registration date in the `PatientIngestionResult` and make it non-nullable. Does it need to go in the `FeedValidationWarning`?

andreia-hartwig (approved):
@zoehehartwig  Perhaps you can extend ClinicalIngestionFeedAdapterTest.`Should run ingestion from proper curation and feed files, read from filesystem` to account for the new field?

jbartletthmf:
I don't have any further comments on the code, but:

1. Can you rename this PR to match convention, i.e. `ACTIN-1825: <brief description>`?
2. Please note that this change needs to be rolled out with a change to actin-pipelines, at least to upgrade the actin-datamodel dependency in that project.
3. It's unusual to request reviews for a PR still in draft state. Were you intentionally keeping it as a draft until the pipelines changes are done, or is it time to mark this ready for review?

zoehehartwig:
I'm still working on it.

andreia-hartwig:
@zoehehartwig Btw I did not realise this was still a draft when I approve the changes.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/ingestion/IngestionResult.kt
```diff
@@ -13,6 +15,7 @@ enum class PatientIngestionStatus {
 
 data class PatientIngestionResult(
```
jbartletthmf, line 16: Do you remember looking at the custom date adapter we use for serialization in ACTIN? The serializer is created here:
https://github.com/hartwigmedical/actin/blob/54d1b875044b97e068f684f13d43ea3e0a907a5f/clinical/src/main/kotlin/com/hartwig/actin/clinical/ClinicalIngestionApplication.kt#L116-L123
I was suggesting that you make another version of `GsonSerializer.create()` or parameterize it to optionally skip loading the special type adapters for date/time types. Otherwise, you'll need to add special type deserialization logic in pipelines to load the ingestion JSON (unless this works magically).

jbartletthmf: It might not be obvious how these results are being serialized since this is just happening in the application's `run()`. You could make this more testable by adding a `IngestionResultSerializer` class/object with a `toJson` function that returns a string, and referencing it in the `ClinicalIngestionApplication` code above. Then, a new `IngestionResultSerializerTest` could ensure that the serialized result is in an appropriate format.

common/src/main/kotlin/com/hartwig/actin/util/json/GsonLocalDateAdapter.kt
```diff
@@ -12,47 +12,50 @@ class GsonLocalDateAdapter : TypeAdapter<LocalDate?>() {
...
+                        month = reader.nextInt()
+                    }
+                    if ((field == "day")) {
+                        reader.peek()
+                        day = reader.nextInt()
+                    }
+                }
+                reader.endObject()
+                return LocalDate.of(year, month, day)
             }
```
jbartletthmf, line 58: I don't think we need to support this case here anymore either. I was considering ORANGE inputs, but we use `hmftools` to deserialize those: https://github.com/hartwigmedical/actin/blob/f57305362f3683df4f609a4308030050b318efbc/molecular/src/main/kotlin/com/hartwig/actin/molecular/MolecularInterpreterApplication.kt#L96

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/ClinicalGsonDeserializer.kt
```diff
@@ -51,10 +51,14 @@ object ClinicalGsonDeserializer {
         ): LocalDate? {
```
jbartletthmf, line 51: Could we just use `GsonLocalDateAdapter` instead and do away with this class?

common/src/main/kotlin/com/hartwig/actin/util/json/GsonLocalDateAdapter.kt
```diff
@@ -18,44 +18,12 @@ class GsonLocalDateAdapter : TypeAdapter<LocalDate?>() {
...
-                        reader.peek()
-                        day = reader.nextInt()
-                    }
-                }
-                reader.endObject()
-                return LocalDate.of(year, month, day)
-            }
         }
+
+        reader.peek()
```
jbartletthmf, line 26: Does this do anything? It seems to be used above to check the value at the cursor, but we aren't checking anything here.

common/src/main/kotlin/com/hartwig/actin/util/json/GsonLocalDateAdapter.kt
```diff
@@ -18,44 +18,12 @@ class GsonLocalDateAdapter : TypeAdapter<LocalDate?>() {
...
+        return LocalDate.parse(reader.nextString())
```
jbartletthmf, line 27: Is this equivalent to:
```kotlin
return if (reader.peek() == JsonToken.NULL) {
    reader.skipValue()
    null
} else LocalDate.parse(reader.nextString())
```
?

PR #1068, opened 2025-04-03
ACTIN-1847 Fix issues with empty molecular results and tested genes

pauldwolfe:
Filter the empty results and instead of a dynamic list based on variants and negative results, we maintain a parallel static
list built at ingestion time.


spdeleeuw (approved):
LGTM

PR #1074, opened 2025-04-10
ACTIN-1870: Make ignoring complete sequencing tests case sensitive

ninajacobs:
... accidentally used ACTIN-1869

Should be: https://hartwigmedical.atlassian.net/browse/ACTIN-1870

Goal: Ignore tests case sensitively, e.g. if test name is "NGS" then ignore only if we curate "NGS" and not "NgS"


ninajacobs:
I was thinking about this after I pushed, but it may not be correct actually, will investigate myself in ~ an hour

ninajacobs:
On review, I think it would actually work, happy to hear!

PR #1075, opened 2025-04-10
ACTIN-1874: Add HAS_KNOWN_HSV_INFECTION rule

ninajacobs:


common/src/main/kotlin/com/hartwig/actin/algo/icd/IcdConstants.kt
```diff
@@ -100,6 +100,7 @@ object IcdConstants {
     const val ACUTE_HEPATITIS_C_CODE = "1E50.2"
     const val CHRONIC_HEPATITIS_C_CODE = "1E51.1"
     const val HIV_BLOCK = "BlockL1-1C6"
+    const val HSV_CODE = "1F00"
```
cbruel, line 103: Wouldn't we also want to evaluate "Anogenital herpes simplex infection: 1A94" for this rule? As you can see in https://icd.who.int/browse/2024-01/mms/en#248851702 this is not part of "1F00"

cbruel (approved):
Looks good, 1 comment! And 1 additional thought (this does not only apply for this new rule but also for rules like `HAS_KNOWN_EBV_INFECTION`) shouldn't we also look at the viruses in MolecularRecord -> Drivers -> viruses. Shouldn't we want a pass here if the `VirusType.HUMAN_HERPES_VIRUS_B` is detected in the WGS of this patient?

common/src/main/kotlin/com/hartwig/actin/algo/icd/IcdConstants.kt
ninajacobs: agreed, added it!

PR #1076, opened 2025-04-11
ACTIN-1877: Implement HasHadFirstLineSystemicTreatmentName

cbruel:


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyserTest.kt
```diff
@@ -92,34 +93,59 @@ class SystemicTreatmentAnalyserTest {
...
+        assertNameForSystemicTreatmentHistoryEntry(treatmentHistory, "4", ::firstSystemicTreatment)
+    }
+
+    private fun assertNameForSystemicTreatmentHistoryEntry(
+        treatmentHistory: List<TreatmentHistoryEntry>,
+        name: String,
+        treatmentSelector: (List<TreatmentHistoryEntry>) -> TreatmentHistoryEntry?
+    ) {
+        val treatment = treatmentSelector(treatmentHistory)
+        assertThat(treatment!!.treatments.first().name).isEqualTo(name)
```
jbartletthmf, line 149: You could inline here and keep it null-safe:
```suggestion
        assertThat(treatmentSelector(treatmentHistory)?.treatments?.first()?.name).isEqualTo(name)
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentNameTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        )
+    }
+
+    @Test
+    fun `Should pass when patient has only received correct treatment but with unknown date`() {
+        assertEvaluation(EvaluationResult.PASS, function.evaluate(withTreatmentHistory(listOf(matchingTreatment))))
+    }
+
+    @Test
+    fun `Should evaluate to undetermined when trial treatment received in first line`() {
```
jbartletthmf, line 62: `with unknown category`?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentNameTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val firstLineTrial = treatmentHistoryEntry(setOf(treatment("trial", true)), isTrial = true, startYear = 2024)
+        assertEvaluation(
+            EvaluationResult.UNDETERMINED,
+            function.evaluate(withTreatmentHistory(listOf(firstLineTrial, otherTreatment.copy(startYear = 2025))))
+        )
+    }
+
+    @Test
+    fun `Should evaluate to undetermined when correct treatment received as first treatment but other treatments with unknown date`() {
```
jbartletthmf, line 71: How about:
```suggestion
    fun `Should evaluate to undetermined when received correct treatment as first treatment and other treatments with unknown date`() {
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentNameTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import com.hartwig.actin.algo.evaluation.EvaluationAssert.assertEvaluation
+import com.hartwig.actin.datamodel.algo.EvaluationResult
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatment
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatmentHistoryEntry
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.withTreatmentHistory
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import org.junit.Test
+
+class HasHadFirstLineSystemicTreatmentNameTest {
```
jbartletthmf, line 11: I think there are some other cases to consider here, e.g. the presence of non-systemic treatments.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentName.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.hartwig.actin.algo.evaluation.EvaluationFactory
+import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+
+class HasHadFirstLineSystemicTreatmentName(private val treatmentToFind: Treatment) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val (treatmentsWithStartDate, treatmentsWithoutStartDate) = record.oncologicalHistory.partition { it.startYear != null }
```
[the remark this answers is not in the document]
cbruel: Fair point, indeed shouldn't matter if radiotherapy or surgery was given before, thanks!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentName.kt
```diff
@@ -0,0 +1,41 @@
...
+            hadTreatmentToFindWithUnknownStartDate && treatmentsWithoutStartDate.size == 1
+        val firstTreatmentIsPotentialTrialMatch = firstTreatment?.isTrial == true && (firstTreatment.categories()
+            .containsAll(treatmentToFind.categories()) || firstTreatment.categories().isEmpty())
+
+        val treatmentToFindDisplay = treatmentToFind.display()
+        return when {
+            (hasHadTreatmentToFindInFirstLine && treatmentsWithoutStartDate.isEmpty())
+                    || hasOnlyHadTreatmentToFindButWithUnknownStartDate -> {
+                EvaluationFactory.pass("Has received $treatmentToFindDisplay as first-line treatment")
+            }
```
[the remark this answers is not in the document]
cbruel: @jbartletthmf 
It could indeed in some cases happen that a treatment is re-introduced. So the treatment being given in e.g. first line and third line. This case would currently pass (added an extra test for this in the test class for completeness).

I don't think I understand your point 100% though because you are referring to treatments with unknown date. I would say we have to evaluate to undetermined when there are any treatments with unknown date since we can't be certain which treatment was given first?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentName.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+
+class HasHadFirstLineSystemicTreatmentName(private val treatmentToFind: Treatment) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        val (treatmentsWithStartDate, treatmentsWithoutStartDate) = record.oncologicalHistory.partition { it.startYear != null }
+        val firstTreatment = SystemicTreatmentAnalyser.firstSystemicTreatment(treatmentsWithStartDate)
+        val hasHadTreatmentToFindInFirstLine = firstTreatment?.allTreatments()?.contains(treatmentToFind) == true
```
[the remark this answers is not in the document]
cbruel: @jbartletthmf 
Thanks for this suggestion! Isn't this "function that returns an ID", kind of equal to the `display()` we already have? So we could use `display()` for comparison? 

But I'm curious why comparing the data classes isn't a good approach? I personally like that better than comparing on naming, therefore did the comparison this way. The `Treatments`s should be exactly equal (i.e. all fields matching)?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentName.kt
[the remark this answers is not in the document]
jbartletthmf: If all the treatment entries with unknown date contain the `treatmentToFind`, and the first treatment with known date also contains the `treatmentToFind`, then we know that `treatmentToFind` was given in the first line regardless of which entry occurred first.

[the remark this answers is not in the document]
cbruel: @jbartletthmf Ahh get it! Quite rare case, but will add it for completeness, thanks for bringing this up!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadFirstLineSystemicTreatmentName.kt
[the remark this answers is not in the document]
cbruel: @jbartletthmf Thanks for the explanation! Makes sense for future set up.

Do you agree on using `display()` for comparison for now?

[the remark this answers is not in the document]
jbartletthmf: I think `display()` overcomplicates the comparison, as there's a regex, inconsistent case, and potential for an override. Instead, I would continue to use `name.lowercase()` as we've done elsewhere.

jbartletthmf (approved):
Looks good, just suggested some refactoring.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSpecificFirstLineSystemicTreatmentTest.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Test
+    fun `Should evaluate to undetermined when correct treatment received with unknown start date and another treatment with known start date`() {
+        assertEvaluation(
+            EvaluationResult.UNDETERMINED,
+            function.evaluate(withTreatmentHistory(listOf(matchingTreatment, otherTreatment.copy(startYear = 2025))))
+        )
+    }
+}
```
ninajacobs, line 117: Test situations look good Charlotte!

I think one situation would be good to additionally to consider: if the correct systemic treatment was not the first line of all treatments, but it was administered as first line after surgery, it could be undetermined (because then it perhaps it may still count as first line in metastatic setting, e.g. if the treatment before was e.g. in neoadjuvant setting)? Similarly, it may also be the situation if there was surgery - adjuvant - and then first line palliative setting..

I'd also be okay with logging this as part of larger review later.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/trial/EligibilityRule.kt
```diff
@@ -121,7 +121,7 @@ enum class EligibilityRule(val input: FunctionInput?, val ihcProtein: String? =
     HAS_HAD_ANY_SYSTEMIC_CANCER_TREATMENT_WITHIN_X_MONTHS(FunctionInput.ONE_INTEGER),
     HAS_HAD_TREATMENT_NAME_X(FunctionInput.ONE_SPECIFIC_TREATMENT),
     HAS_HAD_TREATMENT_NAME_X_WITHIN_Y_WEEKS(FunctionInput.ONE_SPECIFIC_TREATMENT_ONE_INTEGER),
-    HAS_HAD_FIRST_LINE_TREATMENT_NAME_X(FunctionInput.ONE_SPECIFIC_TREATMENT),
+    HAS_HAD_FIRST_LINE_SYSTEMIC_TREATMENT_NAME_X(FunctionInput.ONE_SPECIFIC_TREATMENT),
```
ninajacobs, line 124: what happens now if we accidentally input a non-systemic treatment? can we relatively easily add a check to prevent this, or do you prefer to highlight it clearly in the README?

cbruel: Interesting point :). I added a check for it!

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSpecificFirstLineSystemicTreatmentTest.kt
cbruel: I will create a jira ticket for this!

PR #1077, opened 2025-04-11
ACTIN-1879: Revert changes of ACTIN-1769

cbruel:


kduyvesteyn (approved):
Have only scanned the changes since I had them roughly in my head still, and the change generally looks as-expected. Assume you have tested on an actual patient, and if this looks ok then go-for-it!

PR #1079, opened 2025-04-11
ACTIN-1867 Expand panel gene list to cover full coverage specification

pauldwolfe:
The flow goes like this:
- Allow for curation of the test name so we can remove the regex matching
- The "gene list" becomes a panel specification, allowing each gene to have a list of alterations
which is tests
- The application of this new panel specification moves to the molecular annotation step.
- Rules need to specify the genes and can optionally specify targets as well

This is a draft, tests still need work and not all rules have been adapted.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -26,6 +27,7 @@ class GeneHasVariantInExonRangeOfType(
     }
 
     override fun genes() = listOf(gene)
+    override fun targets() = listOf(MolecularTestTarget.MUTATION, MolecularTestTarget.FUSION)
```
pauldwolfe, line 30: This is an interesting one, I suppose we'd really want to run the rule if we tested for mutation OR fusions and not AND.  With the current code, we could leave targets empty and then it will run as long as the gene is tested for anything (which is logical as we never just test for amps or dels in our current panel specs).  Alternative we could have a way of modelling the OR predicate.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
```diff
@@ -17,44 +18,57 @@ import com.hartwig.actin.datamodel.clinical.provided.ProvidedMolecularTestResult
...
+                CurationResponse.createFromConfigs(
+                    testCuration.find(test.test),
+                    ehrPatientRecord.patientDetails.hashedId,
+                    CurationCategory.SEQUENCING_TEST,
+                    test.test,
+                    "sequencing test result",
+                    false
+                )
+            testCurationConfig.config()?.let { testCuration ->
+                if (!testCuration.ignore) {
```
pauldwolfe, line 39: Previously we had the ability to ignore on " patient_id | test_name " to work around specific issues. Looking at the sheet, I think most test names we ignore we'd want to ignore everywhere (eg. IHC, ALK, Moleculair, etc), but there are some examples where not.  eg.

```
NKI-LUNG-00085 | Archer Lung Panel
NKI-LUNG-00085 | Ngs Ampliseq
```

But maybe going forward we can fix those issues at the source (PALGA LLM).

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasActivatingMutation.kt
```diff
@@ -48,6 +48,7 @@ class GeneHasActivatingMutation(
 ) : MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+
```
ninajacobs, line 51: enter

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ninajacobs: I think from clinical perspective in this case checking only mutation should be enough

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
ninajacobs: see my comment on Slack

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/MolecularTest.kt
```diff
@@ -14,7 +14,7 @@ interface MolecularTest {
     val hasSufficientPurity: Boolean
     val hasSufficientQuality: Boolean
 
-    fun testsGene(gene: String): Boolean
+    fun testsGene(gene: String, molecularTestTargets: List<MolecularTestTarget> = emptyList()): Boolean
```
kduyvesteyn, line 17: No big fan of defaults for parameters of classes, especially if they are not null.

In this case the default input is even sort of invalid -> it is impossible to test a gene but on zero targets. If we'd provide a default then `all values` would be better (if you only provide the tested gene, assume that gene has been tested in its entirety). 

Also, I think a `Set` would be better here than a `List`, there is no use-case for duplicates

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/PanelRecord.kt
```diff
@@ -16,7 +16,10 @@ data class PanelRecord(
     override val hasSufficientQuality: Boolean
 ) : MolecularTest {
 
-    override fun testsGene(gene: String): Boolean {
-        return testedGenes.contains(gene)
+    override fun testsGene(gene: String, molecularTestTargets: List<MolecularTestTarget>): Boolean {
+        if (molecularTestTargets.isEmpty()) {
+            return geneSpecifications[gene] != null
+        }
```
kduyvesteyn, line 22: I think a `Set` would be more suited here.

Also, current implementation is fine I think but maybe you could consider demanding the `molecularTestTargets` to include at least 1 entry or otherwise throw an exception. Every caller of this function should know the targets it is interested in.

molecular/src/main/kotlin/com/hartwig/actin/molecular/MolecularInterpreterConfig.kt
```diff
@@ -19,6 +19,7 @@ data class MolecularInterpreterConfig(
     val ensemblCachePath: String,
     val driverGenePanelPath: String,
     val knownFusionsPath: String,
+    val panelSpecificationsFilePath: String?,
```
kduyvesteyn, line 22: In this optional to ease the rollout of this change?

I imagine in practice we'd always provide this file?

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelSpecifications.kt
kduyvesteyn, line 1: Shouldn't this file be called `PanelSpecificationsFile`?

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelSpecifications.kt
```diff
@@ -0,0 +1,35 @@
...
+            entries.readAll().groupBy(PanelGeneEntry::testName)
+                .mapValues {
+                    it.value.map { g ->
+                        PanelGeneSpecification(
+                            g.gene,
+                            listOfNotNull(
+                                if (g.fusion) MolecularTestTarget.FUSION else null,
+                                if (g.mutation) MolecularTestTarget.MUTATION else null,
+                                if (g.amp) MolecularTestTarget.AMPLIFICATION else null,
+                                if (g.del) MolecularTestTarget.DELETION else null,
```
kduyvesteyn, line 29: Maybe spell-out amp and del in this case as well?

Also, maybe good to consider throwing an exception if this list is empty? I don't think that should ever be the case with proper configuration, so a good way to catch a mistake on the configuration side?

kduyvesteyn (approved):
Looks good, really nice improvement! 

@pauldwolfe I think the current rule of "every test has to be configured in terms of gene specifications" is a good way to start on the molecular interpretation side. Maybe long-term it may be too restrictive but I think that's a better starting point than being too lenient and erroneously forget about curation. 

@ninajacobs One thought I had: Could we maybe combine the `AMPLICATION` and `DELETION` targets into one `COPYNUMBER` target? Are there tests that test only for AMP or DEL but not both?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneIsInactivated.kt
```diff
@@ -15,6 +16,7 @@ import java.time.LocalDate
 class GeneIsInactivated(private val gene: String, maxTestAge: LocalDate? = null) : MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targets() = listOf(MolecularTestTarget.MUTATION)
```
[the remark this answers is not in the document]
ninajacobs: Yes, but I think it would give noise if we require this (many 'gene not tested' warnings) - something to discuss with hospital if desired. That's why I think configuring MUTATION as requirement only is defendable now

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasSpecificExonSkipping.kt
```diff
@@ -14,6 +15,7 @@ class GeneHasSpecificExonSkipping(private val gene: String, private val exonToSk
     MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targets() = listOf(MolecularTestTarget.MUTATION)
```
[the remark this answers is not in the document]
kduyvesteyn: Agreed with your suggestion!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneIsWildType.kt
```diff
@@ -16,6 +17,7 @@ import java.time.LocalDate
 class GeneIsWildType(private val gene: String, maxTestAge: LocalDate? = null) : MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targets() = listOf(MolecularTestTarget.MUTATION)
```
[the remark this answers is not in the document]
ninajacobs: Not sure.. amp doesn't impact wildtype really right? perhaps del does? And also here I think MUTATION may be sufficient, to reduce warnings that a gene is not tested for fusions or del etc?

ninajacobs:
Are there tests that test only for AMP or DEL but not both? -> yes, most cases (see the sheet)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneIsInactivated.kt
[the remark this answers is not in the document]
kduyvesteyn: Valid point. Maybe we can make things more clear in naming, e.g. `minimalTargets` rather than `targets` or something? 

My interpretation at least was that targets should contain the exhaustive set of targets to evaluate the function comprehensively but I now understand its more the minimal set of targets to hide a warning of gene not being tested.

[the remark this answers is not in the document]
ninajacobs: Agreed.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasSufficientCopyNumber.kt
```diff
@@ -47,6 +48,7 @@ class GeneHasSufficientCopyNumber(private val gene: String, private val requeste
     MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targets() = listOf(MolecularTestTarget.AMPLIFICATION, MolecularTestTarget.DELETION)
```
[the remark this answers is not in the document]
kduyvesteyn: This is currently not the case if I am not mistaken, see `PanelRecord`. 

I think we should change `containsAll` to "having a non-empty intersect of geneSpecifications and molecularTestTargets"

```
override fun testsGene(gene: String, molecularTestTargets: List<MolecularTestTarget>): Boolean {
        if (molecularTestTargets.isEmpty()) {
            return geneSpecifications[gene] != null
        }
        return geneSpecifications[gene]?.containsAll(molecularTestTargets) ?: false
    }
```

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
```diff
@@ -12,49 +13,60 @@ import com.hartwig.actin.datamodel.clinical.SequencedFusion
...
+                    val nonIHCTestResults = test.results.filter { it.ihcResult == null }.toSet()
+                    val (onlyFreeTextResults, populatedResults) = nonIHCTestResults
+                        .partition { checkAllFieldsNull(it) }
+                    val mandatoryCurationTestResults = curate(ehrPatientRecord, onlyFreeTextResults)
+                    val optionalCurationTestResults = curate(ehrPatientRecord, populatedResults)
+                    val allResults =
+                        removeCurated(removeCurated(nonIHCTestResults, mandatoryCurationTestResults), optionalCurationTestResults) +
```
jbartletthmf, line 45: Equivalent to:
```suggestion
                        removeCurated(nonIHCTestResults, mandatoryCurationTestResults + optionalCurationTestResults) +
```
?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorSequencingTestExtractor.kt
```diff
@@ -12,49 +13,60 @@ import com.hartwig.actin.datamodel.clinical.SequencedFusion
...
+                                extract(mandatoryCurationTestResults) + extract(optionalCurationTestResults)
```
jbartletthmf, line 46: Same as:
```suggestion
                                extract(mandatoryCurationTestResults + optionalCurationTestResults)
```
? It seems like you could combine these lists above to simplify this step.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/PanelRecord.kt
```diff
@@ -16,7 +16,10 @@ data class PanelRecord(
...
+        return geneSpecifications[gene]?.containsAll(molecularTestTargets) ?: false
```
jbartletthmf, line 23: I don't think the `isEmpty()` check is required (unless you're changing the requirements as Korneel suggested). `containsAll` should return true if `molecularTestTargets` is empty:
```suggestion
        return geneSpecifications[gene]?.containsAll(molecularTestTargets) == true
```

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/PanelSpecifications.kt
```diff
@@ -0,0 +1,15 @@
+package com.hartwig.actin.datamodel.molecular
+
+data class PanelGeneSpecification(val geneName: String, val targets: List<MolecularTestTarget>)
+
+class PanelSpecifications(private val panelSpecifications: Map<String, List<PanelGeneSpecification>>) {
+
+    fun genesForPanel(panelName: String): Map<String, List<MolecularTestTarget>> {
+        return panelSpecifications[panelName]?.groupBy(PanelGeneSpecification::geneName)
+            ?.mapValues { it.value.flatMap { c -> c.targets } }
```
jbartletthmf, line 9: Would it be better to do this once when the `PanelSpecifications` is created instead of every time we retrieve the genes for a panel?

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelSpecifications.kt
```diff
@@ -0,0 +1,35 @@
...
+object PanelSpecificationsFile {
+
+    fun create(panelGeneListTsvPath: String): PanelSpecifications {
+
+        data class PanelGeneEntry(val testName: String, val gene: String, val fusion: Boolean, val mutation: Boolean, val amp: Boolean, val del: Boolean)
+
+        val entries = CsvMapper().apply { registerModule(KotlinModule.Builder().build()) }.readerFor(PanelGeneEntry::class.java)
+            .with(CsvSchema.emptySchema().withHeader().withColumnSeparator('\t')).readValues<PanelGeneEntry>(File(panelGeneListTsvPath))
+        return PanelSpecifications(
+            entries.readAll().groupBy(PanelGeneEntry::testName)
```
jbartletthmf, line 20: You can use the 2-parameter form of `groupBy` to provide a value selector here as well instead of mapping them after the fact.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelSpecifications.kt
```diff
@@ -0,0 +1,35 @@
...
+                    it.value.map { g ->
+                        PanelGeneSpecification(
+                            g.gene,
+                            listOfNotNull(
+                                if (g.fusion) MolecularTestTarget.FUSION else null,
+                                if (g.mutation) MolecularTestTarget.MUTATION else null,
+                                if (g.amp) MolecularTestTarget.AMPLIFICATION else null,
+                                if (g.del) MolecularTestTarget.DELETION else null,
+                            )
+                        )
```
jbartletthmf, line 31: I think it would be better to do this in a helper, or better yet, in a new `PanelGeneEntry.toPanelGeneSpecification()`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneIsInactivated.kt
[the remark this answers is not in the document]
pauldwolfe: Just to double check with this one, if a gene was only tested for DELETION do we want to evaluate or consider it untested?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/MolecularRecord.kt
```diff
@@ -24,8 +26,9 @@ data class MolecularRecord(
     override val evidenceSource: String,
 ) : MolecularTest {
 
-    override fun testsGene(gene: String) =
-        if (experimentType == ExperimentType.HARTWIG_TARGETED) drivers.copyNumbers.any { gene == it.gene } else true
+    override fun testsGene(gene: String, molecularTestTargets: Predicate<List<MolecularTestTarget>>) =
+        if (experimentType == ExperimentType.HARTWIG_TARGETED) geneSpecifications!![gene]?.let { molecularTestTargets.test(it) }
+            ?: false else true
```
kduyvesteyn, line 31: I think it would be good to add tests for this. The `?: false else true` looks counterintuitive to me, though maybe it is correct, would be easier to tell with a test

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/PanelRecord.kt
```diff
@@ -16,7 +17,7 @@ data class PanelRecord(
     override val hasSufficientQuality: Boolean
 ) : MolecularTest {
 
-    override fun testsGene(gene: String): Boolean {
-        return testedGenes.contains(gene)
+    override fun testsGene(gene: String, molecularTestTargets: Predicate<List<MolecularTestTarget>>): Boolean {
+        return geneSpecifications[gene]?.let { molecularTestTargets.test(it) } ?: false
```
kduyvesteyn, line 21: Although does look a little easier, would sleep better if this would be unit-tested :)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularEvaluationFunction.kt
```diff
@@ -48,11 +48,16 @@ abstract class MolecularEvaluationFunction(maxTestAge: LocalDate? = null, useIns
         }
     }
 
+    private fun targetSuffix() = "for " + targetCoveragePredicate().targets.joinToString {
```
ninajacobs, line 51: a space should be added before "for" + it would be clearer to concat with 'or' and 'and' appropriately (i.e. "not tested for fusions or mutations" or "not tested for fusions and mutations" instead of "not tested for fusions, deletions"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetPredicates.kt
```diff
@@ -0,0 +1,31 @@
...
+    Predicate<List<MolecularTestTarget>> {
+
+    override fun test(t: List<MolecularTestTarget>) = predicate.test(t)
+
+    companion object {
+        fun any() = or(*MolecularTestTarget.entries.toTypedArray())
+
+        fun all() = and(*MolecularTestTarget.entries.toTypedArray())
+
+        fun exactly(target: MolecularTestTarget) = and(target)
```
kduyvesteyn, line 16: The logic is hard to review for me but the name suggests that a test should test _exactly_ the target, shouldn't it be _at least_? E.g. for `GeneHasActivatingMutation`, if a molecular test tests this gene both for mutations and fusions then we would consider this gene tested, but the name implies that the molecular test should test `MUTATION` exactly and will fail if the test also checked for fusion?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularEvaluationFunction.kt
ninajacobs: btw it would even be clearer in messaging if we only display the type that was _not_ tested, so if mutation AND fusion required, but only fusion tested -> show "gene X not tested for mutations" instead of "gene X not tested for mutations and fusions" (then people can be more confused)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasFusionInGene.kt
```diff
@@ -4,15 +4,16 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
-import com.hartwig.actin.datamodel.molecular.driver.ProteinEffect
+import com.hartwig.actin.datamodel.molecular.driver.DriverLikelihood
 import com.hartwig.actin.datamodel.molecular.driver.FusionDriverType
+import com.hartwig.actin.datamodel.molecular.driver.ProteinEffect
 import java.time.LocalDate
 
 class HasFusionInGene(private val gene: String, maxTestAge: LocalDate? = null) : MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targetCoveragePredicate() = TargetCoveragePredicate.all()
```
kduyvesteyn, line 16: @ninajacobs This feels wrong to me?

ninajacobs: Agreed

ninajacobs: should be exactly FUSION

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetPredicates.kt
ninajacobs: I agree

ninajacobs:
we're still missing implementation for HasMolecularEventWithSocTargetedTherapyForNSCLCAvailable

I'd say either
TargetCoveragePredicate.or(MolecularTestTarget.MUTATION, MolecularTestTarget.FUSION)

or FUSION for genes in FUSION_LIST and MUTATION for all other mentioned genes.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneIsWildType.kt
```diff
@@ -4,18 +4,19 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
 import com.hartwig.actin.datamodel.molecular.driver.ProteinEffect
 import com.hartwig.actin.datamodel.molecular.driver.Variant
-import com.hartwig.actin.datamodel.molecular.driver.Disruption
-import com.hartwig.actin.datamodel.molecular.driver.HomozygousDisruption
 import java.time.LocalDate
 
 class GeneIsWildType(private val gene: String, maxTestAge: LocalDate? = null) : MolecularEvaluationFunction(maxTestAge) {
 
     override fun genes() = listOf(gene)
+    override fun targetCoveragePredicate() = TargetCoveragePredicate.all()
```
ninajacobs, line 19: I propose exactly(MUTATION)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetPredicates.kt
pauldwolfe: Agreed, noticed the same thing myself.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularEvaluationFunctionTest.kt
```diff
@@ -103,6 +108,25 @@ class MolecularEvaluationFunctionTest {
...
+                            TestMolecularFactory.createMinimalTestPanelRecord()
+                                .copy(geneSpecifications = mapOf("GENE" to listOf(MolecularTestTarget.MUTATION)))
+                        )
+                    )
+                )
+            )
+        assertMolecularEvaluation(EvaluationResult.UNDETERMINED, evaluation)
+        assertThat(evaluation.undeterminedMessages).containsExactly("Gene(s) GENE not tested for fusions")
+        assertThat(evaluation.isMissingMolecularResultForEvaluation).isTrue()
+    }
```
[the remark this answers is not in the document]
pauldwolfe: On second thought, we currently don't have any rules that accept multiple genes and if we did it would be ambiguous whether that should be ALL or ANY genes. I'm going to try to swap the list of genes out with a single nullable gene to simplify.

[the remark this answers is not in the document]
jbartletthmf: `NsclcDriverGeneStatusesAreAvailable` is the one I had in mind but that one doesn't extend `MolecularEvaluationFunction`. I think switching to a single gene is a good idea, and we can handle the multi-gene cases outside of the abstract class 👍

molecular/src/main/kotlin/com/hartwig/actin/molecular/MolecularInterpreterConfig.kt
pauldwolfe: It was more for the current EMC case as I guess our first version of the spec file won't be correct for them and they currently don't do panels.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/PanelSpecifications.kt
```diff
@@ -1,15 +1,26 @@
...
+    fun testsGene(gene: String, molecularTestTargets: Predicate<List<MolecularTestTarget>>) =
+        geneTargetMap[gene]?.let { molecularTestTargets.test(it) } ?: false
+}
+
+class PanelSpecifications(panelGeneSpecifications: Map<String, List<PanelGeneSpecification>>) {
+
+    private val panelSpecifications: Map<String, PanelSpecification> = panelGeneSpecifications.entries.associate {
+        it.key to PanelSpecification(
+            it.value.groupBy(PanelGeneSpecification::geneName).mapValues { v -> v.value.flatMap { c -> c.targets } })
+    }
```
jbartletthmf, line 18: Tried to clean this up a bit:
```suggestion
    private val panelSpecifications: Map<String, PanelSpecification> = panelGeneSpecifications.mapValues { (_, specs) ->
        PanelSpecification(
            specs.groupBy(PanelGeneSpecification::geneName).mapValues { it.value.flatMap(PanelGeneSpecification::targets) })
    }
```

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/MolecularRecordTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.Test
+
+class MolecularRecordTest {
+
+    @Test
+    fun `Should check panel specification for tested gene and target when targeted`() {
+        val targets = listOf(MolecularTestTarget.MUTATION)
+        val gene = "gene"
+        val record =
+            TestMolecularFactory.createMinimalTestMolecularRecord()
```
jbartletthmf, line 13: extra line break

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/MolecularRecordTest.kt
```diff
@@ -0,0 +1,25 @@
...
+                .copy(experimentType = ExperimentType.HARTWIG_TARGETED, specification = PanelSpecification(mapOf(gene to targets)))
+        assertThat(record.testsGene(gene) { it == targets }).isTrue()
+        assertThat(record.testsGene(gene) { it == listOf(MolecularTestTarget.FUSION) }).isFalse()
+    }
+
+    @Test
+    fun `Should always return gene is tested for any target when WGS`() {
+        val record =
+            TestMolecularFactory.createMinimalTestMolecularRecord().copy(specification = null)
```
jbartletthmf, line 22: extra line break

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/PanelSpecificationsTest.kt
```diff
@@ -0,0 +1,41 @@
...
+            PanelSpecifications(mapOf(panelName to listOf(PanelGeneSpecification(geneName, listOf(MolecularTestTarget.MUTATION)))))
+        val specification = specifications.genesForPanel(panelName)
+        assertThat(specification.testsGene(geneName) { it == listOf(MolecularTestTarget.MUTATION) })
+    }
+
+    @Test
+    fun `Should throw illegal state exception when a panel name is not found`() {
+        assertThatThrownBy {
+            val specifications = PanelSpecifications(emptyMap())
+            specifications.genesForPanel("panel")
```
jbartletthmf, line 38: could be inlined

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetCoveragePredicate.kt
jbartletthmf, line 1: It would help to have some tests for these.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -13,10 +13,18 @@ import com.hartwig.actin.trial.input.datamodel.VariantTypeInput
...
+    override val gene: String, private val minExon: Int, private val maxExon: Int, private val requiredVariantType: VariantTypeInput?,
     maxTestAge: LocalDate? = null
-) : MolecularEvaluationFunction(maxTestAge) {
+) : MolecularEvaluationFunction(
+    targetCoveragePredicate = atLeast(
+        MolecularTestTarget.MUTATION,
+        messagePrefix = "Mutation in exon range $minExon to $maxExon${
+            if (requiredVariantType != null) generateRequiredVariantTypeMessage(
+                requiredVariantType
+            ) else ""
```
jbartletthmf, line 24: I was going to suggest shortening with `?.let()`, but `generateRequiredVariantTypeMessage` already returns `""` for a null type so the null check here is redundant.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -200,4 +187,22 @@ class GeneHasVariantInExonRangeOfType(
             }
         }
     }
+}
+
+private fun generateRequiredVariantTypeMessage(requiredVariantType: VariantTypeInput?): String {
+    return if (requiredVariantType == null) {
+        ""
+    } else when (requiredVariantType) {
```
jbartletthmf, line 195:
```suggestion
    return when (requiredVariantType) {
        null -> ""
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetCoveragePredicate.kt
```diff
@@ -3,47 +3,54 @@ package com.hartwig.actin.algo.evaluation.molecular
...
+    private val messagePrefix: String? = null
 ) :
     Predicate<List<MolecularTestTarget>> {
 
     override fun test(t: List<MolecularTestTarget>) = predicate.test(t)
 
-    override fun toString(): String {
-        return stringify.invoke(targets)
+    fun message(gene: String): String {
+        return "${if (messagePrefix != null) "$messagePrefix " else ""}$gene undetermined (not tested for ${stringify.invoke(targets)})"
```
jbartletthmf, line 54:
```suggestion
        return "${messagePrefix?.let { "$it "} ?: ""}$gene undetermined (not tested for ${stringify.invoke(targets)})"
```
It would be even better if the prefix included the space 🙂

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetCoveragePredicateTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+
+private const val PREFIX = "Mutation in"
+
+private const val GENE = "GENE"
+
+class TargetCoveragePredicateTest {
+
+    @Test
+    fun `Should test for all targets when any predicate`() {
```
jbartletthmf, line 15: `when all predicate`?

jbartletthmf (commented):
I think it looks better with the molecular functions passing in vals instead of overriding functions 🙂

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetCoveragePredicate.kt
pauldwolfe: yeah but then everyone who adds a prefix needs to think of that, I'd rather a tiny bit of complexity here to make the caller life easier.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasSpecificExonSkipping.kt
```diff
@@ -11,11 +11,11 @@ import com.hartwig.actin.datamodel.molecular.driver.Fusion
...
 import java.time.LocalDate
 
-class GeneHasSpecificExonSkipping(private val gene: String, private val exonToSkip: Int, maxTestAge: LocalDate? = null) :
-    MolecularEvaluationFunction(maxTestAge) {
-
-    override fun gene() = gene
-    override fun targetCoveragePredicate() = or(MolecularTestTarget.MUTATION, MolecularTestTarget.FUSION)
+class GeneHasSpecificExonSkipping(override val gene: String, private val exonToSkip: Int, maxTestAge: LocalDate? = null) :
+    MolecularEvaluationFunction(
+        targetCoveragePredicate = or(MolecularTestTarget.MUTATION, MolecularTestTarget.FUSION, messagePrefix = "Skipped exon in gene"),
```
ninajacobs, line 16: @pauldwolfe I think "Exon skipping in" is more consistent with other rules

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantWithProteinImpact.kt
```diff
@@ -23,16 +23,16 @@ private enum class VariantClassification {
 private data class VariantAndProteinImpact(val variant: Variant, val proteinImpact: String)
 
 class GeneHasVariantWithProteinImpact(
-    private val gene: String,
+    override val gene: String,
     private val allowedProteinImpacts: Set<String>,
     maxTestAge: LocalDate? = null
-) : MolecularEvaluationFunction(maxTestAge) {
+) : MolecularEvaluationFunction(
+    targetCoveragePredicate = atLeast(MolecularTestTarget.MUTATION, "Mutation with protein impact in"),
```
ninajacobs, line 30: I think we should add the specific protein impacts here too

ninajacobs (commented):
nice work with the prefix!

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInCodonTest.kt
```diff
@@ -178,5 +181,16 @@ class GeneHasVariantInCodonTest {
...
 
+    @Test
+    fun `Should evaluate undetermined with appropriate message when target coverage insufficient`() {
+        val result = function.evaluate(
+            TestPatientFactory.createMinimalTestWGSPatientRecord().copy(
+                molecularHistory = MolecularHistory(molecularTests = listOf(TestMolecularFactory.createMinimalTestPanelRecord()))
+            )
+        )
+        Assertions.assertThat(result.result).isEqualTo(EvaluationResult.UNDETERMINED)
+        Assertions.assertThat(result.undeterminedMessages).containsExactly("Mutation in codons A100, B200 in gene gene A undetermined (not tested for mutations)")
```
jbartletthmf, line 192: In all of these new tests, it would be better to import `assertThat` statically and put `.containsExactly(...)` on a new line.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/TargetCoveragePredicate.kt
jbartletthmf: The people making changes have to learn the convention either way, and when I first saw you pass in a prefix I thought "Don't we need a space there?" 😛 Of course, it's fine either way.

jbartletthmf (commented):
Just minor comments

PR #1080, opened 2025-04-14
ACTIN-1885: Improve HER2 IHC test messages

ninajacobs:


cbruel (approved):
You only have to still fix the errors in `HasPositiveHER2ExpressionByIHCTest` :)

PR #1086, opened 2025-04-16
ACTIN-1895: Rename 'Trial Matching Overview' chapter to 'Other Trial Matching Results' chapter

ninajacobs:
- See title
- Also changed the footNote (about ..Young Adult Patients..) such that it now fits on one line instead of two (checked this separately).

What do you think?


common/src/main/kotlin/com/hartwig/actin/configuration/EnvironmentConfiguration.kt
```diff
@@ -46,7 +46,7 @@ data class ReportConfiguration(
     val includeExternalTrialsInSummary: Boolean = true,
     val filterOnSOCExhaustionAndTumorType: Boolean = false,
     val includeClinicalDetailsChapter: Boolean = true,
-    val includeTrialMatchingChapter: Boolean = true,
+    val includeOtherTrialMatchingResultsChapter: Boolean = true,
```
kduyvesteyn, line 49: I wouldn't rename this param, as it makes things more complicated to upgrade. There is a ticket that cleans this all up.

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/OtherTrialMatchingResultsChapter.kt
```diff
@@ -14,7 +14,7 @@ import com.hartwig.actin.report.trial.TrialsProvider
 import com.itextpdf.kernel.geom.PageSize
 import com.itextpdf.layout.Document
 
-class TrialMatchingChapter(
+class OtherTrialMatchingResultsChapter(
```
kduyvesteyn, line 17: I don't really like the name and also don't think its correct with every permutation of the environment config but I can't think of a better suggestion now

common/src/main/kotlin/com/hartwig/actin/configuration/EnvironmentConfiguration.kt
ninajacobs: check!

kduyvesteyn (approved):
Looks good

PR #1091, opened 2025-04-18
ACTIN-1904: Reformat trial table when no warning are shown

kduyvesteyn:
In addition, filtered external trials are shown at the top of the other trial details chapter.


report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/TrialMatchingOtherResultsChapter.kt
```diff
@@ -70,7 +70,7 @@ class TrialMatchingOtherResultsChapter(
             report.config.countryOfReference
         )
 
-        return localTrialGenerators + listOfNotNull(localExternalTrialGenerator, nonLocalTrialGenerator) + filteredTrialGenerator
+        return listOfNotNull(localExternalTrialGenerator, nonLocalTrialGenerator) + filteredTrialGenerator + localTrialGenerators
```
cbruel, line 73: Besides moving the filtered trials table, I think it would also be cleaner if we only show this table if there are filtered trials? Showing the table with "None" on the report, doesn't add much / is just noise. I would just not show it if there are 0 trials?

kduyvesteyn: Agreed, like the idea, have added!

PR #1092, opened 2025-04-22
ACTIN-1908: Add rule MEETS_SPECIFIC_CRITERIA_REGARDING_METASTASES

cbruel:


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/MeetsSpecificCriteriaRegardingMetastasesTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun `Should fail when no metastatic cancer`() {
+        val alwaysFailsMetastaticCancerEvaluation = mockk<HasMetastaticCancer> {
+            every { evaluate(any()) } returns EvaluationFactory.fail("no metastatic cancer")
+        }
+        val function = MeetsSpecificCriteriaRegardingMetastases(alwaysFailsMetastaticCancerEvaluation)
+        EvaluationAssert.assertEvaluation(EvaluationResult.FAIL, function.evaluate(patientRecord))
+    }
+
+    @Test
+    fun `Should be undetermined when tumor stage unknown`() {
```
jbartletthmf, line 24: "tumor stage unknown" refers to the implementation details of `HasMetastaticCancer`, which isn't being tested here. I would rename this to something like:
```suggestion
    fun `Should be undetermined when unknown if patient has metastatic cancer`() {
```

jbartletthmf (approved):

PR #1093, opened 2025-04-22
ACTIN-1863: Correct standard lesion extraction

ninajacobs:
**Standard data extraction:**
- Only ever interpret brain, CNS, liver or bone lesions in case of standard extraction (others are only null)

**lesionMinCount:**
- Rename lesionCount variables to lesionMinCount variables (this name is more accurate to the data)
- Determine lesionMinCount variables for TumorDetailsExtraction as well (was previously done only for standard extraction)
- Set lesionMinCount variables to null in case of no lesions (instead of 0) to better match the variable name
- Use lesionMinCount adequately in algo (HasMinimumLesionsInSpecificBodyLocation)

**FYI:**
I was unsure whether hasSuspectLesions should better be false or null in case unknown; but since in tumorDetailsExtractor it is a bit unclear if we can say the answer is 'false' if there are none known (since we don't ask for it explicitly) I preferred to use 'null' now in these cases, and aimed to ensure we handle 'null' for suspect lesions consistently in algo. This means essentially that for 'undetermined' outcome we rely on hasLesions being null and ignore hasSuspectLesions being null, see eg TumorMetastasisEvaluator or HasMinimumLesionsInSpecificBodyLocation.

Once we do a larger review of the lesion data model (TBD with Paul: e.g. consider not having a list of booleans but a lesion class with 'active' or 'suspect' as fields or a status enum) I think this is okay.

**// UPDATE 30-04-2025**
- Removed minCount variables since they are not adding much anymore
- added hasPrimaryBrainOrGlioma check also for standard extractor
- PR feedback


clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
```diff
@@ -143,11 +143,29 @@ class TumorDetailsExtractorTest {
...
+                TestCurationFactory.curationDatabase(
+                    lesionLocationConfig(LesionLocationCategory.LIVER).copy(suspected = true)
+                ), TestCurationFactory.curationDatabase(),
+                tumorStageDeriver
+            ), questionnaire, expected
+        )
+    }
+
+    @Test
+    fun `Should override only the aspect that needs to be overridden and leave the other aspect of pair intact`() {
```
andreia-hartwig, line 160: The name is a bit generic, would be easier to understand if it reflects what is actually being tested? 
Something like `Should override has suspected liver lesions when liver lesions are defined as other lesions and curated as suspected.`

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
```diff
@@ -186,22 +186,23 @@ class TumorDetailsExtractor(
...
+        val hasLesions = lesionsConfig.any { it.suspected != true }.takeIf { it }
+        val hasSuspectedLesions = lesionsConfig.any { it.suspected == true }.takeIf { it }
 
-        if ((hasOtherLesions == true || hasSuspectedOtherLesions == true) && hasLesion == false) {
+        if ((hasLesions == true || hasSuspectedLesions == true) && hasLesionsQuestionnaire != true) {
             logger.debug("  Overriding presence of ${lesionLocationCategory.name.lowercase()} lesions")
         }
-        return Pair(hasOtherLesions, hasSuspectedOtherLesions)
+
+        return if (hasLesions == true) Pair(true, hasSuspectedLesions) else Pair(hasLesionsQuestionnaire, hasSuspectedLesions)
```
andreia-hartwig, line 205: You can avoid a bit of repetition by doing: 

`return (if (hasLesions == true) true else hasLesionsQuestionnaire) to hasSuspectedLesions`

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
ninajacobs: Thanks! Got to "Should override has suspected liver lesions when suspected liver lesions are curated but not override known lesions"

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
ninajacobs: Good point. In my latest commit I have now Triple instead of Pair. I believe I can't work with 'to' then right? Do you know something cleaner for below?

return if (hasLesions == true) Triple(true, hasSuspectedLesions, minCount) else Triple(
            hasLesionsQuestionnaire,
            hasSuspectedLesions,
            minCount
        )

andreia-hartwig: You could do 

`return Triple(if (hasLesions == true) true else hasLesionsQuestionnaire, hasSuspectedLesions, minCount)
`

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
```diff
@@ -50,59 +50,65 @@ class TumorDetailsExtractor(
         val otherLesions = filterCurateOtherLesions(curatedOtherLesions?.configs)
         val otherSuspectedLesions = filterCurateOtherLesions(curatedOtherLesions?.configs, true)
 
-
         val hasBrainOrGliomaTumor = primaryTumorDetails.primaryTumorLocation == "Brain" ||
                 primaryTumorDetails.primaryTumorType == "Glioma"
 
-        val (hasBrainLesions, hasSuspectedBrainLesions) = if (hasBrainOrGliomaTumor) {
-            Pair(false, false)
+        val (hasBrainLesions, hasSuspectedBrainLesions, brainLesionsMinCount) = if (hasBrainOrGliomaTumor) {
```
pauldwolfe, line 56: feels like we could centralize all the checks for "hasBrainOrGliomaTumor" in one place. Maybe it could be at the end and we could overwrite contradictory fields at the end of this function before returning? (and perhaps it would be interesting to log a warning or error if primary tumor was brain and we determine brain or cns lesions?)

pauldwolfe (commented):
On the "min count" concept in general, do we ever really expect to get accurate lesion counts? Do we expect an instance of HasMinLesionInSpecificSite with more than 1 count? (i don't see any currently used instance of this rule).

If not maybe we don't need the count?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
```diff
@@ -67,61 +65,52 @@ class StandardTumorDetailsExtractor(
...
+        hasSuspectedLiverLesions = null,
+        liverLesionsMinCount = determineLesionsMinCount(lesions, LesionLocationCategory.LIVER),
+        hasLungLesions = null,
+        hasSuspectedLungLesions = null,
+        lungLesionsMinCount = null,
+        hasLymphNodeLesions = null,
+        hasSuspectedLymphNodeLesions = null,
+        lymphNodeLesionsMinCount = null,
+        otherLesions = null,
+        otherSuspectedLesions = null,
```
cbruel, line 96: do we really need to set all these to null here? it's already defaulted null in `TumorDetails`, so I wouldn't add them, just gives noise I think

cbruel (commented):
Some minor comments + most importantly what we discussed in Slack :)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
ninajacobs: Talked with @cbruel and we actually don't think we should correct for hasBrainOrGliomaTumor in the 'lesions' part, since this part is about 'lesions' and not about 'metastases'. Instead, we'd propose we check for primary tumor presence in the algo part.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
```diff
@@ -67,61 +65,52 @@ class StandardTumorDetailsExtractor(
...
-    }
-
-    private fun extractLesions(patientRecord: ProvidedPatientRecord): List<CurationResponse<LesionLocationConfig>> {
-        val patientId = patientRecord.patientDetails.hashedId
-        val lesionsFromLesionList = extractFromLesionList(patientRecord)
-        val lesionsFromRadiologyReport = extractFromRadiologyReport(patientRecord.tumorDetails.lesionSite, patientId)
-        val lesionsFromOtherConditions = extractFromSecondarySource(patientId, patientRecord.priorOtherConditions) { it.name }
-        val lesionsFromTreatmentHistory = extractFromSecondarySource(patientId, patientRecord.treatmentHistory) { it.treatmentName }
-        return lesionsFromLesionList + lesionsFromRadiologyReport + lesionsFromOtherConditions + lesionsFromTreatmentHistory
+    private fun determineLesionsMinCount(
```
andreia-hartwig, line 108: determineLesionsMinCount always returns 1 when there are lesions or null if there aren't, so it's not returns a minimum count, it's actually just always returning 1 if there are lesions. Perhaps the name of the function can be adjusted or evaluate if minCount is needed when always 1.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
```diff
@@ -354,6 +385,80 @@ class TumorDetailsExtractorTest {
         )
     }
 
+    @Test
+    fun `Should override has suspected liver lesions when suspected liver lesions are curated but then not override known lesions`() {
+        assertThat(baseTumor.hasLiverLesions).isNull()
```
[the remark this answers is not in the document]
ninajacobs: I'm not fully sure what you mean, should I not check for null here? I aimed to make it similar to how it was done in other tests in this class, is this a different situation?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasMinimumLesionsInSpecificBodyLocation.kt
```diff
@@ -13,22 +13,22 @@ class HasMinimumLesionsInSpecificBodyLocation(
...
+                BodyLocationCategory.LIVER -> Triple(hasLiverLesions, hasSuspectedLiverLesions, liverLesionsMinCount)
+                BodyLocationCategory.LUNG -> Triple(hasLungLesions, hasSuspectedLungLesions, lungLesionsMinCount)
+                BodyLocationCategory.LYMPH_NODE -> Triple(hasLymphNodeLesions, hasSuspectedLymphNodeLesions, lymphNodeLesionsMinCount)
                 else -> return EvaluationFactory.undetermined("Undetermined if patient has $messageEnding")
             }
         }
 
         return when {
-            (count ?: if (hasLesions == true) 1 else 0) >= minimumLesions -> EvaluationFactory.pass("Has $messageEnding")
+            (minCount ?: 0) >= minimumLesions -> EvaluationFactory.pass("Has $messageEnding")
```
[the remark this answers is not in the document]
ninajacobs: check

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
```diff
@@ -187,23 +193,24 @@ class TumorDetailsExtractor(
     private fun determineLesionPresence(
         lesionLocationConfigMap: Map<LesionLocationCategory?, List<LesionLocationConfig>>?,
         lesionLocationCategory: LesionLocationCategory,
-        hasLesion: Boolean? = null
-    ): Pair<Boolean?, Boolean?> {
+        hasLesionsQuestionnaire: Boolean? = null
+    ): Triple<Boolean?, Boolean?, Int?> {
```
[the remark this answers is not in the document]
ninajacobs: This is now back to a double (now I removed minCount variableS), as it was before. Is it OK to keep it like as-is now, given it's a pair again?

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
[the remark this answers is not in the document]
andreia-hartwig: The standard would be that you ensure that hasLiverLesions is null by doing the following at the start of the method, instead of doing an assertThat
baseTumor.hasLiverLesions = null

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasMinimumLesionsInSpecificBodyLocation.kt
```diff
@@ -13,22 +13,22 @@ class HasMinimumLesionsInSpecificBodyLocation(
     override fun evaluate(record: PatientRecord): Evaluation {
```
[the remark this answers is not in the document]
ninajacobs: check

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
ninajacobs: (also outdated now)

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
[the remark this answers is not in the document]
ninajacobs: Like this then?

val tumor = baseTumor.copy(hasLiverLesions = null, hasSuspectedLiverLesions = null)

I assume this was not done in all tests because it is already null to start with, but if this is preferred I can change for all tests?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
ninajacobs: I reorganized this a bit now, would you have a look? "ACTIN-1863: Reorganize primary brain evaluation + add a test".

Also very happy to hear if my added test makes sense or if I could have done this simpler :)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
cbruel: will leave this technical question for @pauldwolfe :)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/tumor/TumorStageDeriver.kt
```diff
@@ -91,20 +87,7 @@ class TumorStageDeriver(private val derivationRules: Map<Predicate<TumorDetails>
...
         private fun evaluateMetastases(hasLesions: Boolean, tumor: TumorDetails, doidToMatch: String, doidModel: DoidModel): Boolean {
-            // Currently only for lung cancer multiple lesions are resolved to stage III/IV
-            return if (checkingLungMetastasesForLungCancer(doidModel, doidToMatch, tumor.doids) && tumor.lungLesionsCount != null) {
-                tumor.lungLesionsCount!! >= 2
-            } else {
-                (hasLesions) && !DoidEvaluationFunctions.isOfDoidType(doidModel, tumor.doids, doidToMatch)
-            }
-        }
-
-        private fun checkingLungMetastasesForLungCancer(doidModel: DoidModel, doidToMatch: String, tumorDoids: Set<String>?): Boolean {
```
cbruel, line 102: I've added this logic in https://hartwigmedical.atlassian.net/browse/ACTIN-683, I don't think we should remove this?
I do understand you removed it because you removed all the count booleans, but looking at this PR we once added those to be able to do this evaluation. Maybe only re-introduce `lungLesionsCount` boolean, so we can still do this evaluation?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
pauldwolfe: Yeah... its debatable I guess. I'm more in the camp that finds this noisy.  But I think in this case it's because that datamodel could be improved. If tumor details for instance had a "list of lesions" we wouldn't need to set things to null in either place. 

If it was me, I'd remove these and let the defaults handle it.  But if you feel strongly this makes it more explicit, also fine, hopefully we can find a way to make obsolete in the future.

ninajacobs:
2 vs 1, I'll remove explicit nulls :)

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/tumor/TumorStageDeriver.kt
ninajacobs: As discussed, I won't do this.

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractorTest.kt
```diff
@@ -414,6 +432,39 @@ class TumorDetailsExtractorTest {
         assertThat(evaluation.lesionLocationEvaluatedInputs).isEqualTo(lesions.map(String::lowercase).toSet())
     }
 
+    @Test
+    fun `Should set brain and CNS lesion variables to false in case of primary brain tumor`() {
+        val tumorDetailsExtractor =
+            spyk(TumorDetailsExtractor(TestCurationFactory.curationDatabase(), TestCurationFactory.curationDatabase(), tumorStageDeriver))
```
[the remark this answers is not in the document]
ninajacobs: Thanks! I tried to do it the other way now for StandardTumorDetailsExtractorTest. Could you review if I did this correctly/efficiently?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/TumorMetastasisEvaluatorTest.kt
```diff
@@ -26,7 +26,7 @@ class TumorMetastasisEvaluatorTest {
     fun `Should warn when only suspected metastasis boolean is true`() {
         listOf(false, null).forEach { hasKnownLesion ->
             val warn = TumorMetastasisEvaluator.evaluate(hasKnownLesion, true, METASTASIS_TYPE)
-            val message = "Bone metastases present but only suspected lesions"
+            val message = "Has suspected bone metastases hence uncertain if actually has bone metastases"
```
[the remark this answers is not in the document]
ninajacobs: better! thanks

PR #1096, opened 2025-04-27
ACTIN-1911 Add hospital patient id and additional pathology data

andreia-hartwig:


kduyvesteyn (approved):
Github won't let me review this properly, so I will just approve.

A few things of note:
 - Most IHC classes start with "IHC" but some start with "Ihc" (e.g. `IhcTestFilter`), would make this consistent
 - The `PathologyReport` will need some improvements but can be separate tasks (.e.g `diagnosis` should follow a standardized datamodel rather than string, dates could probably be done smarter)
 -  `hospitalPatientId` is better to rename to `sourceId` which is clear enough in context of patient details. From the perspective of ACTIN, this `sourceId` is simple an ID provided by the source and in which we are supposed to communicate back about. Some other user of ACTIN may provide IDs that are not hospital patient IDs but something else (we could also use this source ID in the context of Hartwig database entries for example). 
 - The IHC tests are now last on the molecular details page rather than first. Not sure what you discussed with @ninajacobs about this, if that change has been agreed then fine with me.

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
```diff
@@ -42,86 +43,86 @@ class MolecularDetailsChapter(
...
-                    Cells.createContent(
-                        ("No successful OncoAct WGS and/or tumor NGS panel could be "
-                                + "performed on the submitted biopsy (insufficient quality for reporting)")
-                    )
-                )
+        val table = Tables.createSingleColWithWidth(contentWidth())
+        for ((pathologyReport, tests) in groupedByPathologyReport) {
+            pathologyReport ?: groupedByPathologyReport.keys.takeIf { it.size > 1 }?.let {
+                table.addCell(Cells.createTitle("Other Tests"))
             }
```
jbartletthmf, line 75: What is this code supposed to be doing? It seems unusual to only add a cell here when `pathologyReport` is null and the map has >1 key. I think it would be clearer if you performed the `topTable.addCell` from line 106 here as well since the cell is added in both (all?) cases.

report/src/main/kotlin/com/hartwig/actin/report/pdf/components/SidePanel.kt
```diff
@@ -9,16 +9,20 @@ import com.itextpdf.kernel.pdf.canvas.PdfCanvas
 import com.itextpdf.layout.Canvas
 import com.itextpdf.layout.element.Div
 import com.itextpdf.layout.element.Paragraph
+import com.itextpdf.layout.element.Text
 import java.time.LocalDate
-import java.util.Locale
+import java.util.*
```
jbartletthmf, line 14: star import

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -0,0 +1,75 @@
...
+                    if (report.isSourceInternal) {
+                        add(Text(" (Collection date: ")).add(Text(date(report.tissueDate)).addStyle(Styles.tableHighlightStyle()))
+                        add(Text(", Authorization date: ")).add(Text(date(report.authorisationDate)).addStyle(Styles.tableHighlightStyle()))
+                    } else {
+                        add(Text(" (Report date: ")).add(Text(date(report.externalDate)).addStyle(Styles.tableHighlightStyle()))
+                    }
+                }
+                .add(Text(", Diagnosis: "))
+                .add(Text(report.diagnosis).addStyle(Styles.tableHighlightStyle())).add(Text(")"))
+        ).addStyle(Styles.tableContentStyle())
```
jbartletthmf, line 43: I find `apply` a little unintuitive here, and I think it's more "functional" to create the full list of text elements and add them at once:
```suggestion
        Cells.create(
            Paragraph().addAll(
                listOfNotNull(
                    prefix?.let {
                        listOf(
                            Text(prefix).addStyle(prefixStyle),
                            Text(" - ").addStyle(Styles.tableHighlightStyle())
                        )
                    },
                    listOf(Text(report.tissueId?.uppercase()).addStyle(Styles.tableTitleStyle())),
                    if (report.isSourceInternal) {
                        listOf(
                            Text(" (Collection date: "),
                            Text(date(report.tissueDate)).addStyle(Styles.tableHighlightStyle()),
                            Text(", Authorization date: "),
                            Text(date(report.authorisationDate)).addStyle(Styles.tableHighlightStyle())
                        )
                    } else {
                        listOf(
                            Text(" (Report date: "),
                            Text(date(report.externalDate)).addStyle(Styles.tableHighlightStyle())
                        )
                    },
                    listOf(
                        Text(", Diagnosis: "),
                        Text(report.diagnosis).addStyle(Styles.tableHighlightStyle()),
                        Text(")")
                    )
                ).flatten()
            )
        ).addStyle(Styles.tableContentStyle())
```

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -0,0 +1,75 @@
...
+        molecularTests: List<MolecularTest>,
+        ihcTests: List<IHCTest>,
+        pathologyReports: List<PathologyReport>?
+    ): Map<PathologyReport?, Triple<List<MolecularRecord>, List<MolecularTest>, List<IHCTest>>> {
+
+        val reportDates = pathologyReports.orEmpty().map { it.date }.toSet()
+
+        val (matchedOrangeResults, unmatchedOrangeResults) = orangeMolecularRecords.partition { it.date in reportDates }
+        val (matchedMolecularTests, unmatchedMolecularTests) = molecularTests.partition { it.date in reportDates }
+        val (matchedIhcTests, unmatchedIHCTests) = ihcTests.partition { it.measureDate in reportDates }
```
jbartletthmf, line 56: For each of these, you could also consider something like `tests.groupBy { it.date.takeIf(reportDates::contains) }`. That way, you could look up each corresponding result set below using the grouped map instead of filtering each time. The unmatched entries would have key `null`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasAvailablePDL1Status.kt
```diff
@@ -8,7 +8,7 @@ import com.hartwig.actin.datamodel.algo.Evaluation
 class HasAvailablePDL1Status : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        return if (IhcTestFilter.allPDL1Tests(record.priorIHCTests).isNotEmpty()) {
+        return if (IHCTestFilter.allPDL1Tests(record.ihcTests).isNotEmpty()) {
```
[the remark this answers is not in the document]
andreia-hartwig: Will follow IhcTest****

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
```diff
@@ -42,86 +43,86 @@ class MolecularDetailsChapter(
...
+        val orangeMolecularRecord = report.patientRecord.molecularHistory.latestOrangeMolecularRecord()
+        val externalPanelResults = report.patientRecord.molecularHistory.molecularTests.filter { it.experimentType == ExperimentType.PANEL }
+        val filteredIhcTests = IHCTestFilter.mostRecentOrUnknownDateIhcTests(report.patientRecord.ihcTests).toList()
+        val groupedByPathologyReport = PathologyReportFunctions.groupTestsByPathologyReport(
+            listOfNotNull(orangeMolecularRecord),
+            externalPanelResults,
+            filteredIhcTests,
+            report.patientRecord.pathologyReports
+        ).filterValues { (orangeTests, molecularTest, ihcTests) ->
+            orangeTests.isNotEmpty() || molecularTest.isNotEmpty() || ihcTests.isNotEmpty()
```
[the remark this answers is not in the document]
andreia-hartwig: True, initially it was not part of the groupTestsByPathologyReport since the orangeTests were handled separately but now that all types of tests are handled by this function it's safe to filter.

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
jbartletthmf: Thanks for clarifying. What do you think of the following to make the title decision more explicit (I moved the title for known reports here as well):
```suggestion
            when {
                pathologyReport != null -> {
                    table.addCell(Cells.create(PathologyReportFunctions.getPathologyReportSummary(report = pathologyReport)))
                }
                groupedByPathologyReport.size > 1 -> {
                    table.addCell(Cells.createTitle("Other Tests"))
                }
                else -> {} // No title needed: all tests are ungrouped
            }
```
You could even drop the `else` case; I just get a little wary about conditional cell additions since it can lead to malformed tables in some cases.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/IHCTest.kt
```diff
@@ -14,4 +13,6 @@ data class PriorIHCTest(
     val scoreValue: Double? = null,
     val scoreValueUnit: String? = null,
     val impliesPotentialIndeterminateStatus: Boolean = false
-)
+) {
+    val test: String = IHC_TEST_TYPE
```
pauldwolfe, line 17: I think you can just get rid of this entirely and set the test = "IHC" in the IHCTestInterpreter

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/PathologyReport.kt
```diff
@@ -0,0 +1,22 @@
...
+    val externalDate: LocalDate? = null,
+    val tissueDate: LocalDate? = null,
+    val authorisationDate: LocalDate? = null,
+    val report: String
+) {
+    val isSourceInternal: Boolean
+        get() = source.equals(INTERNAL_SOURCE, ignoreCase = true)
+
+    companion object {
+        private const val INTERNAL_SOURCE = "internal"
```
pauldwolfe, line 20: doesn't need to be a companion object... what is "source" internal mean in this context? Is it coming from HMF?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/PathologyReport.kt
```diff
@@ -0,0 +1,22 @@
+package com.hartwig.actin.datamodel.clinical
+
+import java.time.LocalDate
+
+data class PathologyReport(
```
pauldwolfe, line 5: are all these fields used downstream in the report? Some of are confusing on first glance. Report requested by whom? What is the authorisation date?  Not a big problem per se, but feel a bit like NKI stuff leaking into our common datamodel.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/IHCExtractor.kt
```diff
@@ -18,8 +18,8 @@ class IHCExtractor : MolecularExtractor<PriorIHCTest, IHCExtraction> {
             }.filter { it.fusionPositiveGenes.isNotEmpty() || it.fusionNegativeGenes.isNotEmpty() }
     }
 
-    private fun ihcFusionGenes(priorIhcTests: List<PriorIHCTest>, scoreText: String): Set<String> {
-        return priorIhcTests.filter { it.test == "IHC" && it.item in IHC_FUSION_GENES && it.scoreText == scoreText }
+    private fun ihcFusionGenes(ihcTests: List<IHCTest>, scoreText: String): Set<String> {
+        return ihcTests.filter { it.test == "IHC" && it.item in IHC_FUSION_GENES && it.scoreText == scoreText }
```
pauldwolfe, line 22: the check for test == "IHC" can be removed?

report/src/main/kotlin/com/hartwig/actin/report/interpretation/MolecularTestInterpretation.kt
```diff
@@ -0,0 +1,31 @@
+package com.hartwig.actin.report.interpretation
+
+import java.time.LocalDate
+
+data class MolecularTestInterpretation(
```
pauldwolfe, line 5: These classes and the following all seem only applicable to IHC tests but names indicate they are for all molecular tests.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/PathologyReport.kt
andreia-hartwig: This source is coming from NKI, where pathology reports have two possible sources - internal or external.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/PathologyReport.kt
andreia-hartwig: Yes most of them are used.  Authorisation date is the latest authorization date for the pathology report, this one is shown in the report. 

Report requested is NKI specific and not actually used so we can ignore it and remove from the data model from now.

PR #1100, opened 2025-04-28
ACTIN-1925: actin transvar image locale fix

kzuberihmf:


kzuberihmf:
Checking the 1.0.0 image, the locale was utf8.

PR #1101, opened 2025-04-29
ACTIN-1924: Split MedicationCategory "Gonadorelin" into "Gonadotropin agonists" and "Gonadotropin antagonists"

cbruel:
We somehow once named this category "Gonadorelin" but that is not a suitable name for it (gonadorelin is an example of an "gonadotropin agonist"). Splitting it is cleaner.

(Encountered this while curating a trial)


ninajacobs (approved):
Looks good! Haven't reviewed ATC codes specifically but if you are certain then I am!

PR #1102, opened 2025-04-29
ACTIN-1056: Overwrite hotspot and driver likelihood information from ORANGE with CKB evidence

cbruel:
PR includes:
- Use the known events from SERVE to annotate external molecular tests + use the CKB part of the known events to overwrite hotspots from ORANGE
- Change display names for driver events (according to email thread _'Term 'VUS' verwijderd uit 'molecular drivers' table'_)

@ninajacobs Mainly added you to review the changes to `MolecularDriverEntryFactory` because they are based on the email you have send / the table you have made. Always welcome, to review the other parts of course :)


kduyvesteyn (changes requested):
We should be able to isolate the differences between CKB and non-CKB inside `EvidenceDatabase` and not expose to the outside users. See email for more context.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolver.kt
kduyvesteyn, line 1: Some comments on this class:
 - I would make this class return "actin" objects rather than "serve" objects consistently now that you have introduced `VariantAlteration`. If not mistaken this means that the functions that now return a SERVE gene alteration should return an ACTIN gene alteration.
 - I would rename all the alteration functions `alterationForX` (e.g. `alterationForVariant`, `alterationForCopyNumber`. I tried to come up with a better term for "alteration" but can't come up with one right now. 
 - I would not make the `filteredKnownEvents` a property of this class but rather a function inside this class. Alternatively you could consider separating the 2 groups into `primaryKnownEvents` and `secondaryKnownEvents` and make them mutually exclusive, maybe that's nicer actually since otherwise we'd call the filter function a lot of times! Related - this would imply renaming `KNOWN_EVENT_SOURCES` to `PRIMARY_KNOWN_EVENT_SOURCE` (I think I would not make it a set anymore).

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/driver/VariantAlteration.kt
```diff
@@ -0,0 +1,5 @@
+package com.hartwig.actin.datamodel.molecular.driver
+
+interface VariantAlteration : GeneAlteration {
```
kduyvesteyn, line 3: I would make this a "data class" rather than an interface. A data class is typically meant for types where you make instances of (which you do), while an interface is usually meant to define to set of properties/functions that are common across different (data) classes.

molecular/src/main/kotlin/com/hartwig/actin/molecular/orange/MolecularRecordAnnotator.kt
```diff
@@ -93,6 +101,16 @@ class MolecularRecordAnnotator(private val evidenceDatabase: EvidenceDatabase) :
...
+    fun reannotateDriverLikelihood(variants: List<Variant>): List<Variant> {
+        val variantsByGene = variants.groupBy { it.gene }
+        return variantsByGene.flatMap { (_, geneVariants) ->
+            val maxDriverLikelihood = geneVariants.minOfOrNull { it.driverLikelihood?.ordinal ?: DriverLikelihood.entries.size }
+            geneVariants.map { variant ->
+                variant.copy(driverLikelihood = maxDriverLikelihood?.let { DriverLikelihood.entries[it] })
+            }
+        }
+    }
+
```
kduyvesteyn, line 113: Haven't reviewed in detail but one addition I would make at least is to only overwrite the driver likelihood if the driver likelihood is not-null to begin with. Now you overwrite _every_ variant in a gene, also those affecting introns only and these get driver likelihood null and are non-reportable. 

Rather than copying the exact logic from purple, I think it would work to say "overwrite DL in case the original algo set a DL in the first place".

kduyvesteyn (commented):
@cbruel This is definitely moving in the right direction!

Have mostly reviewed on a higher level and have a few suggestions that are more minor than previous round. Will do a more detailed review (e.g. tests) once you've addressed those.

molecular/src/main/kotlin/com/hartwig/actin/molecular/orange/MolecularRecordAnnotator.kt
cbruel: Thanks, useful to know!

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolverFactory.kt
```diff
@@ -13,24 +13,29 @@ import com.hartwig.serve.datamodel.molecular.range.KnownExon
 
 object KnownEventResolverFactory {
 
-    val KNOWN_EVENT_SOURCES = setOf(ActionabilityConstants.EVIDENCE_SOURCE)
+    val PRIMARY_KNOWN_EVENT_SOURCE = ActionabilityConstants.EVIDENCE_SOURCE
 
     fun create(knownEvents: KnownEvents): KnownEventResolver {
-        return KnownEventResolver(filterKnownEvents(knownEvents), GeneAggregator.aggregate(knownEvents.genes()))
+        val primaryKnownEvents = filterKnownEvents(knownEvents, filterPrimaryEvents = true)
+        val secondaryKnownEvents = filterKnownEvents(knownEvents, filterPrimaryEvents = false)
```
kduyvesteyn, line 20: For me "include" sounds like a better term than "filter" since filter implies you remove objects or you don't. In the implementation you _include_ objects yes or no

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolverFactory.kt
```diff
@@ -13,24 +13,29 @@ import com.hartwig.serve.datamodel.molecular.range.KnownExon
...
+        return KnownEventResolver(primaryKnownEvents, secondaryKnownEvents, GeneAggregator.aggregate(knownEvents.genes()))
     }
 
-    internal fun filterKnownEvents(knownEvents: KnownEvents): KnownEvents {
+    fun filterKnownEvents(
+        knownEvents: KnownEvents,
+        filterPrimaryEvents: Boolean
+    ): KnownEvents {
```
kduyvesteyn, line 27: Minor but I'd fit this onto a single line

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/EvidenceDatabase.kt
kduyvesteyn, line 1: You changed all the alteration functions to return not-null objects. I don't have a strong opinion on that, but at least I would make it consistent. Now, the fusion lookup can still return null, I would change that to return an object with unknown impact and null associatedWithDrugResistance

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolver.kt
```diff
@@ -2,45 +2,73 @@ package com.hartwig.actin.molecular.evidence.known
...
+    private val secondaryKnownEvents: KnownEvents,
+    private val aggregatedKnownGenes: Set<KnownGene>
+) {
 
-    fun resolveForVariant(variantMatchCriteria: VariantMatchCriteria): GeneAlteration? {
-        return findHotspot(knownEvents.hotspots(), variantMatchCriteria)
-            ?: findCodon(knownEvents.codons(), variantMatchCriteria)
-            ?: findExon(knownEvents.exons(), variantMatchCriteria)
+    fun resolveForVariant(variantMatchCriteria: VariantMatchCriteria): VariantAlteration {
+        val ckbAlteration = findHotspot(primaryKnownEvents.hotspots(), variantMatchCriteria)
```
kduyvesteyn, line 27: Variable should be called `primaryAlteration`. Code doesn't know this is typically CKB

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolver.kt
```diff
@@ -2,45 +2,73 @@ package com.hartwig.actin.molecular.evidence.known
...
+            ?: findCodon(primaryKnownEvents.codons(), variantMatchCriteria)
+            ?: findExon(primaryKnownEvents.exons(), variantMatchCriteria)
             ?: GeneLookup.find(aggregatedKnownGenes, variantMatchCriteria.gene)
+
+        val geneAlteration = GeneAlterationFactory.convertAlteration(variantMatchCriteria.gene, ckbAlteration)
+        val isHotspot =
+            HotspotFunctions.isHotspot(ckbAlteration) || findHotspot(secondaryKnownEvents.hotspots(), variantMatchCriteria) != null
```
kduyvesteyn, line 34: I am not sure how to improve this but in this line of code there are a lot of "hidden" assumptions. Key one being that the primary known event hotspots apparently are not all hotspots, while the secondary are. Also, you assume that the secondary hotspots never contain actual gene alteration information. 

I am fine with leaving as is, but we could maybe adjust `HotspotFunctions` to be used for every hotspot and then implement "if source contains primary -> check for protein effect, else return true". This way the function can be used for any hotspot

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/EvidenceDatabaseTest.kt
```diff
@@ -28,32 +28,32 @@ class EvidenceDatabaseTest {
             driverLikelihood = DriverLikelihood.HIGH,
             isReportable = true
         )
-        assertThat(database.geneAlterationForVariant(variant)).isNotNull()
+        assertThat(database.alterationForVariant(variant)).isNotNull()
```
kduyvesteyn, line 31: This test (and similar tests in this class) are meaningless now that evidence database always returns not-null anyways

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolverFactoryTest.kt
```diff
@@ -14,18 +15,18 @@ class KnownEventResolverFactoryTest {
...
         val hotspot3 = TestServeKnownFactory.hotspotBuilder().addSources(Knowledgebase.UNKNOWN).build()
         val knownEvents = ImmutableKnownEvents.builder().addHotspots(hotspot1, hotspot2, hotspot3).build()
 
-        val filtered = KnownEventResolverFactory.filterKnownEvents(knownEvents)
-        assertThat(filtered.hotspots().size.toLong()).isEqualTo(2)
-        assertThat(filtered.hotspots().contains(hotspot1)).isTrue()
-        assertThat(filtered.hotspots().contains(hotspot2)).isTrue()
-    }
+        val primaryKnownEvents = KnownEventResolverFactory.filterKnownEvents(knownEvents, true)
+        assertThat(primaryKnownEvents.hotspots().size.toLong()).isEqualTo(2)
```
kduyvesteyn, line 24: You could replace this with ```assertThat(primaryKnownEvents.hotspots()).hasSize(2)```

Or even better, you can replace the 3 lines with 
```
assertThat(primaryKnownEvents.hotspots()).containsExactlyInAnyOrder(hotspot1, hotspot2)
```
since this also asserts the size implicitly

common/src/test/kotlin/com/hartwig/actin/molecular/evidence/known/KnownEventResolverTest.kt
kduyvesteyn, line 1: I think you can clean up some not-null asserts in these tests since the code never returns null.

Also, the name of the knowledgebase is "DoCM" rather than "DOCM" (Database of Cancer Mutations)

molecular/src/main/kotlin/com/hartwig/actin/molecular/hotspotcomparison/HotspotComparator.kt
```diff
@@ -16,9 +15,10 @@ object HotspotComparator {
         return orange.purple().allSomaticVariants().filter { it.gene() in knownGenes }
             .mapNotNull { variant ->
                 val criteria = createVariantCriteria(variant)
-                val knownEventResolver = KnownEventResolverFactory.create(serveRecord.knownEvents())
+                val knownEventResolver =
+                    KnownEventResolverFactory.create(KnownEventResolverFactory.filterKnownEvents(serveRecord.knownEvents(), true))
                 val serveGeneAlteration = knownEventResolver.resolveForVariant(criteria)
```
kduyvesteyn, line 20: Should be renamed to serveVariantAlteration

molecular/src/main/kotlin/com/hartwig/actin/molecular/hotspotcomparison/HotspotComparator.kt
kduyvesteyn, line 1: Minor but the term "Comparator" is typically used for classes implementing `Comparator` interface (used for sorting etc). I would rename this class to `HotspotEvaluator` to avoid this confusion.

molecular/src/main/kotlin/com/hartwig/actin/molecular/orange/MolecularRecordAnnotator.kt
```diff
@@ -93,9 +101,18 @@ class MolecularRecordAnnotator(private val evidenceDatabase: EvidenceDatabase) :
...
+    fun reannotateDriverLikelihood(variants: List<Variant>): List<Variant> {
+        val variantsByGene = variants.groupBy { it.gene }
+        return variantsByGene.flatMap { (_, geneVariants) ->
+            val maxDriverLikelihood = geneVariants.minOfOrNull { it.driverLikelihood?.ordinal ?: DriverLikelihood.entries.size }
+            geneVariants.map { variant ->
+                variant.copy(driverLikelihood = maxDriverLikelihood?.let { if (variant.driverLikelihood != null) DriverLikelihood.entries[it] else null })
+            }
+        }
+    }
+
```
kduyvesteyn, line 113: If not mistaken this logic relies on the fact that the ordinal for "HIGH" is lower than the ordinal for "MEDIUM" which is again lower than the ordinal for "LOW"? 

I think I would make an explicit `Comparator` for this and sort, then take firstOrNull. Otherwise your code will break if people change the ordering in the enum (although, your test will catch that, so will leave it up to you if you want to change this!)

molecular/src/test/kotlin/com/hartwig/actin/molecular/orange/MolecularRecordAnnotatorTest.kt
```diff
@@ -30,4 +61,54 @@ class MolecularRecordAnnotatorTest {
...
+    @Test
+    fun `Should not reannotate driver likelihood when variant has driverlikelihood null`() {
+        val variants = listOf(
+            TestVariantFactory.createMinimal().copy(driverLikelihood = DriverLikelihood.HIGH),
+            TestVariantFactory.createMinimal()
+        )
+        val output = annotator.reannotateDriverLikelihood(variants)
+        assertThat(output[0].driverLikelihood).isEqualTo(DriverLikelihood.HIGH)
+        assertThat(output[1].driverLikelihood).isNull()
+    }
```
kduyvesteyn: Two suggestions:
 - I would explicitly set the gene for every variant since your logic depends on having identical genes
 - I would add a test that driver likelihood is not updated for variants on different genes. In theory your code could now update the DL to HIGH for _all_ variants found, there is no test that safeguards for that!

molecular/src/test/kotlin/com/hartwig/actin/molecular/hotspot/HotspotFunctionsTest.kt
kduyvesteyn, line 1: You added this class in molecular package while the implementation exists in common module. Would suggest to move them into the same module!

molecular/src/test/kotlin/com/hartwig/actin/molecular/panel/PanelVariantAnnotatorTest.kt
```diff
@@ -135,23 +106,23 @@ private val PAVE_ANNOTATION = PaveResponse(
...
 )
 
-private val HOTSPOT = TestServeKnownFactory.hotspotBuilder().build()
-    .withGeneRole(ServeGeneRole.ONCO)
-    .withProteinEffect(ServeProteinEffect.GAIN_OF_FUNCTION)
-    .withAssociatedWithDrugResistance(true)
+private val HOTSPOT_CKB =
+    TestVariantAlterationFactory.createVariantAlteration(GENE, GeneRole.ONCO, ProteinEffect.GAIN_OF_FUNCTION, true, true)
+
+private val HOTSPOT_DOCM = TestVariantAlterationFactory.createVariantAlteration(GENE, isHotspot = true)
```
kduyvesteyn, line 112: If not mistaken, these objects are not CKB or DoCM-related? 

Would rename to something like `HOTSPOT_1` and `HOTSPOT_2` or perhaps more descriptive names that at least don't mention something that isn't true!

In general, panel variant annotator doesn't make assumptions about origin and type of alterations anymore so would reflect that in the test (you may be able to get away with a single hotspot, though didn't check in detail)

kduyvesteyn (approved):
Nice job @cbruel !

Logic and structure looks good to me, and report improvements look good.

One thing that may be good to think about is that the fact that your code makes some assumptions about primary vs secondary known events that ACTIN does not know about: e.g. that primary known events contain protein effects and are not always hotspots. 

Don't have concrete proposal here but it may help to isolate and even make explicit these assumptions (e.g. in `HotspotFunctions` you could hardcode "if known hotspot is from CKB then check protein effect, else assume it is always a hotspot"). It is not so nice to explicitly refer to a specific source but it is even more dodgy to make these assumptions anyways and not making it explicit. Another person reading your code would find it hard to understand without looking up the details in SERVE...

molecular/src/main/kotlin/com/hartwig/actin/molecular/orange/MolecularRecordAnnotator.kt
cbruel: I now noticed the `DriverLikelihoodComparator()` already exists :). So will use that, thanks!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasActivatingMutation.kt
```diff
@@ -108,25 +99,16 @@ class GeneHasActivatingMutation(
 
     private fun evaluateVariant(variant: Variant, hasHighMutationalLoad: Boolean?): ActivationProfile {
         val isNoOncogene = variant.geneRole == GeneRole.TSG
-        val isGainOfFunction =
-            variant.proteinEffect == ProteinEffect.GAIN_OF_FUNCTION || variant.proteinEffect == ProteinEffect.GAIN_OF_FUNCTION_PREDICTED
         return if (variant.isReportable) {
             if (variant.driverLikelihood == DriverLikelihood.HIGH) {
                 return when {
                     isAssociatedWithDrugResistance(variant) -> profile(variant.event, ActivationWarningType.ASSOCIATED_WITH_RESISTANCE)
-                    !variant.isHotspot && !isGainOfFunction -> profile(
```
kduyvesteyn, line 117: Why would you want to remove this warning? 

It is still possible for a variant to be HIGH driver and not be a hotspot or have gain function.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasFusionInGene.kt
```diff
@@ -35,11 +34,7 @@ class HasFusionInGene(override val gene: String, maxTestAge: LocalDate? = null)
                     val hasNoEffect =
                         (fusion.proteinEffect == ProteinEffect.NO_EFFECT || fusion.proteinEffect == ProteinEffect.NO_EFFECT_PREDICTED)
                     if (fusion.driverLikelihood != DriverLikelihood.HIGH) {
-                        if (isGainOfFunction) {
-                            fusionsWithNoHighDriverLikelihoodWithGainOfFunction.add(fusion.event)
-                        } else {
-                            fusionsWithNoHighDriverLikelihoodOther.add(fusion.event)
-                        }
+                        fusionsWithNoHighDriverLikelihoodOther.add(fusion.event)
```
kduyvesteyn, line 37: Cleaner to rename to just `fusionsWithNoHighDriverLikelihood`?

kduyvesteyn (approved):

PR #1105, opened 2025-05-02
ACTIN-1596: Treatment ranking algorithm based on evidence scoring

pauldwolfe:
Add the scoring algorithm proposed by the MCGI team which uses tumor match, variant match and approval status to create a score per evidence. This score is summed per treatment across molecular events.  

"Duplicate" evidences are summed with diminishing returns. A piece of evidence is considered duplicated when the event is in the same gene, matched on the same cancer type,  and has the same direction. 

The model is in the new interpretation module and is run in the report.  There is a really basic table in the report which I would love some ideas on how to better flesh out. 

See ticket for the details on the scoring.


interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
```diff
@@ -0,0 +1,65 @@
...
+        })
+    }
+
+    private fun geneTumorMatchAndBenefitVsResistance(s: Score) = Triple(s.gene, s.scoringMatch.tumorMatch, (s.score < 0))
+}
+
+fun saturatingDiminishingReturnsScore(
+    evidenceScores: List<Score>, slope: Double = 1.5, midpoint: Double = 1.0
+) = evidenceScores.sortedDescending().withIndex().map { (index, score) ->
+    score to if (index >= 1) score.score * (1.0 / (1.0 + exp(slope * (index - midpoint)))) else score.score
```
kduyvesteyn, line 64: Nifty :)

interpretation/src/test/kotlin/com/hartwig/actin/treatment/TreatmentRankingModelTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        assertThat(rank[0].score).isEqualTo(2950.0)
+    }
+
+
+    private fun patientRecord(
+        vararg variants: Variant
+    ) = TestPatientFactory.createProperTestPatientRecord().copy(
+        molecularHistory = MolecularHistory(
+            listOf(
+                TestMolecularFactory.createExhaustiveTestMolecularRecord().copy(
```
kduyvesteyn, line 204: I would use the _minimal_ test records here since you ideally want records with _only_ the drivers you pass in?

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularDetailsChapter.kt
```diff
@@ -20,6 +20,7 @@ import com.hartwig.actin.report.pdf.util.Cells
 import com.hartwig.actin.report.pdf.util.Formats
 import com.hartwig.actin.report.pdf.util.Tables
 import com.hartwig.actin.report.trial.ExternalTrialSummary
+import com.hartwig.actin.treatment.TreatmentEvidenceRanking
```
kduyvesteyn, line 23: Unnecessary import

system/src/test/kotlin/com/hartwig/actin/system/example/LocalExampleTreatmentMatchApplication.kt
```diff
@@ -62,7 +62,7 @@ class LocalExampleTreatmentMatchApplication {
 
     private fun createExampleRuleMappingResources(referenceDateProvider: ReferenceDateProvider): RuleMappingResources {
         val resourceDirectory =
-            listOf(System.getProperty("user.home"), "hmf", "repos", "actin-resources-private").joinToString(File.separator)
+            listOf(System.getProperty("user.home"), "Code", "actin-resources-private").joinToString(File.separator)
```
kduyvesteyn, line 65: Switch to the local conventions everyone in AMS uses!!!

kduyvesteyn (commented):
Generally looks good to me! 

This logic should eventually end up in algo and should have an explicit datamodel inside actin-datamodel (so the report just reads from the JSON produced by algo).

As a first step this can be an isolated step in algo (similar to personalization, resistance evidence matching for SOC, etc) but eventually we should:
 a) map treatments from SERVE to the common ACTIN treatment datamodel (rather than string, every TreatmentEvidence should refer to a `Treatment`). 
 b) combine this logic with the other inputs we have about treatment efficacy (such as personalization).

kduyvesteyn:
As for the report, you could consider showing a bit more to help with debugging but that also requires extending the exposed datamodel, and the data that was input to the scoring algorithm is already displayed separately in molecular evidence chapter. I don't have inspiration to further improve, but will give it a bit more thought - and would be useful to merge this in so that we can more easily test on EMC/NKI samples (by switching the environment config param to true and rerunning on a VM)

algo/src/test/kotlin/com/hartwig/actin/algo/TreatmentMatcherTest.kt
```diff
@@ -67,7 +67,7 @@ class TreatmentMatcherTest {
         referenceDateIsLive = true,
         trialMatches = trialMatches,
         standardOfCareMatches = null,
-        maxMolecularTestAge = MAX_AGE
+        maxMolecularTestAge = MAX_AGE,
```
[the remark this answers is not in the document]
pauldwolfe: whatever do you mean :)

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/evidence/TreatmentEvidence.kt
```diff
@@ -1,14 +1,22 @@
...
+    ALL_TYPES,
+    OTHER_TYPE
+}
+
 data class TreatmentEvidence(
     val treatment: String,
     val molecularMatch: MolecularMatchDetails,
     val applicableCancerType: CancerType,
-    val isOnLabel: Boolean,
+    val cancerTypeMatchApplicability: CancerTypeMatchApplicability,
```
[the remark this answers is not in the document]
pauldwolfe: I like having a match details object.

interpretation/src/main/kotlin/com/hartwig/actin/treatment/EvidenceScoringModel.kt
```diff
@@ -0,0 +1,47 @@
...
+) : Comparable<Score> {
+
+    override fun compareTo(other: Score): Int {
+        return score.compareTo(other.score)
+    }
+}
+
+class EvidenceScoringModel {
+
+    fun score(treatment: TreatmentEvidence, gene: String?): Score {
```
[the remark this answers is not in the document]
pauldwolfe: yeah see your point. In ranking model it's more than bookkeeping (use it in the duplicate evidence grouping), but you're right its actually not used in scoring.  I'll see what I can do.

pauldwolfe:
> Generally looks good to me!
> 
> This logic should eventually end up in algo and should have an explicit datamodel inside actin-datamodel (so the report just reads from the JSON produced by algo).
> 
> As a first step this can be an isolated step in algo (similar to personalization, resistance evidence matching for SOC, etc) but eventually we should: a) map treatments from SERVE to the common ACTIN treatment datamodel (rather than string, every TreatmentEvidence should refer to a `Treatment`). b) combine this logic with the other inputs we have about treatment efficacy (such as personalization).

I actually started with an isolated step in algo, but I changed that as it felt a bit tacked on and felt like it was duplicating data in the patient record as everything can be derived from the patient record.  I had the idea that eventually interpretation might output some JSON based output and we'd move more logic from report into interpretation. But I do see the parallels with personalization and resistance. Would we consider decoupling the evidence from the individual variants and instead having a ranked treatment list? 

No disagreement on the mapping treatments from SERVE to our treatment DB though.

interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
```diff
@@ -0,0 +1,85 @@
...
+                it.first.treatmentEvidence.treatment,
+                it.first.target,
+                it.second.scoringMatch.tumorMatch,
+                (it.second.score > 0)
+            )
+        }.mapValues { entry ->
+            saturatingDiminishingReturnsScore(entry.value.map { it.second })
+        }
+
+    private fun treatmentEvidencesWithTargets(actionables: Sequence<Actionable>) =
```
pauldwolfe, line 60: @kduyvesteyn wondering what you think of this. The idea is that we want to group together evidences that are close enough conceptually to be considered duplicates, and that is why gene is getting passed around. However the previous implementation would not have worked properly for characteristics or fusions (they would have been treated as dups).  Also using gene is a bit of a shortcut as we don't know if two variants are part the same CKB "tree"

Then again, this is just an internal logic for an algo we already know is far from perfect, so maybe fine to move forward for now and not spend too much time on this until we have more data on the performance of the algo.

kduyvesteyn: This looks quite ok to me, maybe I'd throw on the else part in this function but that's a detail.

I have started growing more fond of the framework of the evidence ranking last few weeks and when reviewing this PR, there is probably not much more we can do given the data we have (ideally you'd weigh evidence based on # patients tested in the paper but we don't have that information). 

I can't imagine the actual weights we currently have are optimal but think there is a chance that the framework may hold once we actually have some retrospective data + truth to test against.

kduyvesteyn: In any case, the recent changes in grouping for diminishing scores look good to me!

kduyvesteyn:
> > Generally looks good to me!
> > This logic should eventually end up in algo and should have an explicit datamodel inside actin-datamodel (so the report just reads from the JSON produced by algo).
> > As a first step this can be an isolated step in algo (similar to personalization, resistance evidence matching for SOC, etc) but eventually we should: a) map treatments from SERVE to the common ACTIN treatment datamodel (rather than string, every TreatmentEvidence should refer to a `Treatment`). b) combine this logic with the other inputs we have about treatment efficacy (such as personalization).
> 
> I actually started with an isolated step in algo, but I changed that as it felt a bit tacked on and felt like it was duplicating data in the patient record as everything can be derived from the patient record. I had the idea that eventually interpretation might output some JSON based output and we'd move more logic from report into interpretation. But I do see the parallels with personalization and resistance. Would we consider decoupling the evidence from the individual variants and instead having a ranked treatment list?
> 
> No disagreement on the mapping treatments from SERVE to our treatment DB though.

I imagine we'd add an instance of `TreatmentEvidenceRanking` to `TreatmentMatch` to start with, and eventually integrate this with `standardOfCareMatches` and `personalizedDataAnalysis`. `TreatmentMatch` should maybe be renamed to `TreatmentRecommendationAnalysis` or something, since we could also use the instance of `TreatmentEvidenceRanking` for annotating trial matches (in case we know the drug - OMICO already does this form of ranking already)

PR #1106, opened 2025-05-05
ACTIN-1928: Switch from Transvar to Reverse PAVE

zoehehartwig:
Swap Transvar for Reverse PAVE in ACTIN and SERVE

### DIFF
--- a/molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVariantAnnotator.kt
+++ b/molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVariantAnnotator.kt
@@ -1,6 +1,7 @@
 package com.hartwig.actin.molecular.panel
 
 import com.hartwig.actin.datamodel.clinical.SequencedVariant
+import com.hartwig.actin.datamodel.molecular.RefGenomeVersion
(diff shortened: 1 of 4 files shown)


molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVariantAnnotator.kt
```diff
@@ -85,12 +91,22 @@ class PanelVariantAnnotator(
     }
 
     private fun transvarAnnotation(sequencedVariant: SequencedVariant): TransvarVariant? {
+
+        val resourceUrl = javaClass.getClassLoader().getResource("ensembl_mini")
+        try {
+            checkNotNull(resourceUrl)
+            val ensemblDataDir = File(resourceUrl.toURI())
+            val reversePave = ReversePave(ensemblDataDir, RefGenomeVersion.V38, refGenome)
```
kzuberihmf, line 99: Name shadowing of ref genome version, try:

```java
            val reversePave = ReversePave(ensemblDataDir, com.hartwig.hmftools.common.genome.refgenome.RefGenomeVersion.V38, refGenome)
```

PR #1109, opened 2025-05-06
ACTIN-1911 Improvements for ACTIN-1911  based on PR Feedback

andreia-hartwig:
This PR contain the improvements done based on the feedback provided to PR https://github.com/hartwigmedical/actin/pull/1096.

A separate PR was created since the initial one is already closed.


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/SequencingTestExtractor.kt
```diff
@@ -24,11 +24,11 @@ class SequencingTestExtractor(
 ) {
 
     fun extract(patientId: String, questionnaire: Questionnaire?): ExtractionResult<List<SequencingTest>> {
-        if (questionnaire == null || questionnaire.ihcTestResults.isNullOrEmpty()) {
+        if (questionnaire?.ihcTestResults?.isEmpty() == true) {
```
jbartletthmf, line 27: `isNullOrEmpty()` was better here because it would also handle the null case so you wouldn't have to use null-safe calls in the rest of the function. If my suggestion complicated the null handling, I would go back to what you had before.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/MolecularSummaryGenerator.kt
```diff
@@ -46,7 +46,7 @@ class MolecularSummaryGenerator(
             nonIhcTestsIncludedInTrialMatching,
             ihcTestsFiltered,
             patientRecord.pathologyReports
-        ).filterValues { (_, molecularTest, ihcTests) -> molecularTest.isNotEmpty() || ihcTests.isNotEmpty() }
```
jbartletthmf, line 49: Do you still need the filtering here since it doesn't use ORANGE results?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -50,24 +61,30 @@ object PathologyReportFunctions {
...
             }
 
-        val unmatchedEntry = (null to Triple(unmatchedOrangeResults, unmatchedMolecularTests, unmatchedIHCTests))
-            .takeIf { unmatchedOrangeResults.isNotEmpty() || unmatchedMolecularTests.isNotEmpty() || unmatchedIHCTests.isNotEmpty() }
+        val unmatchedEntry = (null to Triple(
+            orangeResultsByDate[null] ?: emptyList(),
+            molecularTestsByDate[null] ?: emptyList(),
+            ihcTestsByDate[null] ?: emptyList()
+        ))
+            .takeIf { it.second.first.isNotEmpty() || it.second.second.isNotEmpty() || it.second.third.isNotEmpty() }
```
jbartletthmf, line 82:
```suggestion
        val unmatchedEntry = Triple(
            orangeResultsByDate[null].orEmpty(), molecularTestsByDate[null].orEmpty(), ihcTestsByDate[null].orEmpty()
        )
            .takeIf { it.toList().any { e -> e.isNotEmpty() } }
            ?.let { null to it }
```

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/MolecularSummaryGenerator.kt
andreia-hartwig: The filtering here was removed.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/SequencingTestExtractor.kt
andreia-hartwig: The previous version did make the null handling for the rest of the function a bit more tricky. I will revert.

jbartletthmf (approved):

PR #1111, opened 2025-05-07
ACTIN-1947: Extract externalDate + show lab on pathology report summary

ninajacobs:
Hi Andreia, would you be willing to review this PR? :)

Changes:
- Extract externalDate (I think this was an oversight)?
- Determine lab if it is null and source is internal -> I think my solution should work but we could of course also ask to actually fill the JSON directly.. Happy to hear what is more (strongly) preferred.
- Add the lab string to the report. I wondered, how did you easily test this yesterday?

### DIFF
--- a/report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/MolecularSummaryGenerator.kt
+++ b/report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/MolecularSummaryGenerator.kt
@@ -1,14 +1,14 @@
 package com.hartwig.actin.report.pdf.tables.molecular
 
-import com.hartwig.actin.algo.evaluation.molecular.IHCTestFilter
-import com.hartwig.actin.datamodel.clinical.IHCTest
+import com.hartwig.actin.algo.evaluation.molecular.IhcTestFilter
+import com.hartwig.actin.datamodel.clinical.IhcTest
 import com.hartwig.actin.datamodel.clinical.PathologyReport
 import com.hartwig.actin.datamodel.molecular.ExperimentType
 import com.hartwig.actin.datamodel.molecular.MolecularRecord
 import com.hartwig.actin.datamodel.molecular.MolecularTest
 import com.hartwig.actin.molecular.filter.MolecularTestFilter
 import com.hartwig.actin.report.datamodel.Report
-import com.hartwig.actin.report.interpretation.IHCTestInterpreter
+import com.hartwig.actin.report.interpretation.IhcTestInterpreter
 import com.hartwig.actin.report.interpretation.InterpretedCohort
 import com.hartwig.actin.report.pdf.tables.TableGenerator
 import com.hartwig.actin.report.pdf.util.Cells
@@ -40,7 +40,7 @@ class MolecularSummaryGenerator(
         val table = Tables.createSingleCol()
         val nonIhcTestsIncludedInTrialMatching =
             molecularTestFilter.apply(patientRecord.molecularHistory.molecularTests).filterNot { it.experimentType == ExperimentType.IHC }
-        val ihcTestsFiltered = IHCTestFilter.mostRecentOrUnknownDateIhcTests(patientRecord.ihcTests).toList()
+        val ihcTestsFiltered = IhcTestFilter.mostRecentOrUnknownDateIhcTests(patientRecord.ihcTests).toList()
         val groupedByPathologyReport = PathologyReportFunctions.groupTestsByPathologyReport(
             emptyList(),
             nonIhcTestsIncludedInTrialMatching,
@@ -76,7 +76,7 @@ class MolecularSummaryGenerator(
     private fun content(
         pathologyReport: PathologyReport?,
         molecularTests: List<MolecularTest>,
-        ihcTests: List<IHCTest>,
+        ihcTests: List<IhcTest>,
         table: Table
     ) {
         for (molecularTest in molecularTests.sortedByDescending { it.date }) {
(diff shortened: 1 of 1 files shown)


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/PathologyReportsExtractor.kt
```diff
@@ -2,21 +2,25 @@ package com.hartwig.actin.clinical.feed.standard.extraction
...
+
     override fun extract(ehrPatientRecord: ProvidedPatientRecord): ExtractionResult<List<PathologyReport>> {
         return ExtractionResult(
             ehrPatientRecord.tumorDetails.pathology.takeIf { !it.isNullOrEmpty() }?.map {
                 PathologyReport(
                     tissueId = it.tissueId,
                     reportRequested = it.reportRequested,
                     source = it.source,
-                    lab = it.lab,
+                    lab = if (it.lab == null && it.source.lowercase() == "internal") requestingHospital else it.lab,
```
andreia-hartwig, line 19: @ninajacobs Since showing the requesting hospital is a presentational aspect, I would not add it to the data model. I would simply use it when generating the report. The requestingHospital is already available in the Report object, so you provide the report as parameter to PathologyReportFunctions, you already have all the data you need.

andreia-hartwig (changes requested):

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/PathologyReportsExtractor.kt
ninajacobs: Thanks, and understood. changed it!

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/PathologyReport.kt
```diff
@@ -8,7 +8,7 @@ data class PathologyReport(
     val tissueId: String? = null,
     val reportRequested: Boolean,
     val source: String,
-    val lab: String? = null,
+    val lab: String?,
```
andreia-hartwig, line 11: I think here you could keep the default null value, follows the same pattern as used for the dates.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/MolecularSummaryGenerator.kt
```diff
@@ -51,7 +52,14 @@ class MolecularSummaryGenerator(
         for ((pathologyReport, tests) in groupedByPathologyReport) {
             val (_, molecularTests, ihcTests) = tests
             pathologyReport?.let {
-                table.addCell(Cells.create(PathologyReportFunctions.getPathologyReportSummary(report = pathologyReport)))
+                table.addCell(
+                    Cells.create(
+                        PathologyReportFunctions.getPathologyReportSummary(
+                            report = pathologyReport,
+                            config = EnvironmentConfiguration()
```
andreia-hartwig, line 59: Instead of adding the EnvironmentConfiguration here, you could change the MolecularSummaryGenerator to have the report as parameter and pass the report to PathologyReportFunctions.getPathologyReportSummary. 
You can get the patientRecord from report.patientRecord and the hospital from report.requestingHospital.

ninajacobs: Thank you!

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -20,8 +20,16 @@ object PathologyReportFunctions {
...
-    fun getPathologyReportSummary(prefix: String? = null, prefixStyle: Style? = null, report: PathologyReport): Cell =
-        Cells.create(
+    fun getPathologyReportSummary(
+        prefix: String? = null,
+        prefixStyle: Style? = null,
+        pathologyReport: PathologyReport,
+        requestingHospital: String?
+    ): Cell {
+        val labString =
+            if (pathologyReport.lab == null && pathologyReport.isSourceInternal && requestingHospital != null) requestingHospital else pathologyReport.lab
```
andreia-hartwig, line 30: pathologyReport.lab is only used for external sources so if the source is internal then pathologyReport.lab will be null. 

So this could simply be : 
val labString  = if (pathologyReport.isSourceInternal) requestingHospital else pathologyReport.lab

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -30,28 +38,32 @@ object PathologyReportFunctions {
                             Text(" - ").addStyle(Styles.tableHighlightStyle())
                         )
                     },
-                    listOf(Text(report.tissueId?.uppercase() ?: "Unknown Tissue ID").addStyle(Styles.tableTitleStyle())),
-                    if (report.isSourceInternal) {
+                    listOf(
+                        Text(pathologyReport.tissueId?.uppercase() ?: "Unknown Tissue ID").addStyle(Styles.tableTitleStyle()),
+                        Text(" ($labString").addStyle(Styles.tableHighlightStyle())
```
andreia-hartwig, line 43: The requestingHospital is always available for NKI, and we only get pathology reports from NKI so this works.
However, the requestingHospital can be null (EMC) and so the labString can in theory be null, in which case the final text would show "<tissue_id> (, ...". 
Perhaps you can add a null check?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
ninajacobs: I agree, but this would add a dependency on what I know about how the data should be right? I thought it's cleaner not to assume anything? Or if we will need to be able to assume this, we'd need to check for it upon ingestion probably?

andreia-hartwig (approved):

PR #1112, opened 2025-05-07
ACTIN-1591: Specify whether a trial is 'ignored' or 'non-evaluable' in the ACTIN report

dkhartwig:
In the report section titled **_trials and cohorts that are not evaluable or ignored_** clarified whether each cohort is classified as  _ignored_,  _non-evaluable_ or  _ignored and non-evaluable_ to enhance reader comprehension. 

Following discussion with Nina, retained a single table for both cohort types and introduced an additional column to explicitly indicate the configuration of each cohort.


ninajacobs (commented):
Looks good conceptually! I checked the example report and I would change the content of "Configuration" to start with a capital letter, such that it is equal with content in other columns (Non-evaluable vs non-evaluable)

Will leave technical review and approval to Andreia!

andreia-hartwig (approved):

PR #1113, opened 2025-05-08
ACTIN-1948: Three small changes to TMB/TML in molecular summary

ninajacobs:
3 small changes:
- Show unit for TMB (only in summary, it didn't look good in details)
- Show only TML and TMB value in summary - i.e. drop "High" and "Low". Of note, I kept "High" and "Low" in details because we do use High and Low for evidence matching.
- If TMB is known and TML is unknown, don't show "TML unknown" in summary (then only show TMB)


report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularCharacteristicFormatTest.kt
```diff
@@ -17,30 +17,76 @@ class MolecularCharacteristicFormatTest {
...
+                withTumorMutationalBurden(score = 61.0, isHigh = true),
+                displayValue = true
+            )
+        ).isEqualTo("TMB 61 mut/Mb")
 
         assertThat(
-            MolecularCharacteristicFormat.formatTumorMutationalBurden(withTumorMutationalBurden(score = 6.0, isHigh = false))
-        ).isEqualTo("TMB Low (6)")
+            MolecularCharacteristicFormat.formatTumorMutationalBurden(
+                withTumorMutationalBurden(score = 6.0, isHigh = true),
```
kduyvesteyn, line 28: My brain struggles with inconsistent test data, would suggest to use 61.0 here as score.

Also, may be nice to have an example with digits, after reading test its not clear if rounding always happens.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/WGSSummaryGeneratorFunctions.kt
```diff
@@ -173,10 +172,10 @@ object WGSSummaryGeneratorFunctions {
...
     }
 
-    private fun tumorMutationalLoadAndTumorMutationalBurdenStatus(molecular: MolecularTest): String {
-        val tmlString = MolecularCharacteristicFormat.formatTumorMutationalLoad(molecular.characteristics)
-        val tmbString = MolecularCharacteristicFormat.formatTumorMutationalBurden(molecular.characteristics)
-        return String.format("%s / %s", tmlString, tmbString)
+    private fun tumorMutationalLoadAndTumorMutationalBurdenStatus(molecular: MolecularTest, returnOnlyTmb: Boolean = false): String {
+        val tmlString = MolecularCharacteristicFormat.formatTumorMutationalLoad(molecular.characteristics, true)
+        val tmbString = MolecularCharacteristicFormat.formatTumorMutationalBurden(molecular.characteristics, true)
+        return if (!returnOnlyTmb) String.format("%s / %s", tmlString, tmbString) else tmbString
```
kduyvesteyn, line 178: This feels a bit itchy, there is a function to return TML and TMB which can be configured to just return TMB.

I would rather make this a separate function instead.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/WGSSummaryGeneratorFunctions.kt
```diff
@@ -94,15 +89,19 @@ object WGSSummaryGeneratorFunctions {
         isShort: Boolean,
         table: Table
     ): Boolean {
-        val tmbStatus = tumorMutationalLoadAndTumorMutationalBurdenStatus(molecular)
-        if (!isShort || tmbStatus != "TML ${Formats.VALUE_UNKNOWN} / TMB ${Formats.VALUE_UNKNOWN}") {
+        val tmbStatus = tumorMutationalLoadAndTumorMutationalBurdenStatus(molecular, returnOnlyTmb = true)
+        val tmlAndTmbStatus = tumorMutationalLoadAndTumorMutationalBurdenStatus(molecular)
```
kduyvesteyn, line 93: Both fields are only used inside an if block, its cleaner to create them there.

kduyvesteyn (approved):
Have only reviewed technically, a few minor comments but other than that looks good!

report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularCharacteristicFormatTest.kt
ninajacobs: check!

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/WGSSummaryGeneratorFunctions.kt
ninajacobs: Agreed, I kept the original function and defined the val now directly (another function felt unneeded)

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/WGSSummaryGeneratorFunctions.kt
ninajacobs: thanks!

kduyvesteyn (approved):
Looks good!

PR #1116, opened 2025-05-09
ACTIN-1957: Extend treatment datamodel for oncoactin

ArnoHMF:
Added fields required for treatment table in Oncoact report.

Small refactoring in ClinicalEvidenceFactory to remove some duplication.


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/evidence/TreatmentEvidence.kt
```diff
@@ -9,7 +9,9 @@ data class TreatmentEvidence(
     val evidenceLevelDetails: EvidenceLevelDetails,
     val evidenceDirection: EvidenceDirection,
     val evidenceYear: Int,
-    val efficacyDescription: String
+    val efficacyDescription: String,
+    val treatmentApproachesDrugClass: Set<String>,
+    val treatmentApproachesTherapy: Set<String>
```
kduyvesteyn, line 14: @LiekeSchoenmaker Could you comment on what these fields are precisely? In the latest SERVE database (`serve_latest` schema), every `treatmentApproachesTherapy` is equal to `treatment`, and the `treatmentApproachesDrugClass` could in my opinion be renamed to `drugClasses` and be just below treatment in the datamodel?

@ArnoHMF Other than that all looks good to me, but I think we only have to add one field `drugClasses` and clean up usage of these fields in oncoact (and rename them accordingly in SERVE)

ArnoHMF: In OncoAct these lists are directly combined, so I see no problem doing that already in ACTIN, so there's only one field. Question is what would be a good name for this field then. 

In the report these are called drug type. So we could go with that? But open to better name suggestions.

kduyvesteyn: `drugTypes` would be a good name I think, the data (based on `serve_latest` contains the CKB-equivalent of the `DrugType` enum in ACTIN. 

Note that eventually we should map the SERVE treatment to ACTIN treatment (val `treatment: Treatment` rather than than `val treatment: String`) and then this field can be dropped again but that's a bigger change.

kduyvesteyn: Actually... would make it `treatmentTypes` since SERVE doesn't decompose a treatment into its individual drugs.

ArnoHMF: Changed it to treatmentTypes

kduyvesteyn (approved):

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/evidence/TreatmentEvidence.kt
LiekeSchoenmaker: @kduyvesteyn / @ArnoHMF ; 
CKB listed two types of treatment approaches based on therapy and drug class. 
A drugclass is linked to a particular evidence if there are multiple entries which linked to the specific drug class. The therapy is added currently when a drug class is not available but ignoring those entries results in missing important evidence. Also not every drug name is assigned to a treatmentapproachTherapy. 
Currently the following logics is used in OncoAct: 
- use treatmentApproach drugclass when available
-  use treatmentApproach tharapy when the drug class is not available 
- when both are absent the evidence can be ignored for reporting 

```java

 @NotNull
    private static TreeSet<String> extractCombinedTreatmentApproaches(com.hartwig.serve.datamodel.Treatment treatment) {
        Set<String> treatmentApproachesDrugClass = treatment.treatmentApproachesDrugClass();
        Set<String> treatmentApproachesTherapy = treatment.treatmentApproachesTherapy();
        return Sets.newTreeSet(treatmentApproachesDrugClass.isEmpty() ? treatmentApproachesTherapy : treatmentApproachesDrugClass);
    }

LiekeSchoenmaker: I have just the naming of CKB but fine to go with the naming of DrugClass in the downstream algorithms

PR #1117, opened 2025-05-09
ACTIN-1811: p1 process evidences for panels in a later step

kzuberihmf:
Combined evidence annotation requires the entire molecular test to be ready, but currently evidence annotations are added as each event is processed. This moves evidence annotations for panels to a separate step in preparation for that change.

No functional changes, only refactoring.


molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelEvidenceAnnotator.kt
```diff
@@ -0,0 +1,102 @@
...
+import com.hartwig.actin.molecular.MolecularAnnotator
+import com.hartwig.actin.molecular.driverlikelihood.GeneDriverLikelihoodModel
+import com.hartwig.actin.molecular.evidence.EvidenceDatabase
+import com.hartwig.actin.molecular.evidence.matching.MatchingCriteriaFunctions
+import com.hartwig.actin.molecular.interpretation.GeneAlterationFactory
+
+class PanelEvidenceAnnotator(
+    private val evidenceDatabase: EvidenceDatabase,
+    private val geneDriverLikelihoodModel: GeneDriverLikelihoodModel,
+) : MolecularAnnotator<PanelRecord, PanelRecord> {
```
pauldwolfe, line 18: Just a thought, he use of the MolecularAnnotator interface here is a bit artificial as it is not used directly in a MolecularInterpreter.  You could have this class just take Drivers as input and produce Drivers as output and have the PanelAnnotator copy it in.

pauldwolfe (approved):

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/TestMolecularFactory.kt
```diff
@@ -293,6 +306,35 @@ object TestMolecularFactory {
...
+            effects = setOf(),
+            codingEffect = CodingEffect.NONE,
+            affectedExon = null
+        ),
+        otherImpacts = emptySet(),
+        extendedVariantDetails = null,
+        isHotspot = false,
+        isReportable = false,
+        event = "",
+        driverLikelihood = DriverLikelihood.LOW,
```
kduyvesteyn, line 330: driverLikelihood should be null in the minimal version of a variant

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/TestMolecularFactory.kt
```diff
@@ -324,6 +366,22 @@ object TestMolecularFactory {
         isAssociatedWithDrugResistance = true,
     )
 
+    fun createMinimalFusion() = Fusion(
```
kduyvesteyn, line 369: Same comment on ordering here. Also, `fusedExonUp` and `fusedExonDown` could be null in a minimal version

kduyvesteyn (commented):
Good catch @kzuberihmf but I suspect you haven't taken it far enough yet.

For the combined evidence to work properly, the entire record has to be fully annotated with known effects, driver likelihood etc.

Currently there are some issues with the chain of events:
 - Evidence for characteristics is still defined within PanelAnnotator.interpret (so doesn't have the annotated variants yet). 
 - When the variants are annotated in `PanelEvidenceAnnotator` the copy numbers aren't annotated with geneAlteration yet. 

I think the only correct approach is to have a fully annotated (with gene alterations, known events, driver likelihood) molecular test and only then perform the evidence matching.

kzuberihmf:
> Currently there are some issues with the chain of events:

Oh, good point. I've tried splitting it further, though not into completely separate classes.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelEvidenceAnnotator.kt
kzuberihmf: > Just a thought, he use of the MolecularAnnotator interface here is a bit artificial as it is not used directly in a MolecularInterpreter. You could have this class just take Drivers as input and produce Drivers as output and have the PanelAnnotator copy it in.

Nice observation, yeah it seems like it could be narrowed to drivers. But I think in a future update for combined evidence, the annotation will need characteristics to be available as well, so it might make sense to keep the entire test record as a unit. Will leave for the moment.

kzuberihmf:
> Good catch @kzuberihmf but I suspect you haven't taken it far enough yet.
> 
> For the combined evidence to work properly, the entire record has to be fully annotated with known effects, driver likelihood etc.
> 
> Currently there are some issues with the chain of events:
> 
> * Evidence for characteristics is still defined within PanelAnnotator.interpret (so doesn't have the annotated variants yet).
> * When the variants are annotated in `PanelEvidenceAnnotator` the copy numbers aren't annotated with geneAlteration yet.
> 
> I think the only correct approach is to have a fully annotated (with gene alterations, known events, driver likelihood) molecular test and only then perform the evidence matching.

I ended up removing EvidenceDatabase from being available directly in PanelAnnotator, now only PanelEvidenceAnnotator has it. For this evidence annotator, it now proceeds in two phases, the first is for attributes requiring the evidence database but not yet annotating clinical evidence (other attributes for drivers), then a second phase that adds only the clinical evidence.

Does that seem satisfactory @kduyvesteyn?

Having the same type in different states of annotation completeness seems a bit worrying, but I think we are going with that for now.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelEvidenceAnnotator.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    override fun annotate(input: PanelRecord): PanelRecord {
+        return input
+            .let(::annotateWithDriverAttributes)
+            .let(::annotateWithClinicalEvidence)
+    }
+
+    fun annotateWithDriverAttributes(input: PanelRecord): PanelRecord {
```
pauldwolfe, line 25: these next two functions can be private

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelAnnotator.kt
```diff
@@ -11,21 +11,25 @@ import com.hartwig.actin.datamodel.molecular.characteristics.TumorMutationalBurd
...
     private val panelCopyNumberAnnotator: PanelCopyNumberAnnotator,
+    private val panelEvidenceAnnotator: PanelEvidenceAnnotator,
     private val panelSpecifications: PanelSpecifications
 ) : MolecularAnnotator<SequencingTest, PanelRecord> {
 
     override fun annotate(input: SequencingTest): PanelRecord {
+        return panelEvidenceAnnotator.annotate(interpret(input))
+    }
+
+    fun interpret(input: SequencingTest): PanelRecord {
```
pauldwolfe, line 32: this function can be private

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelEvidenceAnnotator.kt
```diff
@@ -0,0 +1,126 @@
...
+import com.hartwig.actin.molecular.interpretation.GeneAlterationFactory
+
+class PanelEvidenceAnnotator(
+    private val evidenceDatabase: EvidenceDatabase,
+    private val geneDriverLikelihoodModel: GeneDriverLikelihoodModel,
+) : MolecularAnnotator<PanelRecord, PanelRecord> {
+
+    override fun annotate(input: PanelRecord): PanelRecord {
+        return input
+            .let(::annotateWithDriverAttributes)
```
pauldwolfe, line 21: I like how you've centralized all annotations requiring the evidence database, but this code makes me think we could do one more split, as conceptually I guess driver attributes are not evidence, and as such it might lead to some confusion trying to find this code, or worse getting the order mixed up.

I guess this is causes by evidence database being more than just an evidence database, but I think refactoring that is not necessary at this point. 

What do you think about having a PanelDriverAttributeAnnotator and a PanelEvidenceAnnotator?

PR #1121, opened 2025-05-12
ACTIN-1954: Use common feed model in standard clinical ingestion

jbartletthmf:
There may still be some additional cleanup (e.g. renaming all "Provided" to "Feed"), but I wanted to do this quickly to minimize merge conflicts. I suspect we'll be making a lot of changes to this code as the feed transformation reduces the need for cross-curation.


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
```diff
@@ -6,74 +6,46 @@ import com.hartwig.actin.clinical.curation.CurationResponse
...
-                        diagnosedYear = it.diagnosisDate?.year,
-                        lastTreatmentYear = it.lastTreatmentDate?.year,
-                        lastTreatmentMonth = it.lastTreatmentDate?.monthValue,
-                        status = it.status?.let { status -> TumorStatus.valueOf(status.uppercase()) } ?: TumorStatus.UNKNOWN
-                    )
-                }
-            }), curatedPriorPrimary.extractionEvaluation)
+    private fun fromPriorPrimaries(ehrPatientRecord: FeedPatientRecord): List<ExtractionResult<List<PriorPrimary>>> =
+        ehrPatientRecord.priorPrimaries.map { feedPriorPrimary ->
+            val curatedPriorPrimary = curate(ehrPatientRecord.patientDetails.patientId, feedPriorPrimary.name)
```
andreia-hartwig, line 37: Before we were using input "${it.tumorLocation} | ${it.tumorType}" and now we just use the feedPriorPrimary name. Is this name build with the same format?

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/FeedTestData.kt
```diff
@@ -0,0 +1,57 @@
...
+        tumorLocation = "tumorLocation",
+        tumorType = "tumorType",
+        lesions = emptyList(),
+        hasBrainLesions = false,
+        hasActiveBrainLesions = false,
+        hasBoneLesions = false,
+        hasLiverLesions = false,
+        measurableDiseaseDate = DATE_Y2023_M02_D23,
+        measurableDisease = true,
+        tumorGradeDifferentiation = "tumorGradeDifferentiation",
```
andreia-hartwig, line 53: Is this still used? I thought this used to have the same content as the raw pathology report.

jbartletthmf: It's currently used in `StandardTumorDetailsExtractor` but I don't know if it *should* be used. Do we want the LLM to extract the tumor details from the raw report? We can try this, but may have to set up some guardrails since stage=`IV` resolves correctly but stage=`Stadium IV` does not. If we can't reliably extract the stage with the LLM, we could add a curation sheet.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
andreia-hartwig: Once this change is merged the curation need to be updated then?

clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/FeedTestData.kt
jbartletthmf: `actin-feed` already drops the IHC tests provided by Kaiko and extracts them from the pathology report using the LLM. I didn't think we should be using `tumorGradeDifferentiation` or `rawPathologyReport` on the actin side anymore.

andreia-hartwig: I agree, that is the code that I intend to remove once this PR is merged.

PR #1122, opened 2025-05-13
ACTIN-1971: Only consider clonal variants for inactivation

ninajacobs:
Change: only variants to be considered clonal (or clonal status unknown) should be considered for INACTIVATION of a gene.

This PR also removes the bug that subclonal variants were practically never leading to warnings (due to else-if structure with biallelic check) and there was a concat of inactivationSubclonalVariants for inactivationEventsOnNonCanonicalTranscript.


algo/README.md
```diff
@@ -313,7 +313,7 @@ Trials could lead to `WARN` in case knowing the exact trial treatment is require
...
-| INACTIVATION_OF_GENE_X                                                                | If geneRole of gene X is not 'ONCO', and (min copy nr < 0.5 or homozygous disruption or biallelic mutation(s) with high driver likelihood, and corresponding event does not match with protein effect gain of function (predicted)                                                                                 | `WARN` in case of gene role ONCO, driver likelihood, other protein effect, biallelic status and/or reportability                                                                        |
+| INACTIVATION_OF_GENE_X                                                                | If geneRole of gene X is not 'ONCO', and (min copy nr < 0.5 or homozygous disruption or biallelic mutation(s) with high driver likelihood, and corresponding event does not match with protein effect gain of function (predicted)                                                                                 | `WARN` in case of gene role ONCO, driver likelihood, other protein effect, biallelic status and/or reportability. Variants estimated to be not clonal are not matched.                  |
```
kduyvesteyn, line 316: Maybe change the description to "variants estimated to be subclonal are not matched"? Double-not looks odd to me.

cbruel, line 316: Feels like a lot of "not" "not" :) -> maybe "Only variants that are estimated to be clonal are matched"?

cbruel (commented):
Thanks Nina! Will leave approval to Korneel (don't feel confident I can review this extensively enough)

algo/README.md
kduyvesteyn: Haha nice one :)

PR #1123, opened 2025-05-13
ACTIN-1972: Some improvements to IHC test display

ninajacobs:
Changes to display:
- Display IHC tests with scoreValue & scoreText adequately
- Sort scoreText only first, then scoreText & scoreValue and scoreValue only, and consider grouping in sorting
- Select the item with most recent date for display

Other changes:
- Make item not-null
- drop unused IHC comparator code in common code

Before:
Screenshot 2025-05-13 at 11 54 12

After:
Screenshot 2025-05-13 at 11 54 02


report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/IhcResultGenerator.kt
```diff
@@ -39,7 +38,7 @@ class IhcResultGenerator(
     }
 
     private fun ihcTestInterpretationContents(ihcTestInterpretation: IhcTestInterpretation, table: Table) {
-        ihcTestInterpretation.results.sortedBy { it.sortPrecedence }
+        ihcTestInterpretation.results.sortedBy { it.grouping }.sortedBy { it.sortPrecedence }
```
andreia-hartwig, line 41: Doing two consecutive sortedBy does not sort with both elements, the end result is only sorted by sortPrecedence.
If you want to sort with both fields, grouping and sortPrecedence, you can do: 

`results.sortedWith(compareBy({ it.grouping }, { it.sortPrecedence }))`

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
```diff
@@ -13,23 +13,38 @@ class IhcTestInterpreter {
     private val interpretationBuilder = IhcTestInterpretationBuilder()
 
     fun interpret(ihcTests: List<IhcTest>): List<IhcTestInterpretation> {
-        ihcTests.forEach(::interpret)
+        val latestIhcTestsByItem = ihcTests.groupBy { it.item }.mapValues { (_, testsForItem) ->
```
andreia-hartwig, line 16: If you use map instead of mapValues here, the .values at the end is not needed. 
You could also do the filter on ihcTests, to avoid doing multiple times when there are multiple items.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/IhcResultGenerator.kt
ninajacobs: Your proposal doesn't seem to do what I intended, it now sorts on grouping only. But I wanted to sort on grouping (to make it alphabetically within the precedence), and then precedence (to actually use precedence too).

With your change:
Screenshot 2025-05-13 at 12 56 41

Original (as intended):
Screenshot 2025-05-13 at 12 57 23

andreia-hartwig: To make the same order I think you would have to do : 
`results.sortedWith(compareBy({ it.sortPrecedence }, { it.grouping }))`

but if your solution works as you expect it then I would say keep it.

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
andreia-hartwig: I see now that you want to return the 1st elements when there are none with a measure date. 
Then you can simply do: 
` val testsForItemWithMeasureDate = tests.maxByOrNull { it.measureDate ?: LocalDate.MIN }`

ninajacobs: Like this then? 

val latestIhcTestsByItem = ihcTests.groupBy { it.item }.map { (_, tests) ->
            tests.maxByOrNull { it.measureDate ?: LocalDate.MIN } ?: tests.first()
        }

andreia-hartwig: You don't actually need "?: tests.first()" at the end :)

andreia-hartwig (approved):

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ninajacobs: < resolved with Andreia to keep it like it is :) >

PR #1125, opened 2025-05-13
ACTIN-1811: p3 remove match criteria

kzuberihmf:
Part 3 of the series of updates for introducing combined evidence. While prototyping that implementation, the variant match and fusion match criteria were getting in the way. As far as I can tell they aren't really need (though might have been earlier when panel data was first introduced alongside orange). This just removes them and uses fusion and variant directly. No functional changes.

@cbruel one place that differs a bit more is hotspotcomparator, think that's only used in the hotspot comparator application, could you have a quick look to make sure converting to variant as I did for matching looks ok?


datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/TestMolecularFactory.kt
```diff
@@ -293,6 +293,38 @@ object TestMolecularFactory {
...
+        extendedVariantDetails = null,
+        isHotspot = false,
+        isReportable = false,
+        event = "",
+        driverLikelihood = null,
+        evidence = TestClinicalEvidenceFactory.createEmpty(),
+        gene = "",
+        geneRole = GeneRole.UNKNOWN,
+        proteinEffect = ProteinEffect.UNKNOWN,
+        isAssociatedWithDrugResistance = false
```
kduyvesteyn, line 314: This field can be `null` in the minimal version

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/TestMolecularFactory.kt
```diff
@@ -293,6 +293,38 @@ object TestMolecularFactory {
...
+    )
+
+    fun createMinimalTranscriptImpact() = TranscriptVariantImpact(
```
kduyvesteyn, line 317: Would stick to ordering as in datamodel. Also, `isSpliceRegion` can be null in minimal version

molecular/src/test/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotatorTest.kt
```diff
@@ -33,33 +33,49 @@ private val SEQUENCED_FUSION = SequencedFusion(GENE_START, GENE_END)
 private val FULLY_SPECIFIED_SEQUENCED_FUSION =
     SequencedFusion(GENE_START, GENE_END, TRANSCRIPT_START, TRANSCRIPT_END, FUSED_EXON_UP, FUSED_EXON_DOWN)
 
-private val FUSION_MATCH_CRITERIA = FusionMatchCriteria(
+private val FUSION_MATCH_CRITERIA = TestMolecularFactory.createMinimalFusion().copy(
```
kduyvesteyn, line 36: Rename variables in this class?

kzuberihmf:
Thanks for jumping in with the review!

PR #1126, opened 2025-05-14
ACTIN-1968: Keep non-ignored tests with empty results

jbartletthmf:
Exclude tests with only IHC results.
Show sequencing result warnings simultaneously with sequencing test warnings.


clinical/src/test/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardSequencingTestExtractorTest.kt
```diff
@@ -87,78 +79,34 @@ class StandardSequencingTestExtractorTest {
     }
 
     @Test
-    fun `Should extract sequencing with variants`() {
```
andreia-hartwig, line 90: Just wondering why these tests were removed. This change doesn't seem to impact the actual results.

jbartletthmf: We'll never get test results with variants, fusions, etc. in the current setup because the LLM extracts results from the pathology report and submits them as free text. These tests are an artifact of ingestion from Kaiko's PALGA LLM.

andreia-hartwig (approved):

PR #1128, opened 2025-05-15
ACTIN-1956: Add fallback option for tests where the specifications are not known

pauldwolfe:
Whether the specification is known is governed by the feed transformation.  When the feed marks a test
as having unknown specs, the testsGene function of the test will return true whenever the gene is
present in a mutation, fusion, amplification, deletion or skipped exon.

All other curation of tests remains the same, so you can still curate the name of the test, and any
test results.


pom.xml
```diff
@@ -29,7 +29,7 @@
         <orange-datamodel.version>3.0.0</orange-datamodel.version>
         <actin-transvar.version>1.0.0</actin-transvar.version>
         <actin-pave-lite.version>1.0.0</actin-pave-lite.version>
-        <actin-feed.version>0.7.0</actin-feed.version>
+        <actin-feed.version>local-SNAPSHOT</actin-feed.version>
```
pauldwolfe, line 32: will change when feed is released

jbartletthmf (approved):

ninajacobs (commented):
Nice Paul, logic sounds good!

I can't readily see if you make a distinction between mutation/amp/del/fusion being tested, but I would propose this:

gene has variant -> tested for mutation
gene has amp -> tested for amp + mutation
gene has del -> tested for del + mutation
gene has fusion -> tested for fusion
gene has no mutations found -> tested for mutation

I think these are fair assumptions based on observed panels.

What do you think?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/SequencingTest.kt
```diff
@@ -1,6 +1,7 @@
 package com.hartwig.actin.datamodel.clinical
 
 import com.hartwig.actin.datamodel.Displayable
+import com.hartwig.actin.datamodel.molecular.PanelSpecifications
```
andreia-hartwig, line 4: unused import

datamodel/src/test/kotlin/com/hartwig/actin/datamodel/molecular/DerivedPanelSpecificationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.hartwig.actin.datamodel.clinical.SequencingTest
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+private const val GENE = "gene"
+private const val TEST = "test"
+
+class DerivedPanelSpecificationTest {
+
+    @Test
```
andreia-hartwig, line 13: These first 5 tests do exactly the same and they all test variants only.

pauldwolfe: haha, nice catch, missed a commit.

ninajacobs (commented):
Looks good!

PR #1129, opened 2025-05-16
ACTIN-1811: p2 combined evidence matcher

kzuberihmf:


kzuberihmf:
@kduyvesteyn if you have time, could use a check for if this seems roughly sane.

The idea is to introduce a class `CombinedEvidenceMatcher` that takes a  full molecular test (either ORANGE or Panel), and creates a Map<Actionable, Set<EfficacyEvidence>>. So each Actionable is associated with all the evidences that are completely satisfied by that molecular test and has that Actionable as one of its criteria. Includes both single and combined criteria.  A later step (to be introduced) would then create a fully evidence annotated test record using the map.  

This ends up spreading out some of the logic that was previously nicely contained in the various XyzEvidence classes (FusionEvidence, DisruptionEvidence, etc) which I don't like, its easy to introduce bugs. I've pushed some of it back down into those classes by adding some helpers there to be used by the CombinedEvidenceMatcher, but only partially so far. 

Thoughts or recommendations appreciated!

kzuberihmf:
> I think there is no way around a different approach to evidence matching. We should move to "per efficacy evidence" rather than "per variant", and treat evidence for which 1 biomarker has to be present the same as combined (N=1 is just a specific example of combined evidence).
> 

Ok, this is implementation is tracking N=1, so I think we are in sync on this.

> Your `CombinedEvidenceMatcher` has the right structure I think, and should simply become the "evidence matcher" (also for non-combined evidence). Minor detail - if `matchEvidence` returns `Success`, are you sure that every condition has been met? Might just be me understanding the code poorly :)
> 

The idea is matchEvidence() returns Success if all the individual criterium sub-component matchers (like matchHotspots, matchCodons ...) return success. 

The sub-component matchers return Success when that sub-component is completely satisfied (all hotspots match), or don't need any matches because they are empty.

When aggregating these parts with combine(), any sub-component failure will cause the result to be Failure. The application of matching is lazy & fail-fast so early failures should short-circuit having to test all the parts, which I assume to be the common case.

Bugs of course possible, can see about improving tests, but that's the idea.

> For completeness-sake, I think the cleanest chain of steps is:
> 
> * Iterate over all evidences
> * Evaluate all criteria per evidence
> * Keep the evidence that matched every single criterium
> * Assign the matching evidence to every actionable which was part of the exhaustive match (so we duplicate the evidence in the case of combined evidence matches).
> 

Thanks for spelling this out, it always help to be explicit 😄. Here's what is happening now:

* Iterate over all evidences
* Evaluate all criteria per evidence against the molecular test
* When evaluating, keep the actionable from the molecular test that were completely satisfied (in Success)
* Aggregate the successes in the EvidencesForActionable Map which connects actionable -> Set<EfficacyEvidence> only for the fully satisfied cases.

So if we have evidence E with criteria requiring hotspot H and fusion B, that are both matched in the molecular test with variant h and fusion b, we will have have Map(h->Set(E), b ->Set(E)),

I think this is matching your proposal.

> So I think your `CombinedEvidenceMatcher` goes in the right direction but not even far enough (VariantEvidence should maybe be gone completely.
> 

By this I think you mean all the logic in VariantEvidence that I'm depending on should be pulled into the CombinedEvidenceMatcher? Trial matching currently still happens in there, and there are some shared constants, so I was attempting to keep using those but can move the evidence part of the logic out.

> Side note: Matching external trials should be mappable to this new structure as well but I suspect that will work fine out of the box since their molecular matching is an instance of combined matching if not mistaken (the OR / partially required may require slightly different approach though but imagine that's a detail)

Hopefully!

kzuberihmf:
After discussion with @pauldwolfe we decided to merge. The matcher code is not wired in, so won't affect current function. There will be follow-up PR's connecting up the functionality along with regression etc.

PR #1130, opened 2025-05-16
ACTIN-1447: Restructure datamodel to expose it for trial-api

dkhartwig:
This PR refactors the codebase to use the `datamodel` component from [trial-api](https://github.com/hartwigmedical/actin-trial-api) as a dependency, eliminating code duplication. The `EligibilityRule` extraction logic from `inclusionCriteria`, also used in trial-api, is now reused to avoid redundancy. (new [solution](https://github.com/hartwigmedical/actin-trial-api/pull/39)).

**Key changes:**

- [x] Restructured components for improved modularity.
- [x] Replaced EligibilityFactory class with a simpler object that generates EligibilityFunction from criteria.
- [x] Delegated parameter verification logic to FunctionInputResolver.
- [x] Changed tests according to new structure 

*Note*:
[Pull request](https://github.com/hartwigmedical/actin-trial-api/pull/39) in trial-api depends on this changes.

### DIFF
--- a/trial/src/main/kotlin/com/hartwig/actin/trial/TrialCreatorApplication.kt
+++ b/trial/src/main/kotlin/com/hartwig/actin/trial/TrialCreatorApplication.kt
@@ -78,7 +78,7 @@ class TrialCreatorApplication(private val config: TrialCreatorConfig) {
             trialIngestion.ingest(objectMapper.readValue(File(config.trialConfigJsonPath), object : TypeReference<List<TrialConfig>>() {}))
 
         val outputDirectory = config.outputDirectory
-        when (result) {
+        when (result) { 
             is Either.Right -> {
                 LOGGER.info("Writing ${result.value.size} trials to [$outputDirectory]")
                 TrialJson.write(result.value, outputDirectory)


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/trial/EligibilityFactory.kt
kduyvesteyn, line 1: This kind of class does not belong in `datamodel` (effectively this module should only contain data classes and interfaces, no logic). I guess you moved it here to allow for reuse in trial-api but  this should be handled differently

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
```diff
@@ -74,9 +76,13 @@ class TrialIngestion(private val eligibilityFactory: EligibilityFactory) {
 
     private fun toEligibility(inclusionCriterion: InclusionCriterionConfig) =
         try {
+            val function = EligibilityFactory.generateEligibilityFunction(inclusionCriterion.inclusionRule).apply {
+                val hasValidInputs = functionInputResolver.hasValidInputs(this)
+                check(!(hasValidInputs == null || !hasValidInputs)) { "Function ${this.rule} has invalid inputs: '${this.parameters}' (source criterion: '${inclusionCriterion.inclusionRule}')" }
+            }
```
kduyvesteyn, line 82: Out of interest, is there a reason why you use `apply` here? Why not just 3 separate commands one after another?

trial/src/test/kotlin/com/hartwig/actin/trial/TestEligibilityFactoryFactory.kt
kduyvesteyn, line 1: I would suggest a different name for this class, it is not anymore a factory to create factories but rather something like `CriterionEvaluator`. You could also consider having this functionality in a real class and calling it from TrialIngestion, even perhaps wrapping the generation of eligibility functions inside there (the logic would be -> 1. create function 2. check function).

kduyvesteyn (commented):
I am missing some context in the description but assuming you want to refactor this code to allow exposing various files outside of ACTIN such that actin-trial-api can depend on it.

My suggestion is to not expose the eligibility factory via datamodel since that doesn't fit with the scope of the module (datamodel only, no functionality), but don't have a concrete suggestion otherwise (though plenty of options)

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/trial/EligibilityFactory.kt
dkhartwig: Fixed. Now, `datamodel` contains only data (as it was before). Moved logic into `common` module. 

Since `common` module already had `datamodel` as dependency, I'll add it to trial-api, means `datamodel` will be auto included as well.

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
dkhartwig: To concisely verify object's property within a block or just 'syntax sugar':)

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/trial/EligibilityFactory.kt
kduyvesteyn: I like your pragmatism but not sure I like the concept of exposing common :) 

common module contains code used by multiple actin components and currently is not used by other projects. There is no backwards compatibility and other things that come with proper dependencies. 

I have no objections to having a dependency on the short but I feel there is a better solution. One idea could be that maybe the trial API should receive proper eligibility functions rather than the inputs to create them (if I understand the architecture correctly). In a way it is a wrong design if both the trial creator and trial api both need to know how to translate a raw criterion string into a proper eligibility function - why can't either one have this functionality exclusively?

dkhartwig: Quick explanation my solution:

1) Existing solution

Every trial ingestion, `trial-api` saves into database information about inclusion criteria (table `inclusion_criteria`). The example of column `inclusion_rule` in this table: 

`AND(HAS_HEMOGLOBIN_MMOL_PER_L_OF_AT_LEAST_X[5], HAS_NEUTROPHILS_ABS_OF_AT_LEAST_X[1.5], HAS_THROMBOCYTES_ABS_OF_AT_LEAST_X[100], HAS_CREATININE_ULN_OF_AT_MOST_X[1.5], HAS_CREATININE_CLEARANCE_CG_OF_AT_LEAST_X[30], HAS_TOTAL_BILIRUBIN_ULN_OF_AT_MOST_X[2], HAS_ALAT_ULN_OF_AT_MOST_X[5], HAS_ASAT_ULN_OF_AT_MOST_X[5])`

2) New Solution
I'll create separate cronjob in actin-pipeline, which will call trial-api to get all unused rules. For that purpose, trial-api should read `inclusion_criteria` table and convert text in `inclusion_rule` column into list of rules. And then compare with all existing rules and return unused. So, as you can see, trial-api should be able to parse text into Eligibility Rule. 

I don't quite understand what do you mean by `trial API should receive proper eligibility functions`. We either should expose something to trial-api or re-implement existing logic in api.

kduyvesteyn: Thanks for the additional context.

I am not super familiar with the trial-api so maybe better for @pauldwolfe to comment, but apparently the trial api persists a criterion as a string. If the trial api would persist the actual `EligibilityFunction` object then your cronjob in actin-pipeline wouldn't have to parse them anymore. 

In a way it is weird that the trial api datamodel supports criteria as strings, since they may be wrong altogether, and this would only be discovered once an actual client pulls the trials from the api and converts them to the real objects. Sounds more logical to catch that upon ingestion rather than distribution

dkhartwig: Updated logic in accordance with our discussion and your proposed plan!

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
```diff
@@ -22,9 +30,11 @@ data class UnmappableCohort(val cohortId: String, val mappingErrors: List<Eligib
 
 data class EligibilityMappingError(val inclusionRule: String, val error: String)
 
+data class TrialWrapper(val trial: Trial, val rules: List<EligibilityRule>)
```
pauldwolfe, line 33: Often when adding these named tuples we'll name them something like `TrialAndEligibilityRules` for clarity.

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
```diff
@@ -65,11 +75,30 @@ class TrialIngestion(private val eligibilityFactory: EligibilityFactory) {
...
+        val (errors, trialWrappers) = trialsAndUnmappableTrials.partitionAndJoin()
+        if (errors.isNotEmpty()) {
+            return errors.left()
+        }
+
+        val (trials, usedRules) = with(trialWrappers) { map { it.trial } to flatMap { it.rules }.distinct() }
+        val eligibilityRulesState = with(EligibilityRule.entries.associateWith { EligibilityRuleUsedStatus.UNUSED }.toMutableMap()) {
+            usedRules.forEach { this[it] = EligibilityRuleUsedStatus.USED }
+            this.map { EligibilityRuleState(it.key, it.value) }
         }
```
pauldwolfe, line 97: Could this be simplified to something like:

```suggestion
EligibilityRule.entries.map { rule ->
            EligibilityRuleState(rule, if (rule in usedRules) EligibilityRuleUsedStatus.USED else EligibilityRuleUsedStatus.UNUSED)
        }
 ```

pauldwolfe (commented):
Looks good! 

Left a few suggestion for small improvements but no need to wait for another approval.

pauldwolfe (approved):

trial/src/main/kotlin/com/hartwig/actin/trial/TrialIngestion.kt
dkhartwig: Yes, but it's more complicated (in terms of O complexity). But I think readability is more important than optimization here:) So, fixed!

andreia-hartwig (approved):

PR #1131, opened 2025-05-16
ACTIN-1979 Remove the usage of pathology reports when extracting IHC Tests

andreia-hartwig:


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardIhcTestExtractor.kt
```diff
@@ -8,101 +8,57 @@ import com.hartwig.actin.clinical.curation.extraction.CurationExtractionEvaluati
...
-
-private const val IHC_STRING = "immunohistochemie"
+import java.time.LocalDate
 
 class StandardIhcTestExtractor(
     private val molecularTestCuration: CurationDatabase<IhcTestConfig>
 ) : StandardDataExtractor<List<IhcTest>> {
     override fun extract(ehrPatientRecord: FeedPatientRecord): ExtractionResult<List<IhcTest>> {
-        val extractedIhcTestsFromTumorDifferentiation = extractIhc(molecularTestCuration, ehrPatientRecord)
         val extractedFromOtherConditions = extractFromOtherConditions(ehrPatientRecord)
```
pauldwolfe, line 17: do we still want to extract from other conditions? I thought they should be classified by now?

andreia-hartwig: You're right, these should be properly classified in actin-feed. I will remove from here.

pauldwolfe (approved):

PR #1134, opened 2025-05-16
ACTIN-1980: Evaluate ERBB2 as molecular event in NSCLC

ninajacobs:
.. as quick improvement.

Better fix will come in https://hartwigmedical.atlassian.net/browse/ACTIN-1986


kduyvesteyn (approved):
Amazing!

PR #1136, opened 2025-05-19
ACTIN-1994: Add HE4 lab measure

ninajacobs:
See also https://hartwigmedical.atlassian.net/browse/ACTIN-1994


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/LabMeasurement.kt
```diff
@@ -119,30 +123,29 @@ enum class LabMeasurement(val display: String, val defaultUnit: LabUnit) : Displ
     GRANULOCYTES_LQ("granulocytes in liquor", LabUnit.BILLIONS_PER_LITER),
     GROWTH_HORMONE("growth hormone", LabUnit.MICROGRAMS_PER_LITER),
     HAPTOGLOBIN("haptoglobin", LabUnit.GRAMS_PER_LITER),
-    FETAL_HEMOGLOBIN_F("HbF fraction", LabUnit.NONE),
-    FETAL_HEMOGLOBIN_PCT("HbF percentage", LabUnit.PERCENTAGE),
-    HUMAN_CHORIONIC_GONADOTROPIN("HCG", LabUnit.INTERNATIONAL_UNITS_PER_LITER),
     HDL_CHOLESTEROL("HDL", LabUnit.MILLIMOLES_PER_LITER),
+    HE4("HE4", LabUnit.PICOMOLES_PER_LITER),
```
cbruel, line 127: I tried to keep the enum values the full names everywhere so we can't accidentally confuse things. So I would rename the enum value to `"HUMAN_EPIDIDYMIS_PROTEIN_4"` (the display can just remain "HE4" so we show the abbreviation on the report)

cbruel (approved):
Looks good!

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/LabMeasurement.kt
ninajacobs: thanks, clear!

PR #1139, opened 2025-05-22
ACTIN-2010 handle null surgery end date

andreia-hartwig:


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/surgery/HasHadAnySurgeryAfterSpecificDateTest.kt
```diff
@@ -29,6 +29,11 @@ class HasHadAnySurgeryAfterSpecificDateTest {
         assertEvaluation(EvaluationResult.FAIL, function.evaluate(withSurgery(surgery(minDate.minusWeeks(4)))))
     }
 
+    @Test
+    fun `Should fail with surgery without end date`() {
```
ninajacobs, line 33: @andreia-hartwig I think to be correct we should resolve to 'undetermined' without end date

ninajacobs (changes requested):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/surgery/HasHadAnySurgeryAfterSpecificDate.kt
```diff
@@ -61,6 +61,10 @@ class HasHadAnySurgeryAfterSpecificDate(private val minDate: LocalDate, private
                 EvaluationFactory.fail("Recent surgery got cancelled")
             }
 
+            record.surgeries.isNotEmpty() && record.surgeries.all { it.endDate == null } -> {
+                EvaluationFactory.undetermined("Undetermined when surgery occurred")
+            }
+
```
ninajacobs, line 67: Quick look but I think this code

surgicalTreatmentsOccurredAfterMinDate.any { it == null } -> {
                EvaluationFactory.undetermined("Undetermined if previous surgery is recent")
            }

also looks for surgeries with unknown date (via oncological history). I think for clarity of what the rule does best to combine this new part with that part (so an or-statement)?

I'm also not sure why you do .all : if there are surgeries that lead to fail and other surgeries that lead to undetermined, the outcome should be undetermined.

ninajacobs (approved):
Looks good!

PR #1141, opened 2025-05-22
ACTIN-1982: Allow setting HRD and gene amp copy nr for panel, allow ploidy & minCopy nr to be null and show copy nr in summary for amp

ninajacobs:
This PR does:
- Allow capturing of gene copy nr in case of amplifications & HR status for panel
- Allow gene copy nr to be null for panel + handle in algo
- Set ploidy to null for panel + handle in algo
- Show gene copy nr for amplifications in molecular summary
- General review/refactoring of GeneIsAmplified and GeneHasSufficientCopyNr, including removing reported/unreported WARN as I think it should be captured in ingestion if this flag matters (WARN_HIGH_COPY_NR_NOISE)
- Allow some HR fields to be null as that is now more appropriate based on panel

@jbartletthmf, I didn't add you to PR as Korneel has already had a small look at it, but is it correct that no changes to actin-feed are needed for this?


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/characteristics/HomologousRecombination.kt
kduyvesteyn, line 1: Every single field can now be null, while the entire HR object inside `MolecularCharacteristics` can also be null. How do they differ? What is a null HR object vs an actual HR object with all fields null?

My recommendation would be to have at least one non-null property, probably `isDeficient` is the most logical candidate?

kduyvesteyn (approved):
Nice job!

kduyvesteyn:
Note that I am not familiar enough with actin-feed to determine whether changes are required there, @jbartletthmf ?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/characteristics/HomologousRecombination.kt
ninajacobs: I did consider making isDeficient non-nullable, but CHORD can lead to undetermined outcomes for isDeficient too... I'll investigate what may be a good option here!

ninajacobs: How about:
- If CHORD = CANNOT BE DETERMINED -> set whole HR object to null
- then isDeficient is not nullable anymore

kduyvesteyn: Yes I think that is a good proposal, assuming we don't use any of the other fields when CHORD = CANNOT_BE_DETERMINED anyways. 

Some related changes I propose alongside:
 - Make `isDeficient` the first property of the class, making it clear it is the central piece and the other fields are further annotations that may or may not be available.
 - Remove `CANNOT_BE_DETERMINED` from the list of potential HomologousRecombinationType since that situation should yield a null object in the first place (in other words, if the HomologousRecombination object inside Characteristics is not-null, then we know for sure whether the result is deficient yes/no)

kduyvesteyn: Haven't checked this but you may be correct there (on 2nd bullet)!

PR #1142, opened 2025-05-22
ACTIN-2014: Only consider prior primary inputs for extraction

jbartletthmf:
Allow one input to be curated into multiple prior primaries.

### DIFF
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
@@ -7,45 +7,30 @@ import com.hartwig.actin.clinical.curation.config.PriorPrimaryConfig
 import com.hartwig.actin.clinical.curation.extraction.CurationExtractionEvaluation
 import com.hartwig.actin.datamodel.clinical.PriorPrimary
 import com.hartwig.actin.datamodel.clinical.ingestion.CurationCategory
-import com.hartwig.feed.datamodel.DatedEntry
 import com.hartwig.feed.datamodel.FeedPatientRecord
 
 class StandardPriorPrimariesExtractor(private val priorPrimaryCuration: CurationDatabase<PriorPrimaryConfig>) :
     StandardDataExtractor<List<PriorPrimary>> {
     override fun extract(ehrPatientRecord: FeedPatientRecord): ExtractionResult<List<PriorPrimary>> {
-        val priorPrimaries = fromPriorPrimaries(ehrPatientRecord)
-        val patientId = ehrPatientRecord.patientDetails.patientId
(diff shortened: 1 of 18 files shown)


clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
```diff
@@ -7,45 +7,30 @@ import com.hartwig.actin.clinical.curation.config.PriorPrimaryConfig
...
-        return sourceList.map { curate(patientId, it.name) }
-            .filter { it.configs.isNotEmpty() }
-            .map {
-                ExtractionResult(it.configs.mapNotNull { c -> c.curated }, it.extractionEvaluation)
-            }
-    }
-
-    private fun fromPriorPrimaries(ehrPatientRecord: FeedPatientRecord): List<ExtractionResult<List<PriorPrimary>>> =
-        ehrPatientRecord.priorPrimaries.map { feedPriorPrimary ->
+        return ehrPatientRecord.priorPrimaries.map { feedPriorPrimary ->
```
pauldwolfe, line 15: maybe should rename ehrPatientRecord to feedPatientRecord?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
```diff
@@ -7,45 +7,30 @@ import com.hartwig.actin.clinical.curation.config.PriorPrimaryConfig
...
             val curatedPriorPrimary = curate(ehrPatientRecord.patientDetails.patientId, feedPriorPrimary.name)
             ExtractionResult(
-                listOfNotNull(
-                    curatedPriorPrimary.config()?.takeUnless { it.ignore }?.curated?.copy(
+                curatedPriorPrimary.configs.filterNot { it.ignore }.mapNotNull {
+                    it.curated?.copy(
                         diagnosedMonth = feedPriorPrimary.startDate?.monthValue,
```
pauldwolfe, line 20: might be confusing if the curation specifies these dates?  In the past we've preferred the curated dates if someone took the time to fill them out. Alternatively we could hide them in the sheet so the will always be null at this point.

pauldwolfe (approved):
Couple suggestions but otherwise looks good.

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardPriorPrimariesExtractor.kt
jbartletthmf: Good idea. Since this is set in the interface, I had to rename for all implementations to support named-argument access.

andreia-hartwig (approved):

PR #1147, opened 2025-05-27
ACTIN-2018: Multiple new things for new trial

cbruel:
* Add L01FE to medication categories
* Add TYROSINE_KINASE_INHIBITOR_GEN_3 (will add to drug.json after merging this PR)
* Add HAS_LUNG_AND_OR_LUNG_LYMPH_NODE_METASTASES_ONLY (after trying to implement this yesterday and this morning, came to the conclusion it's quite a lot of work to do this. And since this rule is for a cohort that is not open in NKI and so will never be an option for a patient, feels like not worth my time, so decided to just evaluate to 'undetermined')


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/treatment/DrugType.kt
```diff
@@ -196,6 +196,7 @@ enum class DrugType(override val category: TreatmentCategory, private val displa
     TYROSINE_KINASE_INHIBITOR(TreatmentCategory.TARGETED_THERAPY),
     TYROSINE_KINASE_INHIBITOR_GEN_1(TreatmentCategory.TARGETED_THERAPY),
     TYROSINE_KINASE_INHIBITOR_GEN_2(TreatmentCategory.TARGETED_THERAPY),
+    TYROSINE_KINASE_INHIBITOR_GEN_3(TreatmentCategory.TARGETED_THERAPY),
```
ninajacobs, line 199: will you update drug.json too?

ninajacobs (approved):

PR #1148, opened 2025-05-27
ACTIN-2020: Add warn for amplifications

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
```diff
@@ -272,8 +272,9 @@ class MolecularRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
 
     private fun anyGeneFromSetIsOverExpressedCreator(): FunctionCreator {
         return { function: EligibilityFunction ->
-            val geneSet = functionInputResolver().createManyGenesInput(function).geneNames
-            AnyGeneFromSetIsOverexpressed(maxMolecularTestAge(), geneSet)
+            val genes = functionInputResolver().createManyGenesInput(function).geneNames
+            val genesToAmplification = genes.associateWith { GeneIsAmplified(it, null, maxMolecularTestAge()) }
+            AnyGeneFromSetIsOverexpressed(maxMolecularTestAge(), genesToAmplification)
```
jbartletthmf, line 277: I feel like it would make more sense to just pass in the genes and max test age here, and add the association with `GeneIsAmplified` in `AnyGeneFromSetIsOverexpressed`. That would limit the complexity of the mapper, which has limited testing, and simplify the constructor parameters of `AnyGeneFromSetIsOverexpressed`.

jbartletthmf (approved):

PR #1151, opened 2025-05-28
ACTIN-2031: Fix sorting for variants in summary

ninajacobs:


kduyvesteyn (approved):
Looks good! A tough one... :)

PR #1154, opened 2025-05-28
ACTIN-1811: Use combo for trial matchers

pauldwolfe:


kduyvesteyn (commented):
This looks really neat!

I would suggest to move `EvidenceAnnotator` and related into common, this code is the main entry point for evidence annotation and the rest of it could be considered private. It feels wrong to have the main entry point into this code reside in another module/package. 

Also, I would suggest to add (at least) one higher level scenarios as a unit test, 
When a trial has the following 4 criteria:
 - One single criterium on non-applicable gene
 - One combined criterium with one leg being a non-applicable gene and the other leg and criterium that would match (e.g. High TMB)
 - One combined criterium with 2 legs that will both match (e.g. BRAF hotspot + High TMB)
 - One single criterium that would match (e.g. BRAF hotspot)

Then when I call ActionabilityMatcher created via the standard factory, with a molecular test that has matching crtiteria (BRAF hotspot + High TMB) I expect the `MatchesForActionable` to contain:
 - One entry for BRAF hotspot with the trial mapped to the 2 matching criteria
 - One entry for High TMB with the same trial mapped to the same 2 matching criteria.

Another remark - I think `ResistanceEvidenceMatcher` is not yet entirely correct, or at least I expect the `ActionabilityMatcher` to be created at a later stage with just the filtered evidence and no trials. I am fine to keep out of scope since this class will be revised in the next month anyways. 

Finally, note that `ServeCleaner` only cleans combined evidence and not combined trials!

kzuberihmf:
Performed regression against current production and both evidences and trials matched up.

PR #1155, opened 2025-05-29
ACTIN-2035: Drop cyberknife as type of radiotherapy

ninajacobs:
See rationale in ticket: https://hartwigmedical.atlassian.net/browse/ACTIN-2035

I also update the treatment and drug JSONs


ninajacobs:
If I understand correctly too :)

PR #1164, opened 2025-06-03
ACTIN-2069: Cleanup unused file

ninajacobs:
Hi, if not mistaken I don't think we need this file anymore (also removed it from the trial databases).


dkhartwig (approved):
Yes, you're right

PR #1176, opened 2025-06-05
ACTIN-2094: Remove combined evidence filter from ACTIN

kzuberihmf:
Changed the trial filtering slightly to drop criteria that are multi but keep singles.

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeCleaner.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeCleaner.kt
@@ -4,12 +4,13 @@ import com.hartwig.serve.datamodel.ImmutableServeDatabase
 import com.hartwig.serve.datamodel.ImmutableServeRecord
 import com.hartwig.serve.datamodel.ServeDatabase
 import com.hartwig.serve.datamodel.ServeRecord
+import com.hartwig.serve.datamodel.trial.ImmutableActionableTrial
 
 object ServeCleaner {
 
     fun cleanServeDatabase(database: ServeDatabase): ServeDatabase {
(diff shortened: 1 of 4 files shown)


common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeCleaner.kt
```diff
@@ -18,21 +19,15 @@ object ServeCleaner {
...
-            .from(record)
-            .evidences(cleanedEvidences)
-            .build()
-    }
-
     private fun cleanCombinedTrials(record: ServeRecord): ServeRecord {
-        val cleanedTrials = record.trials().filterNot { trial ->
-            trial.anyMolecularCriteria().any {
-                ServeVerifier.isCombinedProfile(it)
+        val cleanedTrials = record.trials().mapNotNull { trial ->
```
pauldwolfe, line 23: I thought that trials on combined profiles didn't exist yet in serve? or are they not correct and that's why we filter them?

pauldwolfe (approved):
Looks good, one question for my understanding.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/ServeCleaner.kt
kzuberihmf: They don't exist, that's why I didn't filter them and only filtered the evidences. I think Korneel mentioned this so you added the filter for the trials? And now I'm just tweaking.

So right now, no impact, just theoretical safety. When we add combined molecular criteria trials to SERVE they'll still have no impact, until we explicitly remove the filter. Not actually sure we need it, the point of the filter was that we didn't have combined support implemented in ACTIN but had it for evidences in SERVE. But now we do support 😄

PR #1180, opened 2025-06-06
ACTIN-2102 Minor change to non high driver events + remove ":"

ninajacobs:
Last time! :)


cbruel (approved):
Only noticing now that previously we used ":" only for this row in this table, good catch!

ninajacobs:
me too!

PR #1181, opened 2025-06-06
ACTIN-2058: Remove tumor (sub)location and tumor (sub)type from prior primary data model

ninajacobs:
See title

Upon merge I'll fix the curation sheets (only deleting 4 columns, we have the 'name' column already).


jbartletthmf (approved):
Nice simplification! No notes 🙂

PR #1182, opened 2025-06-06
ACTIN-2067: Add 'name' to PrimaryTumor data model and remove location/type/details

ninajacobs:
WIP


jbartletthmf (commented):
Nice refactor 🙂

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/emc/extraction/TumorDetailsExtractor.kt
```diff
@@ -79,7 +79,8 @@ class TumorDetailsExtractor(
             LesionLocationCategory.LYMPH_NODE
         )
 
-        val hasBrainOrGliomaTumor = primaryTumorDetails.primaryTumorLocation == "Brain" || primaryTumorDetails.primaryTumorType == "Glioma"
+        val hasBrainOrGliomaTumor =
+            primaryTumorDetails.name.lowercase().contains("brain") || primaryTumorDetails.name.lowercase().contains("glioma")
```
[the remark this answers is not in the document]
ninajacobs: I checked the standard extractor too, I think what we did there is not correct. We won't have "brain" anyway (since Dutch) and the dutch word for brain can be wrong (can be brain metastases). I changed that a bit now too

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLocallyAdvancedCancer.kt
```diff
@@ -4,25 +4,26 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
-        return if (isStageMatch(stage, TumorStage.III)) {
-            EvaluationFactory.pass("Stage $stage is considered locally advanced")
-        } else if (isStageMatch(stage, TumorStage.II)) {
-            EvaluationFactory.warn("Unclear if stage $stage is considered locally advanced")
-        } else {
-            EvaluationFactory.fail("Stage $stage is not considered locally advanced")
-        }
-    }
+        val stageIIIEvaluationResult = HasTumorStage(setOf(TumorStage.III)).evaluate(record).result
+        val stageIIEvaluationResult = HasTumorStage(setOf(TumorStage.II)).evaluate(record).result
```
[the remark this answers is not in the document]
ninajacobs: Thanks, I didn't see that correctly and mis-corrected it I think. Would you review if it is now okay?

clinical/src/main/kotlin/com/hartwig/actin/clinical/feed/standard/extraction/StandardTumorDetailsExtractor.kt
```diff
@@ -50,22 +50,23 @@ class StandardTumorDetailsExtractor(
 
     private fun tumorDetails(ehrPatientRecord: FeedPatientRecord): TumorDetails {
         with(ehrPatientRecord.tumorDetails) {
-            val hasBrainOrGliomaTumor =
-                tumorType?.lowercase()?.contains("glioma") == true || tumorLocation?.lowercase()?.contains("brain") == true
+            val hasGliomaTumor =
+                listOf("glioma").any { tumorType?.lowercase()?.contains(it) == true || tumorLocation?.lowercase()?.contains(it) == true }
```
jbartletthmf, line 54:
```suggestion
            val hasGliomaTumor = listOfNotNull(tumorType, tumorLocation).any { it.lowercase().contains("glioma") }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLocallyAdvancedCancer.kt
```diff
@@ -10,20 +10,20 @@ import com.hartwig.actin.datamodel.clinical.TumorStage
 class HasLocallyAdvancedCancer : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
+        val stage = record.tumor.stage ?: return EvaluationFactory.undetermined("Locally advanced cancer undetermined (tumor stage missing)")
         val stageIIIEvaluationResult = HasTumorStage(setOf(TumorStage.III)).evaluate(record).result
         val stageIIEvaluationResult = HasTumorStage(setOf(TumorStage.II)).evaluate(record).result
```
jbartletthmf, line 15: The fact that this evaluation function is wrapped by the derived stage evaluator means we shouldn't have to call functions that use derived stages internally. We should be able to drop these.

The derived evaluator will find the derived stages for this patient in the case where the real stage is not known, and evaluate this function for all derived stages. If there's a unique derived stage, it essentially works the same as if the stage was known. If there are multiple derived stages, it will combine the evaluation results for all of them according to `DerivedTumorStageEvaluationFunction`. For example, if the rule passes for one derived stage but not all, it will return `UNDETERMINED`.

Is there any reason we can't use the derived stage evaluator here? If there is, we could drop its use in the mapper and continue to use the `HasTumorStage` evaluations.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLocallyAdvancedCancer.kt
```diff
@@ -10,20 +10,20 @@ import com.hartwig.actin.datamodel.clinical.TumorStage
...
-        val stageMessage = record.tumor.stage?.display() ?: "(derived stage: ${record.tumor.derivedStages?.joinToString(" or ")})"
 
         return when {
             stageIIIEvaluationResult == EvaluationResult.PASS -> {
```
jbartletthmf, line 18:
```suggestion
            (stage.category ?: stage) == TumorStage.III -> {
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLocallyAdvancedCancer.kt
```diff
@@ -10,20 +10,20 @@ import com.hartwig.actin.datamodel.clinical.TumorStage
...
-                EvaluationFactory.pass("Stage $stageMessage is considered locally advanced")
+                EvaluationFactory.pass("Stage $stage is considered locally advanced")
             }
 
             stageIIEvaluationResult != EvaluationResult.FAIL || stageIIIEvaluationResult != EvaluationResult.FAIL -> {
```
jbartletthmf, line 22:
```suggestion
            (stage.category ?: stage) == TumorStage.II -> {
```

ninajacobs:
Hi John, so if I understand/understood correctly:
- DerivedTumorStageEvaluationFunction handles all rules in case of derived stages. Hence none of the rules have to handle derived stage logic (except for stage messaging - I couldn't find code for that, so I did make a stageMessage for all rules)?
- Based on this, I refactored all the classes, and also added DerivedTumorStageEvaluationFunction for HasTumorStage.

Hope I now interpreted correctly!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTumorStage.kt
```diff
@@ -1,73 +1,31 @@
...
-import com.hartwig.actin.algo.evaluation.EvaluationFactory.undetermined
+import com.hartwig.actin.algo.evaluation.EvaluationFactory
 import com.hartwig.actin.algo.evaluation.EvaluationFunction
-import com.hartwig.actin.algo.evaluation.util.Format
 import com.hartwig.actin.datamodel.PatientRecord
 import com.hartwig.actin.datamodel.algo.Evaluation
-import com.hartwig.actin.datamodel.algo.EvaluationResult
 import com.hartwig.actin.datamodel.clinical.TumorStage
 
 class HasTumorStage(private val stagesToMatch: Set<TumorStage>) : EvaluationFunction {
```
jbartletthmf, line 9: I'm not sure why this class didn't rely on the derived evaluation before (messages? testing?) but this seems like a nice refactor.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasUnresectableCancer.kt
```diff
@@ -9,28 +9,23 @@ import com.hartwig.actin.datamodel.clinical.TumorStage
 class HasUnresectableCancer : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val stage = record.tumor.stage ?: return EvaluationFactory.undetermined(
-            "Undetermined if cancer is unresectable (tumor stage missing)"
-        )
+        val stage =
+            record.tumor.stage ?: return EvaluationFactory.undetermined("Undetermined if cancer is unresectable (tumor stage missing)")
+        val stageMessage =
+            record.tumor.stage?.display() ?: "(derived stage: ${record.tumor.derivedStages?.joinToString(" or ") { it.display() }})"
```
jbartletthmf, line 15: `stage.display()`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasUnresectableCancer.kt
```diff
@@ -9,28 +9,23 @@ import com.hartwig.actin.datamodel.clinical.TumorStage
...
-                EvaluationFactory.fail("No unresectable cancer")
+                EvaluationFactory.fail("No unresectable cancer (stage $stageMessage)")
             }
         }
     }
-
-    companion object {
-        private fun isStageMatch(stage: TumorStage, stageToMatch: TumorStage): Boolean {
-            return stage == stageToMatch || stage.category == stageToMatch
-        }
```
jbartletthmf, line 34: I don't mind using this function in place of the `stage.category ?: stage` evaluation used above. It might actually be a little more obvious and easy to read. We could move this to a helper object/`TumorStage` for reuse so we don't have to remember to check the category everywhere.

jbartletthmf (commented):
I like how much of the tests we can clean up when we rely on the derived stage evaluator.

jbartletthmf:
> * DerivedTumorStageEvaluationFunction handles all rules in case of derived stages. Hence none of the rules have to handle derived stage logic (except for stage messaging - I couldn't find code for that, so I did make a stageMessage for all rules)?

In `DerivedTumorStageEvaluationFunction`:
```kotlin
        val derivedResults = record.tumor.derivedStages?.associateWith { tumorStage -> evaluatedDerivedStage(record, tumorStage) }
        // ...
    }

    private fun evaluatedDerivedStage(record: PatientRecord, newStage: TumorStage): Evaluation {
        return originalFunction.evaluate(
            record.copy(tumor = record.tumor.copy(stage = newStage))
        )
    }
```
The derived stage evaluator evaluates the function for each derived stage assuming that the real stage is the derived stage (`record.copy(tumor = record.tumor.copy(stage = newStage))`), so there's no need to consider the derived stages at all in each function. Does that make sense?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTumorStage.kt
ninajacobs: me neither..!

jbartletthmf (approved):

PR #1183, opened 2025-06-06
ACTIN-1600 Add evaluation rule for HasRadiologicalProgressionFollowingLatestTreatmentLine

beracira-hartwig:
JIRA: https://hartwigmedical.atlassian.net/browse/ACTIN-1600

The approach we take here is to break up the existing `HasHadPDFollowingSomeSystemicTreatments` into two smaller rules `HasHadSomeSystemicTreatments` and `HasRadiologicalProgressionFollowingLatestTreatmentLine` and combine them in the rule mapper, as having two conditions to check in the same function makes it unnecessarily complicated and error messages were not clear.

### DIFF
--- a/algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLineTest.kt
+++ b/algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLineTest.kt
@@ -21,7 +21,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         val treatments = TreatmentTestFactory.withTreatmentHistory(emptyList())
         val evaluation = function.evaluate(treatments)
         assertEvaluation(EvaluationResult.FAIL, evaluation)
-        assertThat(evaluation.failMessages).containsExactly("No systemic treatments found in treatment history.")
+        assertThat(evaluation.failMessagesStrings()).containsExactly("No systemic treatments found in treatment history.")
     }
 
     @Test
@@ -29,7 +29,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         val treatments = listOf(TreatmentTestFactory.treatmentHistoryEntry(setOf(TreatmentTestFactory.treatment("1", false))))
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.FAIL, evaluation)
-        assertThat(evaluation.failMessages).containsExactly("No systemic treatments found in treatment history.")
+        assertThat(evaluation.failMessagesStrings()).containsExactly("No systemic treatments found in treatment history.")
     }
 
     @Test
@@ -49,7 +49,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
             )
         )
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
-        assertThat(evaluation.passMessages).containsExactly("All systemic treatments resulted in progressive disease.")
+        assertThat(evaluation.passMessagesStrings()).containsExactly("All systemic treatments resulted in progressive disease.")
         assertEvaluation(EvaluationResult.PASS, evaluation)
     }
 
@@ -75,7 +75,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.PASS, evaluation)
-        assertThat(evaluation.passMessages).containsExactly("Last systemic treatment resulted in PD (assumed PD is radiological)")
+        assertThat(evaluation.passMessagesStrings()).containsExactly("Last systemic treatment resulted in PD (assumed PD is radiological)")
     }
 
     @Test
@@ -100,7 +100,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = functionCannotAssumePDIfStopYearProvided.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.FAIL, evaluation)
-        assertThat(evaluation.failMessages).containsExactly("Last systemic treament did not result in progressive disease.")
+        assertThat(evaluation.failMessagesStrings()).containsExactly("Last systemic treament did not result in progressive disease.")
     }
 
     @Test
@@ -123,7 +123,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.PASS, evaluation)
-        assertThat(evaluation.passMessages).containsExactly("Last systemic treatment stopped and radiological progression is assumed.")
+        assertThat(evaluation.passMessagesStrings()).containsExactly("Last systemic treatment stopped and radiological progression is assumed.")
     }
 
     @Test
@@ -148,7 +148,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = functionPDMustBeRadiological.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.PASS, evaluation)
-        assertThat(evaluation.passMessages).containsExactly("Last systemic treatment resulted in PD")
+        assertThat(evaluation.passMessagesStrings()).containsExactly("Last systemic treatment resulted in PD")
     }
 
     @Test
@@ -169,7 +169,7 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.UNDETERMINED, evaluation)
-        assertThat(evaluation.undeterminedMessages).containsExactly("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
+        assertThat(evaluation.undeterminedMessagesStrings()).containsExactly("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
     }
 
     @Test
@@ -190,6 +190,6 @@ class HasHadProgressionFollowingLatestTreatmentLineTest {
         )
         val evaluation = function.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
         assertEvaluation(EvaluationResult.UNDETERMINED, evaluation)
-        assertThat(evaluation.undeterminedMessages).containsExactly("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
+        assertThat(evaluation.undeterminedMessagesStrings()).containsExactly("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
     }
 }


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -6,9 +6,34 @@ import com.hartwig.actin.datamodel.PatientRecord
 import com.hartwig.actin.datamodel.algo.Evaluation
 
 //TODO: Implement according to README
-class HasRadiologicalProgressionFollowingLatestTreatmentLine : EvaluationFunction {
+class HasRadiologicalProgressionFollowingLatestTreatmentLine(
+    private val mustBeRadiological: Boolean
```
jbartletthmf, line 10: Given that the name of the function includes "Radiological", I think we can assume this is always true and drop the parameter.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -6,9 +6,34 @@ import com.hartwig.actin.datamodel.PatientRecord
...
+) :
+    EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        return EvaluationFactory.recoverableUndetermined("Radiological progression following latest treatment line undetermined")
+        val undeterminedMessage = "Radiological progression following latest treatment line undetermined"
+
+        val treatmentHistory = record.oncologicalHistory
+        val systemicTreatments =
```
pauldwolfe, line 18: Feels like we could share some code here with HasHadPDFollowingSomeSystemicTreatments? Could be a simple as moving more to SystemicTreatmentAnalyser. Or perhaps we could even use a single rule for both cases with either an different predicate or optional arguments such that we can use a single class for both rules?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -6,9 +6,34 @@ import com.hartwig.actin.datamodel.PatientRecord
...
+            treatmentHistory.filter { SystemicTreatmentAnalyser.treatmentHistoryEntryIsSystemic(it) }
+        val (systemicTreatmentsWithStartDate, systemicTreatmentsWithoutStartDate) = systemicTreatments.partition { it.startYear != null }
+        val lastTreatment = SystemicTreatmentAnalyser.lastSystemicTreatment(systemicTreatmentsWithStartDate)
+        val lastTreatmentResultedInPD =
+            lastTreatment?.let { ProgressiveDiseaseFunctions.treatmentResultedInPD(it) } == true
+
+        return when {
```
jbartletthmf, line 25: Treatments with unknown dates create some complications because we no longer know if the treatment with the latest date occurred last. This may need to be handled as a separate case here that would result in `UNDETERMINED`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -6,9 +6,34 @@ import com.hartwig.actin.datamodel.PatientRecord
...
+            lastTreatmentResultedInPD -> {
+                EvaluationFactory.pass("Last systemic treatment resulted in radiological progression.")
+            }
+            // or stop year is provided, and PD assumed to be radiological
+            lastTreatment?.treatmentHistoryDetails?.stopYear != null && mustBeRadiological -> {
```
pauldwolfe, line 30: This "mustBeRadiological" argument confuses a bit.  The name of this class would suggest that this rule can only be used when progression can assumed to be radiologically confirmed. Whereas in HasHadPDFollowingSomeSystemicTreatments would seem to support both the case where is can and can't be assumed.  Would be good to check with Charlotte on that.

jbartletthmf (commented):
This looks like a good solution, though I think it's missing handling for some edge cases around unknown dates like the reference implementation in `HasHadPDFollowingSomeSystemicTreatments`.

After reviewing that class, it appears that the `EvaluationFunction` requested in this ticket is just a special case of that class/rule. The README entry for the corresponding rule (`HAS_PROGRESSIVE_DISEASE_FOLLOWING_AT_LEAST_X_TREATMENT_LINES` based on `TreatmentRuleMapper`) is:
>  Treatment history > minimal nr of lines in case systemic = 1 => X, Stop reason or best response of latest treatment is PD, or stop year provided and PD can be assumed. Also passes if all treatments have unknown dates and all have stop reason PD.
> In cases where latest treatment cannot be determined, stop reason is toxicity, or stop year not provided, resolve to `UNDETERMINED`

Could the trial requirements be satisfied by `HAS_PROGRESSIVE_DISEASE_FOLLOWING_AT_LEAST_X_TREATMENT_LINES[1]`? This is probably a question for @cbruel, but @beracira-hartwig I'm interested in hearing your thoughts as well.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
jbartletthmf: @pauldwolfe would you agree that this is a special case of `HasHadPDFollowingSomeSystemicTreatments`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
beracira-hartwig: The variable name is confusing i think. It meant to say if the treatment stopped, can we always assume the reason was radiological progression.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
cbruel: I agree as John already says in his comment below that `HasRadiologicalProgressionFollowingLatestTreatmentLine` is `HasHadPDFollowingSomeSystemicTreatments(1, true)`. Though by simply evaluating it like this, the messaging will be ugly. We would like the messaging be more specific about “latest treatment line” instead of stating “at least 1 systemic treatment”. But we could definitely share some code between the two, because they mostly overlap, we just need the make sure the messages a specific enough

jbartletthmf (commented):
A few minor comments, but I like this direction! I think the messages for the combined rules will work out nicely.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLineTest.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentResponse
+import org.assertj.core.api.Assertions.assertThat
+
+
+import org.junit.Test
+
+class HasRadiologicalProgressionFollowingLatestTreatmentLineTest {
+
+    @Test
+    fun shouldFailWhenTreatmentHistoryEmpty() {
```
pauldwolfe, line 16: We take advantage of the kotlin feature to backtick quote test names eg

``fun `Should fail when treatment history is empty`()``

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLineTest.kt
```diff
@@ -0,0 +1,177 @@
...
+                startMonth = null
+            )
+        )
+        val evaluation = FUNCTION.evaluate(TreatmentTestFactory.withTreatmentHistory(treatments))
+        assertEvaluation(EvaluationResult.UNDETERMINED, evaluation)
+        assertThat(evaluation.undeterminedMessages).containsExactly("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
+    }
+
+
+    companion object {
```
pauldwolfe, line 172: no need to use companion object here. You may see this elsewhere but it's normally due to old code being converted from java to kotlin.

Convention would be to make the objects under test fields of the test class.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -4,11 +4,57 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+        return when {
+            systemicTreatments.isEmpty() -> {
+                EvaluationFactory.fail("No systemic treatments found in treatment history.")
+            }
+
+            systemicTreatments.all { ProgressiveDiseaseFunctions.treatmentResultedInPD(it) == true } -> {
+                EvaluationFactory.pass("All systemic treatments resulted in progressive disease.")
+            }
+
+            // check if all treatments without start date agree on PD status.
```
pauldwolfe, line 35: we normally don't comment code, they almost always drift from the code. I think here the code is readable enough without the comment, but you could also consider extracting a well named function or variable instead.

pauldwolfe (approved):
Some small comments to address but otherwise looks ready to merge.  No need for follow up review from my end.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
jbartletthmf: I agree with Paul. For example, if you set:
```kotlin
val treatmentWithoutDateDiffersInPDStatusFromLastTreatment = systemicTreatmentsWithoutStartDate.any {
    (ProgressiveDiseaseFunctions.treatmentResultedInPD(it) == true) != lastTreatmentResultedInPD
}
```
above, the condition below becomes simply `treatmentWithoutDateDiffersInPDStatusFromLastTreatment`, which clearly indicates the case that's being handled.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasRadiologicalProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -4,11 +4,57 @@ import com.hartwig.actin.algo.evaluation.EvaluationFactory
...
+                EvaluationFactory.pass("Last systemic treatment stopped and radiological progression is assumed.")
+            }
+
+            lastTreatment?.treatmentHistoryDetails?.stopReason == StopReason.TOXICITY || lastTreatment?.treatmentHistoryDetails?.stopYear == null -> {
+                EvaluationFactory.fail("Last systemic treament did not result in progressive disease.")
+            }
+
+            else -> {
+                EvaluationFactory.undetermined("Radiological progression following latest treatment line undetermined.");
+            }
```
jbartletthmf, line 57: Why are these evaluations switched compared to `HasHadPDFollowingSomeSystemicTreatments`? That class's `evaluate` function ends with:
```kotlin
            lastTreatment?.let {
                it.treatmentHistoryDetails?.stopReason == StopReason.TOXICITY || it.treatmentHistoryDetails?.stopYear == null
            } == true -> {
                EvaluationFactory.undetermined(undeterminedMessage)
            }

            else -> {
                EvaluationFactory.fail("Has not received at least $minSystemicTreatments systemic treatments with PD")
            }
```
I don't remember why `StopReason.TOXICITY` should lead to `UNDETERMINED` (@cbruel?), but it's captured in the test cases for that class. It makes sense that an unknown stop year would lead to `UNDETERMINED` because it may mean the treatment is ongoing and the final outcome is still unknown.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            lastTreatmentResultedInPD -> {
+                val radiologicalNote = if (mustBeRadiological) " (assumed PD is radiological)" else ""
+                EvaluationFactory.pass("Last systemic treatment resulted in PD$radiologicalNote")
+            }
+
+            lastTreatment?.treatmentHistoryDetails?.stopYear != null && canAssumePDIfStopYearProvided -> {
+                EvaluationFactory.pass("Last systemic treatment stopped and radiological progression is assumed.")
+            }
```
jbartletthmf, line 48: When would it be appropriate to assume PD if the stop year is provided? `canAssumePDIfStopYearProvided` defaults to true and we don't set it in the mapper, so it seems like we always make this assumption, even without checking `TreatmentHistoryDetails.bestResponse`. A treatment history entry could have bestResponse = `REMISSION` and we would still assume PD here if the stop year is set.

I think the intention from the ticket was to rely on `ProgressiveDiseaseFunctions.treatmentResultedInPD` which returns true when `stopReason == null && treatmentDuration != null && treatmentDuration > MIN_WEEKS_TO_ASSUME_STOP_DUE_TO_PD`. This ignores `bestResponse` as well but that's out of scope for this ticket. I believe the correct handling here would be to drop this case from the `when` block.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -0,0 +1,59 @@
...
+
+            lastTreatment?.treatmentHistoryDetails?.stopReason == StopReason.TOXICITY || lastTreatment?.treatmentHistoryDetails?.stopYear == null -> {
+                EvaluationFactory.fail("Last systemic treament did not result in progressive disease.")
+            }
```
jbartletthmf, line 52: The evaluation for
```kotlin
lastTreatment?.treatmentHistoryDetails?.stopReason == StopReason.TOXICITY || lastTreatment?.treatmentHistoryDetails?.stopYear == null
```
flipped from `UNDETERMINED` to `FAIL` between the previous implementation and this one and it isn't clear why. Without other information (such as `lastTreatment?.let { ProgressiveDiseaseFunctions.treatmentResultedInPD(it) } == false`), a treatment with no stop year (or derived stop year) could still be ongoing and it may not be safe to rule out progression.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -0,0 +1,59 @@
...
+
+            else -> {
+                EvaluationFactory.undetermined("Radiological progression following latest treatment line undetermined.");
+            }
```
jbartletthmf, line 56: As long as `canAssumePDIfStopYearProvided` is true, it appears that we'll never reach the `else` condition. Patients with known stop year will pass and those with unknown stop year or no last treatment will fail. This could be resolved by dropping the final `pass` case above.

jbartletthmf (commented):
Sorry for the late reply here but I didn't get an email about your latest commits before merging. The conditions at the bottom of the new `EvaluationFunction` still don't sit well with me and I don't think they match the intention of the ticket.

PR #1188, opened 2025-06-11
ACTIN-1600 Refactor redudent logic in previous version

beracira-hartwig:
Simplify the logic as suggested in #1183. Instead of trying to infer status for stopped treatments, we will reuse results from `treatmentResultedInPD` function.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadProgressionFollowingLatestTreatmentLine.kt
```diff
@@ -38,21 +35,17 @@ class HasHadProgressionFollowingLatestTreatmentLine(
                 EvaluationFactory.undetermined("Unable to determine radiological progression following latest treatment line due to treatments without start date.")
             }
 
-            lastTreatmentResultedInPD -> {
+            lastTreatmentResultedInPD == true -> {
                 val radiologicalNote = if (mustBeRadiological) " (assumed PD is radiological)" else ""
-                EvaluationFactory.pass("Last systemic treatment resulted in PD$radiologicalNote")
+                EvaluationFactory.pass("Last systemic treatment resulted in PD$radiologicalNote.")
```
jbartletthmf, line 40: We don't usually end messages with a period since they may be aggregated in a list.

jbartletthmf (approved):
Nice cleanup! Just a minor comment: the evaluation messages should not end in periods.

PR #1189, opened 2025-06-12
ACTIN-2112: Remove criterion reference from trial datamodel

kduyvesteyn:
I tested on NKI-LUNG-00122 using `0.4.5-beta.1` of ACTIN using test framework (PR-2 from `PatientRunnerTests`) and my impression is that too much was nuked from the trial API. The inclusion criteria configs should still have a list of references (String), similar to the trial sheets.

The extended report should look like the extended report in this branch for the  regression test (system module, example_reports in resources, and the trial details chapter is all the way to the end). 

It may be related - I noticed that `InclusionCriterionConfig` has nullable references but this should always be present (can be empty though).


nedleitch (approved):
Looks good to me. I used this to build my changes for ACTIN-2116 and everything seems to work properly.

Not sure if I've missed you for the day, might merge myself so I can have a proper `actin` version to include in my pipelines changes.

PR #1190, opened 2025-06-12
ACTIN-2115: Populate effects on canonical transcript based off paver

kduyvesteyn:
Note:
 - @pauldwolfe One of the causes of this bug imo was the fact that some properties in `TranscriptVariantImpact` had defaults, I consistently remove these when I run into them, they are dodgy since it doesn't force people to set every field.
 - @beracira if reviewing this PR doesn't fit in your agenda no worries, I assume @spdeleeuw  will review tomorrow when he is back! 
 
Finally, I noticed there is some room for cleanup in molecular test factories but left it out of this change.


spdeleeuw (approved):
LGTM

PR #1191, opened 2025-06-12
ACTIN-2108: Add "no slots" after cohort name and remove grey color

cbruel:
See ACTN01020459 on actin prod vm (yes still need to read up on the new ACTIN Test Framework) for example how this looks like


kduyvesteyn (approved):
I mainly reviewed the changes compared to yesterday and they all look good to me!

PR #1193, opened 2025-06-12
ACTIN-2081 Add new error type in curation when name = <ignore> and other fields are present

beracira-hartwig:
This PR adds a new validation in curation config. The error is raised when `name` field has value `<ignore>` but other fields are also present as described in [ACTIN-2081](https://hartwigmedical.atlassian.net/browse/ACTIN-2081). The existing validation process should already prevent the row being created in the curation db, but it does so silently. The newly added validation aims to make it more explicit.

We're not able to reproduce the error by using the example provided in the JIRA ticket, but one theory is that the name used in the example contains whitespace, i.e. `<ignore> ` instead of `<ignore>`. Therefore another sanitization step is added to the string comparison. 

[ACTIN-2081]: https://hartwigmedical.atlassian.net/browse/ACTIN-2081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/CurationConfigValidation.kt
+++ b/clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/CurationConfigValidation.kt
@@ -6,7 +6,10 @@ import com.hartwig.actin.datamodel.clinical.IcdCode
 import com.hartwig.actin.datamodel.clinical.ingestion.CurationConfigValidationError
 import com.hartwig.actin.icd.IcdModel
 
-data class ValidatedCurationConfig<T : CurationConfig>(val config: T, val errors: List<CurationConfigValidationError> = emptyList())
+data class ValidatedCurationConfig<T : CurationConfig>(
+    val config: T,
+    val errors: List<CurationConfigValidationError> = emptyList()
+)
 
 fun validateDoids(
     curationCategory: CurationCategory,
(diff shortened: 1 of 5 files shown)


clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/CurationConfigValidation.kt
```diff
@@ -137,6 +161,33 @@ inline fun <reified T : Enum<T>> validateEnum(
     }
 }
 
+fun validateIgnore(
```
pauldwolfe, line 164: I think this function is specific to ComplicationConfigFactory so best to keep it there.  The other functions in this class are shared across multiple factories.

clinical/src/test/kotlin/com/hartwig/actin/clinical/curation/config/ComplicationConfigFactoryTest.kt
```diff
@@ -89,4 +89,24 @@ class ComplicationConfigFactoryTest {
             )
         )
     }
+
+    @Test
+    fun `Should return null for curated complication when ignore is true`() {
```
pauldwolfe, line 94: Name of this test isn't accurate

pauldwolfe (changes requested):

beracira-hartwig:
did some unnecessary formatting reverting that

pauldwolfe (approved):

PR #1198, opened 2025-06-17
ACTIN-1707 Merging Comorbidity Configs

beracira-hartwig:
All Comorbidities are currently located in different tabs, we're looking to merge them into one tab and use an additional field (type) to determine the subtype. 

The approach used here is to wrap the existing `ConfigFactory` with another layer `ComorbidityConfigFactory`, which distributes calls to the corresponding sub type's `ConfigFactory` depending on the `type` column.


jbartletthmf (commented):
Logically, this looks good to me, though I think we clean up some redundant code by consolidating or refactoring the individual config factories. For example, each one performs an ICD validation. We should be able to do this once either with `CurationCategory.COMORBIDITY` or by resolving the category from the `type` field.

beracira-hartwig:
> > Logically, this looks good to me, though I think we clean up some redundant code by consolidating or refactoring the individual config factories. For example, each one performs an ICD validation. We should be able to do this once either with `CurationCategory.COMORBIDITY` or by resolving the category from the `type` field.
> 
> I agree we could go further in consolidating. Also think its a bit confusing to have several factory implementations that should never be used standalone. I'd advocate for folding those factories into a single ComorbidityFactory. Is that where you landed @beracira-hartwig @jbartletthmf?

yeah that's what i'm working on right now. though that will be a bigger change than i initially planned.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+import com.hartwig.actin.icd.IcdModel
+import com.hartwig.actin.clinical.curation.CurationUtil
+import com.hartwig.actin.datamodel.clinical.*
+import com.hartwig.actin.datamodel.clinical.ingestion.CurationConfigValidationError
+import com.hartwig.actin.datamodel.clinical.ingestion.CurationCategory
+import kotlin.reflect.full.memberProperties
+
+
+class ComorbidityConfigFactory(private val icdModel: IcdModel) : CurationConfigFactory<ComorbidityConfig> {
+    private val requiredFields = mapOf(
```
beracira-hartwig, line 12: maybe converting this into a function with when will make more sense?

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+        "toxicity" to setOf("name", "grade", "icd"),
+        "complication" to setOf("impliesUnknownComplicationState", "name", "icd", "year", "month"),
+        "other_condition" to setOf("name", "year", "month", "icd", "isLVEF", "lvefValue"),
+        "intolerance" to setOf("name", "icd"),
+        "ecg" to setOf("interpretation", "icd", "isQTCF", "isJTC", "qtcfValue", "qtcfUnit", "jtcValue", "jtcUnit"),
+        "infection" to setOf("interpretation", "icd")
+    )
+
+    private val fieldProcessingFunction = mapOf(
```
beracira-hartwig, line 21: ditto

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
jbartletthmf: I think you could move this validation into the `when` block in `generateConfig` so we only have to resolve the `type` string once. You could have a validation function that takes in a set of fields as a parameter.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
jbartletthmf: Having multiple maps keyed on strings seems pretty brittle. If these fields are meaningful and going to be referenced in multiple places, we could consider adding `enum class ConfigField` in this class and define each one with `val`s representing the string used to access them and the appropriate validation function. The use of generics could make this tough however, so maybe using `when` for validated extraction is a better idea.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+                    EcgMeasure(
+                        value = fieldValues["jtcValue"]!! as Int,
+                        unit = parts[fields["jtcUnit"]!!]
+                    )
+                } else null
+            )
+
+            "infection" ->
+                OtherCondition(
+                    name = fieldValues["interpretation"]!! as String?,
```
jbartletthmf, line 131: I think we could simplify a bit by renaming these `interpretation` fields as `name`. That way, all comorbidities could have validated name and ICD extraction at the top and then conditional extraction of other values within the `when` block.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+        }.toMap()
+        val fieldValues = retval.mapValues { it.value.first }
+        val validationErrors = retval.flatMap { it.value.second }
+        val allErrors = validationErrors.fold(emptyList<CurationConfigValidationError>()) { acc, current ->
+            acc + current
+        }
+        var curated: Comorbidity? = when (type) {
+            "complication" ->
+                Complication(
+                    name = fieldValues["name"]!! as String?,
```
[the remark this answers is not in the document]
jbartletthmf: It will be null if the column does not exist. That shouldn't happen, but it's better to throw an exception in that case than to silently use `""`.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+        parts: Array<String>
+    ): ValidatedCurationConfig<ComorbidityConfig> {
+        val retval = requiredFields[type]!!.map { it ->
+            it to fieldProcessingFunction[it]!!.call(it, fields, parts)
+        }.toMap()
+        val fieldValues = retval.mapValues { it.value.first }
+        val validationErrors = retval.flatMap { it.value.second }
+        val allErrors = validationErrors.fold(emptyList<CurationConfigValidationError>()) { acc, current ->
+            acc + current
+        }
```
jbartletthmf, line 81: Does this step do anything? It converts the list of errors to a list of errors.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+            validateIcd(category, inputValue, fName, flds, prts, icdModel)
+        }
+
+    override fun create(fields: Map<String, Int>, parts: Array<String>): ValidatedCurationConfig<ComorbidityConfig> {
+        val type = parts[fields["type"]!!]
+        val input = parts[fields["input"]!!]
+        val ignore = CurationUtil.isIgnoreString(parts[fields["name"]!!])
+        if (type !in requiredFields.keys) {
+            return generateNoTypeError(type, input, ignore)
+        }
```
jbartletthmf, line 62: If you check the required fields in the `when` block below, you could return this no-type error in the `else` branch.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,178 @@
...
+                )
+
+            "intolerance" ->
+                Intolerance(
+                    name = fieldValues["name"]!! as String?,
+                    icdCodes = fieldValues["icd"]!! as? Set<IcdCode> ?: emptySet()
+                )
+
+            "other_condition" ->
+                OtherCondition(
```
pauldwolfe, line 98: Every subclass of Comorbidity by definition as name, icdCodes, year and month. We should be able to reduce this duplication somehow, any ideas @beracira-hartwig?

ps. ToxicityCuration seems to be a special case, but not sure why exactly @jbartletthmf?.  I would expect in the new single sheet setup it would be fine if a user wanted to add a year/month to a toxicity when it is known.

jbartletthmf: I haven't used delegation in kotlin but it seems like a good fit here. How would `Toxicity` look?

beracira-hartwig: new to kotlin, i'm experimenting to see if withDefaultDate can be populated separately.

beracira-hartwig:
note:                     rawIntolerance.copy(name = curated.name, icdCodes = curated.icdCodes)
                    rawIntolerance.copy(rawIntolerance.baseComorbidity.copy(name = curated.name, icdCodes = curated.icdCodes))

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/Complication.kt
```diff
@@ -7,7 +7,16 @@ data class Complication(
...
-): Comorbidity {
+) : Comorbidity {
+    constructor(
+        baseComorbidity: BaseComorbidity
+    ) : this(
+        baseComorbidity.name,
+        baseComorbidity.icdCodes,
+        baseComorbidity.year,
+        baseComorbidity.month
+    )
```
jbartletthmf, line 18: Given the problem of "we don't want to have to pass 4 identical parameters into every comorbidity implementation's constructor", I think this is a pretty elegant solution. However, I'm not convinced that problem is one we should prioritize. Now, instead of our factory calling:
```kotlin
        val curated: Comorbidity? = when (type) {
            "complication" -> Complication(name, icd, year, month)
            "intolerance" -> Intolerance(name, icd, year, month)
            "other_condition" -> OtherCondition(name, icd, year, month)
            "toxicity" -> ToxicityCuration(name, icd, year, month, grade)
        }
```
it calls:
```kotlin
        val baseComorbidity = BaseComorbidity(name, icd, year, month)
        val curated: Comorbidity? = when (type) {
            "complication" -> Complication(baseComorbidity)
            "intolerance" -> Intolerance(baseComorbidity)
            "other_condition" -> OtherCondition(baseComorbidity)
            "toxicity" -> ToxicityCuration(baseComorbidity, grade)
        }
```
I don't see the savings as significant, and it comes with the cost of a more complicated datamodel (base pseudo-class, extra constructors, additional `ComorbidityClass`). Could we do away with the `BaseComorbidity` and simplify? I know Paul and I encouraged you to pursue some alternatives but I don't see the benefits outweighing the cost here.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/Comorbidity.kt
```diff
@@ -15,4 +15,18 @@ interface Comorbidity : Displayable {
     }
 
     fun withDefaultDate(date: LocalDate): Comorbidity
+}
+
+data class BaseComorbidity(
```
jbartletthmf, line 20: If we do decide to use a helper data class and alternate constructors, is there any reason this class needs to implement `Comorbidity`? Dropping the interface would eliminate the unused overrides, reducing the cost of the alternate constructor approach.

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/ComorbidityConfigFactory.kt
```diff
@@ -0,0 +1,175 @@
...
+        }
+        val isLVEF by lazy { lazyValidateBoolean("isLVEF", fields, parts, allErrors) }
+        val lvefValue by lazy { lazyValidateDouble("lvefValue", fields, parts, allErrors) }
+        val isQTCF by lazy { lazyValidateBoolean("isQTCF", fields, parts, allErrors) }
+        val isJTC by lazy { lazyValidateBoolean("isJTC", fields, parts, allErrors) }
+        val qtcfValue by lazy { lazyValidateInteger("qtcfValue", fields, parts, allErrors) }
+        val qtcfUnit by lazy { lazyValidateString("qtcfUnit", fields, parts) }
+        val jtcValue by lazy { lazyValidateInteger("jtcValue", fields, parts, allErrors) }
+        val jtcUnit by lazy { lazyValidateString("jtcUnit", fields, parts) }
+        val interpretation by lazy { lazyValidateString("interpretation", fields, parts) }
```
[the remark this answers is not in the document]
jbartletthmf: The use of `lazy` is slick here but I'm not sure it's needed. Couldn't we just extract and validate the base fields here using existing tools:
```kotlin
        val name = parts[fields["name"]!!].trim().ifEmpty { null }
        val (icdCodes, icdValidationErrors) = validateIcd(CurationCategory.COMORBIDITY, input, "icd", fields, parts, icdModel)
        val (year, yearValidationErrors) = validateInteger(CurationCategory.COMORBIDITY, input, "year", fields, parts)
        val (month, monthValidationErrors) = validateInteger(CurationCategory.COMORBIDITY, input, "month", fields, parts)
```
and then validate the additional fields for each comorbidity below as needed?
```kotlin
        val (curated, additionalErrors) = when (ComorbidityClass.valueOf(type.uppercase()) {
            ComorbidityClass.COMPLICATION -> Complication(baseComorbidity) to emptyList()
            ComorbidityClass.INTOLERANCE -> Intolerance(baseComorbidity) to emptyList()
            ComorbidityClass.OTHER_CONDITION -> OtherCondition(baseComorbidity) to emptyList()
            ComorbidityClass.TOXICITY -> {
                val (grade, gradeValidationErrors) = validateInteger(CurationCategory.COMORBIDITY, input, "grade", fields, parts)
                ToxicityCuration(baseComorbidity, grade) to gradeValidationErrors
            }
            // and so on...
        }
```
You could then get the total error collection by combining base and additional errors without using any mutable objects.

beracira-hartwig:
Going to close this one as we already moved to actin-clinical.

PR #1199, opened 2025-06-17
ACTIN-2128: Remove driver likelihood column

cbruel:
https://hartwigmedical.atlassian.net/browse/ACTIN-2128


report/src/main/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactory.kt
```diff
@@ -41,7 +41,13 @@ class MolecularDriverEntryFactory(private val molecularDriversInterpreter: Molec
 
     private fun fromVariant(variant: Variant): MolecularDriverEntry {
         val mutationTypeString = formatMutationType(variant)
-        val driverType = "Mutation ($mutationTypeString)"
+        val driverLikelihood = if (variantsGroupedByGene[variant.gene]?.any { it.isCancerAssociatedVariant } == true) {
+            ""
+        } else {
+            variant.driverLikelihood?.let { ", " + formatDriverLikelihood(it) } ?: ""
```
kduyvesteyn, line 47: You could maybe keep "N/A" for non-CAVs without a driver likelihood? 

Also, I think I'd put the logic to handle null DL inside `formatDriverLikelihood` such that it can return `String` rather than `String?`

report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactoryTest.kt
```diff
@@ -108,11 +109,17 @@ class MolecularDriverEntryFactoryTest {
...
-        assertVariantType(oncoNoCav.copy(proteinEffect = ProteinEffect.GAIN_OF_FUNCTION), "Mutation (Gain of function)")
+        assertVariantType(
+            oncoNoCav.copy(proteinEffect = ProteinEffect.UNKNOWN),
+            "Mutation (No known cancer-associated variant, high driver)"
+        )
+        assertVariantType(oncoNoCav.copy(proteinEffect = ProteinEffect.NO_EFFECT), "Mutation (No protein effect, high driver)")
+        assertVariantType(oncoNoCav.copy(proteinEffect = ProteinEffect.NO_EFFECT_PREDICTED), "Mutation (No protein effect, high driver)")
+        assertVariantType(
+            oncoNoCav.copy(proteinEffect = ProteinEffect.LOSS_OF_FUNCTION),
+            "Mutation (No known cancer-associated variant, high driver)"
```
kduyvesteyn, line 120: This test looks a bit weird to me, it is essentially a bug in our datamodel. How can a variant be non-CAV yet have loss-of-function? I think for now I'd just remove these inconsistent tests and at same point improve the datamodel

report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactoryTest.kt
```diff
@@ -178,22 +194,31 @@ class MolecularDriverEntryFactoryTest {
...
         assertVariantType(
             bothNoCavBi.copy(proteinEffect = ProteinEffect.UNKNOWN),
-            "Mutation (No known cancer-associated variant, biallelic)"
+            "Mutation (No known cancer-associated variant, biallelic, high driver)"
+        )
+        assertVariantType(bothNoCav.copy(proteinEffect = ProteinEffect.NO_EFFECT), "Mutation (No protein effect, high driver)")
+        assertVariantType(bothNoCav.copy(proteinEffect = ProteinEffect.NO_EFFECT_PREDICTED), "Mutation (No protein effect, high driver)")
+        assertVariantType(
+            bothNoCav.copy(proteinEffect = ProteinEffect.LOSS_OF_FUNCTION),
+            "Mutation (No known cancer-associated variant, high driver)"
```
kduyvesteyn, line 209: Same here, looks weird... loss of function yet "no known cancer-associated variant"?

kduyvesteyn (approved):
Looks good!

report/src/main/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactory.kt
cbruel: @kduyvesteyn I first let `formatDriverLikelihood` return `String` indeed, but thought we wouldn't want to show any text for `null` driverlikelihood, so ditch the `N/A`? Because otherwise the text would become something like, which doesn't make sense since we don't have the driver likelihood column anymore: `"No known cancer-associated variant, N/A"`. Or where would you put this `N/A`?

kduyvesteyn: You have a good point there, didn't think this through. Empty is better than N/A indeed. 

You could still implement empty as part of `formatDriverLikelihood` I think though (you can transform any driver likelihood to a string in the end)

report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactoryTest.kt
```diff
@@ -273,9 +301,35 @@ class MolecularDriverEntryFactoryTest {
...
+    @Test
+    fun `Should not show driver likelihood for copy numbers and disruptions`() {
+        val copyNumbers = listOf(TestMolecularFactory.createProperCopyNumber())
+        val disruptions = listOf(TestDisruptionFactory.createMinimal().copy(isReportable = true))
+        val record = TestMolecularFactory.createProperTestMolecularRecord().copy(
+            drivers = TestMolecularFactory.createProperTestDrivers()
+                .copy(variants = emptyList(), copyNumbers = copyNumbers, disruptions = disruptions)
+        )
+        val result = createFactoryForMolecularRecord(record).create()
+        assertThat(result[0].driverType).isEqualTo("Deletion")
```
ninajacobs, line 318: maybe add what hom disruptions look like for completeness?

report/src/test/kotlin/com/hartwig/actin/report/interpretation/MolecularDriverEntryFactoryTest.kt
```diff
@@ -273,9 +301,35 @@ class MolecularDriverEntryFactoryTest {
...
+    fun `Should show driver likelihood for fusions and viruses`() {
+        val fusions = listOf(TestMolecularFactory.createProperFusion().copy(driverLikelihood = DriverLikelihood.LOW))
+        val viruses = listOf(TestVirusFactory.createMinimal().copy(isReportable = true, driverLikelihood = DriverLikelihood.MEDIUM))
+        val record = TestMolecularFactory.createProperTestMolecularRecord().copy(
+            drivers = TestMolecularFactory.createProperTestDrivers()
+                .copy(variants = emptyList(), copyNumbers = emptyList(), viruses = viruses, fusions = fusions)
+        )
+        val result = createFactoryForMolecularRecord(record).create()
+        assertThat(result[0].driverType).isEqualTo("Virus (medium driver)")
+        assertThat(result[1].driverType).isEqualTo("Known fusion (low driver)")
```
ninajacobs, line 332: I imagine in case of known fusions we could also drop the driver likelihood? known & low shouldn't be possible?

ninajacobs: we could consider adding 'phasing' information too (eg Known fusion, in frame) or (Promiscuous fusion, skipped exons, low driver)?

cbruel: @ninajacobs I don't think this 'phasing' information is readily available in the datamodel? I can't find it at least :). I think if we want to add this, we should consider that a separate ticket? Or am I overlooking something and can we easily determine this information?

ninajacobs: In that case, nvm! :) Could be that we don't store it easily reachable now!

PR #1201, opened 2025-06-17
ACTIN-2030: Allow interpretation of virus results for panel

ninajacobs:
'Virus' input in curation must be one of: 

SequencedVirusInput {
    EBV,
    HBV,
    HHV_8,
    HPV_HIGH_RISK,
    HPV_LOW_RISK,
    MCV
}

Related changes: integrations is now nullable (+ handle nullable integrations in reporter)

Could you review specifically whether SequencedVirusInput has acceptable naming and location (in SequencingTest)? I wasn't completely sure about that (or the rationale I should use to pick the location). Is it ok to store it in SequencingTest, or should I store it separately (or maybe even in another location)?


clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/SequencingTestResultConfigFactory.kt
```diff
@@ -38,4 +40,7 @@ class SequencingTestResultConfigFactory : CurationConfigFactory<SequencingTestRe
             }
         )
     }
+
+    private fun convertToSequencedVirusInput(input: String?): SequencedVirusInput? =
+        input?.let { SequencedVirusInput.valueOf(it) }
```
jbartletthmf, line 45: This will throw an exception and stop ingestion if the input is invalid. Instead, I think we should use `CurationConfigValidation.validateOptionalEnum` here to return a validation error if the input is invalid.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/SequencingTest.kt
```diff
@@ -39,6 +39,17 @@ data class SequencedSkippedExons(val gene: String, val exonStart: Int, val exonE
...
+data class SequencedVirus(val virus: SequencedVirusInput)
+
+enum class SequencedVirusInput {
+    EBV,
+    HBV,
+    HHV_8,
+    HPV_HIGH_RISK,
+    HPV_LOW_RISK,
+    MCV
+}
```
jbartletthmf, line 51: Given these definitions, I think you could do without the `SequencedVirus` class and just store a `Set<SequencedVirusInput>` in the `SequencingTest`. However, I think it would be better to reference the existing `VirusType` enum and just store a `DriverLikelihood` or boolean here to capture the same information:
```suggestion
data class SequencedVirus(
    val type: VirusType,
    val isLowLikelihood: Boolean = false
)
```

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVirusAnnotator.kt
```diff
@@ -0,0 +1,39 @@
...
+class PanelVirusAnnotator {
+
+    fun annotate(viruses: Set<SequencedVirus>): List<Virus> {
+        return viruses.map { createVirus(it) }
+    }
+
+    private fun createVirus(sequencedVirus: SequencedVirus): Virus {
+        val virusType = resolveVirusNameAndType(sequencedVirus.virus).second
+        return Virus(
+            name = resolveVirusNameAndType(sequencedVirus.virus).first,
```
jbartletthmf, line 19: If `SequencedVirus` holds a reference to `VirusType`, resolving the virus is trivial here and we can do away with the lookup table.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVirusAnnotator.kt
```diff
@@ -0,0 +1,39 @@
...
+            type = virusType,
+            isReliable = true,
+            integrations = null,
+            event = "$virusType positive",
+            isReportable = true,
+            driverLikelihood = if (sequencedVirus.virus == SequencedVirusInput.HPV_LOW_RISK) DriverLikelihood.LOW else DriverLikelihood.HIGH,
```
jbartletthmf, line 25: It looks like we're overloading this enum to store both virus type and driver likelihood. My proposal would separate them and make the likelihood more obvious here.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVirusAnnotator.kt
```diff
@@ -0,0 +1,39 @@
...
+        )
+    }
+
+    private fun resolveVirusNameAndType(sequencedVirusInput: SequencedVirusInput): Pair<String, VirusType> {
+        return when (sequencedVirusInput) {
+            SequencedVirusInput.EBV -> "Epstein-Barr virus" to VirusType.EPSTEIN_BARR_VIRUS
+            SequencedVirusInput.HBV -> "Hepatitis B virus" to VirusType.HEPATITIS_B_VIRUS
+            SequencedVirusInput.HHV_8 -> "Human herpesvirus 8" to VirusType.HUMAN_HERPES_VIRUS_8
+            SequencedVirusInput.HPV_HIGH_RISK, SequencedVirusInput.HPV_LOW_RISK -> "Human papillomavirus" to VirusType.HUMAN_PAPILLOMA_VIRUS
+            SequencedVirusInput.MCV -> "Merkel cell polyomavirus" to VirusType.MERKEL_CELL_VIRUS
```
jbartletthmf, line 36: I would move these names into the enum so they don't have to be defined in this annotator.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/molecular/driver/VirusType.kt
```diff
@@ -1,20 +1,20 @@
 package com.hartwig.actin.datamodel.molecular.driver
 
 enum class VirusType {
```
jbartletthmf, line 3: You could add a `display` field here to facilitate display wherever we use it:
```kotlin
enum class VirusType(private val display: String) : Displayable {
    EPSTEIN_BARR_VIRUS("Epstein-Barr virus"),
    HEPATITIS_B_VIRUS("Hepatitis B virus"),
    HUMAN_HERPES_VIRUS_8("Human herpesvirus 8"),
    HUMAN_PAPILLOMA_VIRUS("Human papillomavirus"),
    MERKEL_CELL_VIRUS("Merkel cell polyomavirus"),
    OTHER("Other");

    override fun display(): String = display
```
Then, the `PanelVirusAnnotator` could simply call `virus.display()`.

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelVirusAnnotator.kt
```diff
@@ -0,0 +1,39 @@
...
+            driverLikelihood = if (sequencedVirus.virus == SequencedVirusInput.HPV_LOW_RISK) DriverLikelihood.LOW else DriverLikelihood.HIGH,
+            evidence = ExtractionUtil.noEvidence()
+        )
+    }
+
+    private fun resolveVirusNameAndType(sequencedVirusInput: SequencedVirusInput): Pair<String, VirusType> {
+        return when (sequencedVirusInput) {
+            SequencedVirusInput.EBV -> "Epstein-Barr virus" to VirusType.EPSTEIN_BARR_VIRUS
+            SequencedVirusInput.HBV -> "Hepatitis B virus" to VirusType.HEPATITIS_B_VIRUS
+            SequencedVirusInput.HHV_8 -> "Human herpesvirus 8" to VirusType.HUMAN_HERPES_VIRUS_8
```
jbartletthmf, line 34: Should there be another space here?

clinical/src/main/kotlin/com/hartwig/actin/clinical/curation/config/SequencingTestResultConfigFactory.kt
ninajacobs: good point - thanks!

PR #1202, opened 2025-06-19
ACTIN-2139 Provide a way to return external trials unfiltered

ArnoHMF:
Added a way to return the external trials without any filtering.

Refactored the TrialProvider so it's a bit easier to test

Extended the tests to it covers more scenarios

### DIFF
--- a/interpretation/src/main/kotlin/com/hartwig/actin/report/trial/TrialsProvider.kt
+++ b/interpretation/src/main/kotlin/com/hartwig/actin/report/trial/TrialsProvider.kt
@@ -60,10 +60,14 @@ class TrialsProvider(
     }
 
     fun externalTrials(): ExternalTrials {
-       return externalTrials(internalTrialIds, eligibleCohortsWithSlotsAvailableAndNotIgnore(), isYoungAdult)
+        return externalTrials(internalTrialIds, eligibleCohortsWithSlotsAvailableAndNotIgnore(), isYoungAdult)
     }
 
-    private fun externalTrials(internalTrialIds:Set<String>, internalEvaluatedCohorts: List<InterpretedCohort>, isYoungAdult: Boolean): ExternalTrials {
+    private fun externalTrials(
+        internalTrialIds: Set<String>,
+        internalEvaluatedCohorts: List<InterpretedCohort>,
+        isYoungAdult: Boolean
+    ): ExternalTrials {
         val eligibleExternalTrials = externalTrials.filterInternalTrials(internalTrialIds)
 
         val (nationalTrials, internationalTrials) = partitionByCountry(eligibleExternalTrials, countryOfReference)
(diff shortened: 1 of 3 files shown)


interpretation/src/main/kotlin/com/hartwig/actin/report/trial/TrialsProvider.kt
```diff
@@ -60,42 +55,40 @@ class TrialsProvider(
...
+        return externalTrials(setOf(), listOf(), false)
     }
 
     fun externalTrials(): ExternalTrials {
-        val evaluated = eligibleCohortsWithSlotsAvailableAndNotIgnore()
+       return externalTrials(internalTrialIds, eligibleCohortsWithSlotsAvailableAndNotIgnore(), isYoungAdult)
+    }
 
-        val eligibleExternalTrials = externalEligibleTrials().filterInternalTrials(treatmentMatch.trialMatches)
+    private fun externalTrials(internalTrialIds:Set<String>, internalEvaluatedCohorts: List<InterpretedCohort>, isYoungAdult: Boolean): ExternalTrials {
```
cbruel, line 66: formatting; missing space after `internalTrialIds:`

interpretation/src/main/kotlin/com/hartwig/actin/report/trial/TrialsProvider.kt
```diff
@@ -104,12 +97,37 @@ class TrialsProvider(
...
         return MolecularFilteredExternalTrials(
-            original.toSet(),
-            (if (enableExtendedMode) original else filtered).toSet()
+            original,
+            if (enableExtendedMode) original else filtered
         )
     }
 
     companion object {
+        fun createTrialsProvider(patientRecord: PatientRecord,
```
cbruel, line 106: This formatting is also off? Would reformat the whole file, I think there are a few formatting issues? Do you use `actin_code_scheme`?

interpretation/src/test/kotlin/com/hartwig/actin/report/trial/TrialsProviderTest.kt
cbruel, line 1: Would also look at the formatting of this file, when I reformat it, a few lines change

interpretation/src/test/kotlin/com/hartwig/actin/report/trial/TrialsProviderTest.kt
```diff
@@ -131,6 +107,61 @@ class TrialsProviderTest {
         assertThat(result).containsExactly(notFiltered)
     }
 
+    @Test
+    fun `externalTrialsUnfiltered should not filter external trials`() {
+        val country1Trial1 = EventWithExternalTrial(EGFR_TARGET, BASE_EXTERNAL_TRIAL.copy(countries = countrySet(NETHERLANDS)))
```
cbruel, line 112: This looks a bit odd to me. I would expect `val trial = EventWithExternalTrial(EGFR_TARGET, BASE_EXTERNAL_TRIAL.copy(countries = countrySet(NETHERLANDS, BELGIUM)))` ?

cbruel (commented):
Logic looks good! Comments mostly on formatting

interpretation/src/test/kotlin/com/hartwig/actin/report/trial/TrialsProviderTest.kt
ArnoHMF: I think you mean combining country1Trial1 and country2Trial1 into one trial because they have the same trial id and molecular target?

I understand why you think it looks a bit weird. But that would actually test something different. And this works for what this intends to test. However, I changed the target and the nctId for the second trial to make that clearer that they are different trials and need to end up in different lists.

interpretation/src/main/kotlin/com/hartwig/actin/report/trial/TrialsProvider.kt
ArnoHMF: I do but I also use the OncoAct project that conveniently has different convcentions. I copied the actin settings into the project, and reformatted, so hopefully this works better.

Maybe you can try to reformat this file once more and let me know if there are still issues?

cbruel (approved):

PR #1203, opened 2025-06-19
ACTIN-2138: Add deserialization of treatment efficacy prediction

kduyvesteyn:
PR now ready for review. Attached how the table looks like for example data.

Note that this is really a first version - there will be many more iterations but I believe the inputs to algo will remain as they are.

Screenshot 2025-06-20 at 16 59 43

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/personalization/serialization/TreatmentEfficacyPredictionJson.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/personalization/serialization/TreatmentEfficacyPredictionJson.kt
@@ -8,7 +8,10 @@ import org.apache.logging.log4j.LogManager
 import java.io.File
 import java.nio.file.Files
 
-private data class TreatmentEfficacyPrediction(@JsonProperty("survival_probs") val survivalProbs: List<Double>)
+private data class TreatmentEfficacyPrediction(
+    @JsonProperty("survival_probs")
+    val survivalProbs: List<Double>
+)
 
 object TreatmentEfficacyPredictionJson {
 
@@ -17,11 +20,12 @@ object TreatmentEfficacyPredictionJson {
 
     fun read(treatmentEfficacyPredictionJson: String): Map<String, List<Double>> {
         logger.info("Loading treatment efficacy predictions from $treatmentEfficacyPredictionJson")
-        val results: Map<String, TreatmentEfficacyPrediction> =
-            mapper.readValue(Files.readString(File(treatmentEfficacyPredictionJson).toPath()))
+
+        val json = Files.readString(File(treatmentEfficacyPredictionJson).toPath())
+        val results = mapper.readValue<Map<String, TreatmentEfficacyPrediction>>(json).mapValues { it.value.survivalProbs }
 
         logger.info(" Loaded treatment efficacy predictions for ${results.keys.size} treatments")
-        
-        return results.mapValues { it.value.survivalProbs }
+
+        return results
     }
 }
--- a/report/src/main/kotlin/com/hartwig/actin/report/interpretation/soc/SurvivalPredictionEntryFactory.kt
+++ b/report/src/main/kotlin/com/hartwig/actin/report/interpretation/soc/SurvivalPredictionEntryFactory.kt
@@ -3,19 +3,19 @@ package com.hartwig.actin.report.interpretation.soc
 object SurvivalPredictionEntryFactory {
 
     fun create(survivalPredictions: Map<String, List<Double>>): List<SurvivalPredictionEntry> {
-        return survivalPredictions.map { (key, value) ->
+        return survivalPredictions.map { (treatment, dailySurvivalProbabilities) ->
             SurvivalPredictionEntry(
-                treatment = key,
-                day30SurvivalProbability = survivalAfterDays(value, 30),
-                day90SurvivalProbability = survivalAfterDays(value, 90),
-                day180SurvivalProbability = survivalAfterDays(value, 180),
-                day360SurvivalProbability = survivalAfterDays(value, 360),
-                day720SurvivalProbability = survivalAfterDays(value, 720)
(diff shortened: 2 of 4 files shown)


common/src/main/kotlin/com/hartwig/actin/personalization/serialization/TreatmentEfficacyPredictionJson.kt
```diff
@@ -1,17 +1,22 @@
...
+
     private val mapper = ObjectMapper().registerModule(KotlinModule.Builder().build())
 
-    fun read(treatmentEfficacyPredictionJson: String): Map<String, TreatmentEfficacyPrediction> {
-        return mapper.readValue(Files.readString(File(treatmentEfficacyPredictionJson).toPath()))
+    fun read(treatmentEfficacyPredictionJson: String): Map<String, List<Double>> {
+        val results: Map<String, TreatmentEfficacyPrediction> =
+            mapper.readValue(Files.readString(File(treatmentEfficacyPredictionJson).toPath()))
+        
+        return results.mapValues { it.value.survivalProbs }
```
jbartletthmf, line 20: You could also do this without the extra assignment if you provide a type hint:
```suggestion
        return mapper.readValue<Map<String, TreatmentEfficacyPrediction>>(Files.readString(File(treatmentEfficacyPredictionJson).toPath()))
            .mapValues { it.value.survivalProbs }
```

report/src/main/kotlin/com/hartwig/actin/report/interpretation/soc/SurvivalPredictionEntryFactory.kt
```diff
@@ -0,0 +1,21 @@
+package com.hartwig.actin.report.interpretation.soc
+
+object SurvivalPredictionEntryFactory {
+
+    fun create(survivalPredictions: Map<String, List<Double>>): List<SurvivalPredictionEntry> {
+        return survivalPredictions.map { (key, value) ->
```
jbartletthmf, line 6: Could you use more descriptive terms here like `(treatment, dailySurvivalProbability) ->`?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/soc/SurvivalPredictionPerTreatmentGenerator.kt
```diff
@@ -0,0 +1,46 @@
...
+        table.addHeaderCell(Cells.createHeader("720-day survival"))
+        
+        val entries = SurvivalPredictionEntryFactory.create(survivalPredictions)
+        entries.forEach { entry ->
+            table.addCell(Cells.createContent(entry.treatment))
+            table.addCell(Cells.createContent(formatSurvivalPercentage(entry.day30SurvivalProbability)))
+            table.addCell(Cells.createContent(formatSurvivalPercentage(entry.day90SurvivalProbability)))
+            table.addCell(Cells.createContent(formatSurvivalPercentage(entry.day180SurvivalProbability)))
+            table.addCell(Cells.createContent(formatSurvivalPercentage(entry.day360SurvivalProbability)))
+            table.addCell(Cells.createContent(formatSurvivalPercentage(entry.day720SurvivalProbability)))
```
jbartletthmf, line 37: Is there an advantage to putting these specific survival intervals in the datamodel? It seems like we could just store the lists and then retrieve survival percentages for arbitrary days on demand at display time. The coupling with the datamodel seems unnecessary.

report/src/test/kotlin/com/hartwig/actin/report/interpretation/soc/SurvivalPredictionEntryFactoryTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+class SurvivalPredictionEntryFactoryTest {
+
+    @Test
+    fun `Should create survival prediction entries when input is empty`() {
+        assertThat(SurvivalPredictionEntryFactory.create(emptyMap())).isEmpty()
+    }
+
+    @Test
+    fun `Should create survival prediction for sparse input`() {
```
jbartletthmf, line 14: Is "sparse" the right word here? That makes me think of arrays with gaps in them, e.g. if some values were null. How about "for input over limited days"?

common/src/main/kotlin/com/hartwig/actin/personalization/serialization/TreatmentEfficacyPredictionJson.kt
```diff
@@ -1,17 +1,22 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.KotlinModule
 import com.fasterxml.jackson.module.kotlin.readValue
 import java.io.File
 import java.nio.file.Files
 
-data class TreatmentEfficacyPrediction(val survivalProbs: List<Double>)
+private data class TreatmentEfficacyPrediction(@JsonProperty("survival_probs") val survivalProbs: List<Double>)
```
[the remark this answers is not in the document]
kduyvesteyn: Agreed

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/soc/SurvivalPredictionPerTreatmentGenerator.kt
kduyvesteyn: Note that `SurvivalPredictionEntry` is a class in the report, not in actin datamodel. I typically like to convert any datamodel to a tabular datamodel with the exact fields, such that rendering becomes trivial. I could also make this class internal to the generator perhaps to isolate it even more, but my take is that since it is within scope of report it is fine anyways

jbartletthmf (approved):

PR #1204, opened 2025-06-19
ACTIN-2138: Deserialize using data class

jbartletthmf:


kduyvesteyn (commented):
Thanks for the suggestions! Have merged them into main branch, but with slight changes:
 - Have renamed property back to `survival_probs` since that is expected, and map the value using @JsonProperty
  - Have made the deserialisation helper class private, I don't think it adds much outside of the scope of deserialisation. All in all looks a lot cleaner though!

PR #1206, opened 2025-06-20
ACTIN-2142: Add new rule HAS_HAD_TREATMENT_WITH_ANY_DRUG_X_AND_AT_LEAST_Y_CYCLES

cbruel:


ninajacobs (commented):
Logics LGTM,

@jbartletthmf would you be willing to review for technical feedback?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCycles.kt
```diff
@@ -0,0 +1,73 @@
...
+        val namesToMatch = drugsToFind.map { it.name.lowercase() }.toSet()
+        val drugList = concatItemsWithOr(drugsToFind)
+
+        val drugsByEvaluationResult: Map<EvaluationResult, Set<Drug>> = effectiveTreatmentHistory
+            .flatMap { entry ->
+                val result = if (minCycles == null) EvaluationResult.PASS else when (entry.treatmentHistoryDetails?.cycles) {
+                    null -> EvaluationResult.UNDETERMINED
+                    in minCycles..Int.MAX_VALUE -> EvaluationResult.PASS
+                    else -> EvaluationResult.FAIL
+                }
```
jbartletthmf, line 27: The name could be better, and I might have used a more general `when` to handle all cases:
```suggestion
                val hasCycles = entry.treatmentHistoryDetails?.cycles.let { cycles ->
                    when {
                        minCycles == null -> EvaluationResult.PASS
                        cycles == null -> EvaluationResult.UNDETERMINED
                        cycles >= minCycles -> EvaluationResult.PASS
                        else -> EvaluationResult.FAIL
                    }
                }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCycles.kt
```diff
@@ -0,0 +1,73 @@
...
+                entry.allTreatments()
+                    .mapNotNull { it as? DrugTreatment }
+                    .flatMap { it.drugs }
+                    .filter { it.name.lowercase() in namesToMatch }
+                    .map { result to it }
```
jbartletthmf, line 32: I don't think it's appropriate to associate `allTreatments` with the cycle evaluation above, since each additional treatment stage (switch/maintenance treatments) has distinct cycle information. It would be better to evaluate the cycles for each treatment stage as well as the primary treatment.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCycles.kt
```diff
@@ -0,0 +1,73 @@
...
+                EvaluationFactory.undetermined(
+                    "Has received treatments with ${concatItemsWithAnd(matchingDrugs)} " +
+                            "but undetermined if at least $minCycles cycles"
+                )
+            }
+
+            effectiveTreatmentHistory.any {
+                it.isTrial && it.allTreatments().any { treatment ->
+                    (treatment as? DrugTreatment)?.drugs?.isEmpty() ?: treatment.categories().isEmpty()
+                }
```
jbartletthmf, line 55: Our convention is to use `com.hartwig.actin.algo.evaluation.treatment.TrialFunctions#treatmentMayMatchAsTrial` for cases like this.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCycles.kt
```diff
@@ -0,0 +1,73 @@
...
+            }
+            .groupBy({ it.first }, { it.second })
+            .mapValues { it.value.toSet() }
+
+        return when {
+            drugsByEvaluationResult[EvaluationResult.PASS]?.isNotEmpty() == true -> {
+                val matchingDrugs = drugsByEvaluationResult[EvaluationResult.PASS]!!
+                val cyclesString = minCycles?.let { " for at least $minCycles cycles" } ?: ""
+                EvaluationFactory.pass("Has received treatments with ${concatItemsWithAnd(matchingDrugs)}$cyclesString")
+            }
```
jbartletthmf, line 42: I always look to avoid using `!!` even though it's reasonable in this case. Here, you could pre-assign all the groups and then check them in the `when` block. With `groupBy`, we can just do a null check since it will never create an empty group:
```suggestion
        val (drugsMatchingCycles, drugsWithUnknownCycles, drugsNotMatchingCycles) =
            listOf(EvaluationResult.PASS, EvaluationResult.UNDETERMINED, EvaluationResult.FAIL).map(drugsByEvaluationResult::get)

        return when {
            drugsMatchingCycles != null -> {
                val cyclesString = minCycles?.let { " for at least $minCycles cycles" } ?: ""
                EvaluationFactory.pass("Has received treatments with ${concatItemsWithAnd(drugsMatchingCycles)}$cyclesString")
            }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCyclesTest.kt
```diff
@@ -0,0 +1,149 @@
...
+        val medication = WashoutTestFactory.medication().copy(
+            drug = Drug(
+                name = "match", category = TREATMENT_CATEGORY, drugTypes = setOf(DrugType.HER2_ANTIBODY)
+            )
+        )
+        assertEvaluation(
+            EvaluationResult.PASS,
+            functionWithoutCycles.evaluate(TreatmentTestFactory.withTreatmentsAndMedications(treatmentHistory, listOf(medication)))
+        )
+    }
```
jbartletthmf, line 148: Test cases should be added to verify the expected behaviour around switch and maintenance treatments.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugWithCycles.kt
cbruel: @jbartletthmf Thanks! Wasn't aware of this. I've tried to use `HasHadCombinedTreatmentNamesWithCycles` as an example to help me how to implement this

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugAndCycles.kt
```diff
@@ -0,0 +1,77 @@
...
+                            cycles >= minCycles -> EvaluationResult.PASS
+                            else -> EvaluationResult.FAIL
+                        }
+                    }
+
+                    val matchingDrugs = matchingEntry.allTreatments()
+                        .mapNotNull { it as? DrugTreatment }
+                        .flatMap { it.drugs }
+                        .filter { it.name.lowercase() in namesToMatch }
+                    hasCycles to matchingDrugs
```
jbartletthmf, line 39: It might help with readability here if you moved this code into a helper function.

jbartletthmf (approved):

PR #1213, opened 2025-06-25
ACTIN-2140: Remove clinical module as it moves to actin-clinical repo

pauldwolfe:


jbartletthmf (approved):
Never liked it anyway 😛

PR #1216, opened 2025-06-27
ACTIN-1494: Remove "profile" from environment config

kduyvesteyn:
To make sure @pauldwolfe cannot test us anymore :)

This removal was on my list for a while, decided today's bug is a good trigger to actually do it! 

I couldn't find any usage of the `profile` param in either pipelines or infra so assume change be merged without risk.


pauldwolfe (approved):
I still think the profiles are a nice idea but indeed not really used currently so best to just to remove and revisit.

kduyvesteyn:
> I still think the profiles are a nice idea but indeed not really used currently so best to just to remove and revisit.

To be clear: I think they are a nice idea as well but I feel they have become redundant by the way we configure our environment configs (clean-and-easy, single place, and also more explicit / not requiring code changes)

PR #1217, opened 2025-06-30
ACTIN-2161: Add rule DRIVER_EVENT_IN_ANY_GENE_WITH_APPROVED_THERAPY_A…

spdeleeuw:
…VAILABLE


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/MolecularRuleMapper.kt
```diff
@@ -115,6 +117,17 @@ class MolecularRuleMapper(resources: RuleMappingResources) : RuleMapper(resource
         )
     }
 
+    private fun hasMolecularDriverEventInAnyGenesWithApprovedTherapyAvailableCreator(): FunctionCreator {
+        return {
+            AnyGeneHasDriverEventWithApprovedTherapy(
+                emptySet(),
```
cbruel, line 123: I think passing `emptySet()` here will lead to wrong evaluations for NSCLC patients? We should pass `null`.

cbruel (approved):

PR #1218, opened 2025-06-30
ACTIN-2162: add rule TUMOR_INFILTRATES_IN_BLOOD_VESSEL

spdeleeuw:


algo/README.md
```diff
@@ -174,6 +174,7 @@ The following criterion algorithms ('rules') are available:
...
+| TUMOR_INFILTRATES_IN_BLOOD_VESSEL                                            | Currently resolves to `UNDETERMINED`                                                                                                                                                                                                                                                                                                                                                                        |                                                                                                                                                                                                                                                                                                                                                                                                       |
```
cbruel, line 177: Would name this rule something like `HAS_TUMOR_INFILTRATING_IN_BLOOD_VESSEL` (to be similar to our other rules) or maybe even better something like `HAS_LESIONS_INFILTRATING_BLOOD_VESSEL`?

cbruel: And update the class name and function name in TumorRuleMapper as well

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/TumorInfiltratesBloodVessel.kt
```diff
@@ -0,0 +1,13 @@
+package com.hartwig.actin.algo.evaluation.tumor
+
+import com.hartwig.actin.algo.evaluation.EvaluationFactory
+import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+
+class TumorInfiltratesBloodVessel: EvaluationFunction {
```
cbruel, line 8: Add test file for completeness?

algo/README.md
cbruel, line 177: Maybe you can add this rule under `HAS_LESIONS_CLOSE_TO_OR_INVOLVING_AIRWAY` instead of here?

cbruel (approved):
Approved with some comments

PR #1221, opened 2025-07-01
ACTIN-2019: Add new rule HAS_CANCER_WITH_LARGE_CELL_COMPONENT

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/TumorEvaluationFunctions.kt
```diff
@@ -24,6 +24,12 @@ object TumorEvaluationFunctions {
         return hasSmallCellDoid || hasSmallCellName
     }
 
+    fun hasTumorWithLargeCellComponent(doidModel: DoidModel, tumorDoids: Set<String>?, tumorName: String): Boolean {
+        val hasLargeCellDoid = isOfAtLeastOneDoidType(doidModel, tumorDoids, DoidConstants.LARGE_CELL_CANCER_DOIDS)
+        val hasLargeCellName = TumorTermConstants.LARGE_CELL_TERMS.any { tumorName.lowercase().contains(it) }
```
spdeleeuw, line 29: I think you can use `ValueComparison.stringCaseInsensitivelyMatchesQueryCollection` here?

PR #1231, opened 2025-07-07
ACTIN-2212: Remove sampleId from treatment match

kduyvesteyn:
Property doesn't make sense anymore on the level of treatment match


cbruel (approved):
Nice clean-up

PR #1236, opened 2025-07-09
ACTIN-2225: Change "no mutations found" to "no relevant genomic alter…

spdeleeuw:
…ations found" to be more accurate for all tests


report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/WGSSummaryGeneratorFunctions.kt
```diff
@@ -71,7 +71,7 @@ object WGSSummaryGeneratorFunctions {
             if (filteredContents.isEmpty() && !hasTmbData &&
                 actionableEventsWithLowOrMediumDriver.isEmpty() && actionableEventsWithUnknownDriver.isEmpty()
             ) {
-                table.addCell(Cells.createSpanningContent("No mutations found", table))
+                table.addCell(Cells.createSpanningContent("No relevant genomic alterations found", table))
```
ninajacobs, line 74: Maybe remove "genomic", since Archer is RNA, and I don't think 'genomic' adds anything except making the sentence longer?

Or maybe even simply "None"?

ninajacobs (approved):
see comment :)

PR #1237, opened 2025-07-09
Add two body location categories

ninajacobs:


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/BodyLocationCategory.kt
```diff
@@ -14,6 +14,7 @@ enum class BodyLocationCategory(private val display: String): Displayable {
     GALLBLADDER("gallbladder"),
     HEAD_AND_NECK("head and neck"),
     KIDNEY("kidney"),
+    LARGE_INTESTINE("large intestine"),
```
cbruel, line 17: There is a large overlap between "colorectum" and "large intestine"? Do we really need both?

large intestine = cecum, colon, rectum, anal canal
colorectum = colon, rectum

PR #1254, opened 2025-07-14
ACTIN-2275: Add new rule HAS_HISTORY_OF_GASTROINTESTINAL_FISTULA

cbruel:
I used the ICD constants you mentioned in your comment on the trial, though doubting a bit that I would think gastrointestinal tract includes also stomach / esophagus? Shouldn't we include fistula ICD constants for those organs too?


spdeleeuw (commented):
I agree with addition of oesophagus and stomach

spdeleeuw (approved):

PR #1256, opened 2025-07-14
ACTIN-2283: Show 'other mutations' in Molecular Details

ninajacobs:
I first made a second boolean but then felt probably better to change isShort into an enum?


report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularSummaryFunctions.kt
```diff
@@ -0,0 +1,22 @@
...
+
+object MolecularSummaryFunctions {
+
+    enum class SummaryType {
+        SHORT_SUMMARY,
+        LONG_SUMMARY,
+        FULL
+    }
+
+    fun selectMolecularSummary(isMolecularDetailsPage: Boolean, experimentType: ExperimentType? = null): SummaryType {
```
cbruel, line 15: I would simply make this file an `enum class SummaryType` and remove this function. As commented above you can easily inline `SummaryType.FULL` in `MolecularDetailsChapter` and then this function is only still used in one place (`MolecularSummaryGenerator`) where you could add a private function to `MolecularSummaryGenerator` which you can use in line 80.

```
private fun selectMolecularSummary(experimentType: ExperimentType? = null): SummaryType {
return when {
            experimentType == HARTWIG_TARGETED || experimentType == PANEL -> SummaryType.SHORT_SUMMARY
            else -> SummaryType.LONG_SUMMARY
        }
```

cbruel (commented):
Looks good. Few suggestions

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/MolecularSummaryFunctions.kt
ninajacobs: very much agreed!

PR #1277, opened 2025-07-31
ACTIN-2313: Take into account exon information for PROMISCUOUS fusions in panel annotation

cbruel:
@ninajacobs I could also ask Korneel next week to review if you prefer
See: https://hartwigmedical.atlassian.net/browse/ACTIN-2313?atlOrigin=eyJpIjoiNjY4MDQ5MDVhNjE0NDAzYjkwODg1Nzk0OTI4MzFhZTYiLCJwIjoiaiJ9


molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
```diff
@@ -89,6 +95,34 @@ class PanelFusionAnnotator(
         return FusionDriverType.NONE
     }
 
+    fun isPromiscuousWithMatchingExons(driverType: FusionDriverType, sequencedFusion: SequencedFusion): Boolean? {
+        return when (driverType) {
+            FusionDriverType.PROMISCUOUS_3 -> {
+                sequencedFusion.exonDown?.let {
+                    knownFusionCache.withinPromiscuousExonRange(
+                        KnownFusionType.PROMISCUOUS_3,
+                        sequencedFusion.transcriptDown,
```
cbruel, line 104: I don't know much about transcripts, but hopefully one of you knows: For the specific example we found (see jira ticket) I can find a transcript in the full pathology report for the FGFR2 gene which starts with NM_.. , the canonical transcript for FGFR2 we have is ENST .. . So I think these transcripts will never match up and this function will always return false? Isn't this a problem for panels we have in more places?

cbruel: Thanks both. Changed code to always use the canonical transcript

ninajacobs (commented):
Hereby my thoughts, sorry have to leave now.. @kzuberihmf would you have a look too?

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
```diff
@@ -89,12 +96,48 @@ class PanelFusionAnnotator(
...
+            }
+
+            FusionDriverType.PROMISCUOUS_5 -> {
+                sequencedFusion.exonUp?.let {
+                    knownFusionCache.withinPromiscuousExonRange(
+                        KnownFusionType.PROMISCUOUS_5,
+                        sequencedFusion.geneUp?.let { gene ->
+                            val canonicalTranscript = canonicalTranscriptForGene(gene)
+                            checkCanonicalTranscript(canonicalTranscript, gene, it)
+                            canonicalTranscriptForGene(gene).transcriptName()
```
kzuberihmf, line 123: Is calling canonicalTranscriptForGene(gene) again redundant here, can use canonicalTranscript from above?

molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
```diff
@@ -114,11 +157,17 @@ class PanelFusionAnnotator(
...
         val geneData = ensembleDataCache.findGeneDataByName(gene)
             ?: throw IllegalArgumentException("No gene data found for gene $gene")
-        val transcript = ensembleDataCache.findCanonicalTranscript(geneData.geneId())?.transcriptName()
+        return ensembleDataCache.findCanonicalTranscript(geneData.geneId())
             ?: throw IllegalStateException("No canonical transcript found for gene $gene")
-        return transcript
+    }
+
+    private fun checkCanonicalTranscript(transcript: TranscriptData, gene: String, exonEnd: Int) {
+        val exonSize = transcript.exons().size
```
kzuberihmf, line 168: I realize you just renamed this, but I still misread it as meaning the size of a particular exon in bases 😄. Don't have a further suggestion though.

molecular/src/test/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotatorTest.kt
```diff
@@ -101,6 +108,117 @@ class PanelFusionAnnotatorTest {
...
+                matchingExon,
+                matchingExon
+            )
+        } returns true
+
+        assertThat(annotator.isPromiscuousWithMatchingExons(FusionDriverType.PROMISCUOUS_5, fusionMatchingExons)).isEqualTo(true)
+    }
+
+    @Test
+    fun `Should return true for promiscuous 5 fusion outside exon range`() {
```
kzuberihmf, line 190: Should return *false*?

kzuberihmf (approved):

PR #1282, opened 2025-08-05
ACTIN-2357: Add two new rules

cbruel:
Last one, promise :). And this one is easy to review. Will create jira ticket to add implementation later


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentCategoryOfOnlyTypesAndMinimumMonthsAsMostRecent.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentType
+
+class HasHadTreatmentCategoryOfOnlyTypesAndMinimumMonthsAsMostRecent(
+    private val category: TreatmentCategory,
+    private val types: Set<TreatmentType>,
+    private val months: Int
+) : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        return EvaluationFactory.undetermined(
```
spdeleeuw, line 18: Is this implementation temporarily? I think we should be able to actually evaluate this with the data we have?

spdeleeuw (approved):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentCategoryOfOnlyTypesAndMinimumMonthsAsMostRecent.kt
cbruel: @spdeleeuw Yes agree, I wanted to add this as undetermined now, to be able to add this trial and will add implementation later

PR #1289, opened 2025-08-14
ACTIN-2407: Evaluate to LOW DL and provide logging instead of exception when exon not on canonical transcript

cbruel:
After my changes in https://github.com/hartwigmedical/actin/pull/1277 where I throw an exception if the provided exon is not present on the canonical transcript, Nina (and I) felt like a warning would be more appropriate here, since the input isn't really invalid (the exon (probably) exists on the provided transcript). Therefore changed to a warn and evaluating to LOW DL if the provided exon is not present on the canonical transcript.

We also created this related ticket: https://hartwigmedical.atlassian.net/browse/ACTIN-2406

### DIFF
--- a/molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
+++ b/molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
@@ -104,7 +104,7 @@ class PanelFusionAnnotator(
                         KnownFusionType.PROMISCUOUS_3,
                         sequencedFusion.geneDown?.let { gene ->
                             val canonicalTranscript = canonicalTranscriptForGene(gene)
-                            checkCanonicalTranscript(canonicalTranscript, gene, it)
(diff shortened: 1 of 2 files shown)


molecular/src/main/kotlin/com/hartwig/actin/molecular/panel/PanelFusionAnnotator.kt
kduyvesteyn, line 1: I find it somewhat confusing that you "return false" somewhere in the middle of a if-cascade. 

I would refactor such that the function does the following:
 1. Resolve the gene of interest (up or down)
 2. Resolve exon of interest (up or down)
 3. Check if exon exists on canonical transcript of gene, return false if not.
 4. Check if exon is within promiscuous exon range.

Might be worth a try if that makes things more readable? 

Somewhat related - there are many "it" references in the `isPromiscuousWithMatchingExons`, that may add to my level of confusion as well

kduyvesteyn:
To be clear: Logically looks all correct to me and a good improvement! So feel free to merge once you are happy with dealing with (or ignoring!) my feedback

PR #1290, opened 2025-08-14
ACTIN-2411: Make exon/codon rank available for OncoActin

LiekeSchoenmaker:
Make a AggregatedEvidenceKey, which made all the relevant driver information available for OncoActin instead of only the event

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/molecular/interpretation/AggregatedEvidence.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/molecular/interpretation/AggregatedEvidence.kt
@@ -9,8 +9,8 @@ import com.hartwig.actin.datamodel.molecular.evidence.TreatmentEvidenceCategorie
 import com.hartwig.actin.datamodel.molecular.evidence.TreatmentEvidenceCategories.suspectResistant
 
 data class AggregatedEvidence(
-    val treatmentEvidencePerEvent: Map<String, Set<TreatmentEvidence>> = emptyMap(),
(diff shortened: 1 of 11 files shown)


kduyvesteyn (changes requested):
A few issues:
 - I don't understand the rationale behind this change, and ACTIN-2411 doesn't seem to exist (??) so can't check there. 
 - The abstractions are not clean. Adding a codon and exon index to a datamodel that should work for any type of actionable, but only exists for variants is no clean design. 

My impression is that this task is fairly related to ACTIN-1966, and my last comment on https://github.com/hartwigmedical/actin/pull/1119 would be a cleaner route to get evidence codon and exon rank back into evidence datamodel (although adding it to MolecularMatchDetails is not a nice abstraction either, especially since these fields are already part of Variant datamodel anyways

ArnoHMF:
I have looked into this some more, and if we skip using the AggregateEvidenceFactory in OncoACTIN and instead use the ACTIN data model directly we can solve this without any changes in the ACTIN code. 

We can also solve the other remaining naming difference that way by overriding the genomic event names used in that AggregateEvidenceFactory. 

I suggest we take this route first, and that might result in a better replacement for the AggregateEvidenceFactory in ACTIN at a later stage. The AggregateEvidenceFactory doesn't do all that much really and it's used with a limited scope. 

This would mean that we abandon both this branch and the one for ACTIN-1966.

I expect to check in a solution in OncoAct later this week that solves the remaining naming differences.

PR #1295, opened 2025-08-15
ACTIN-2414: Make lab and diagnosis nullable

cbruel:


report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/molecular/PathologyReportFunctions.kt
```diff
@@ -46,7 +46,7 @@ object PathologyReportFunctions {
                         Text(" ("),
                     ),
                     listOfNotNull(
-                        pathologyReport.lab.takeIf { it.isNotEmpty() }?.let {
+                        pathologyReport.lab?.let {
```
andreia-hartwig, line 49: Perhaps we should still check for empty values since these are mandatory for NKI.

andreia-hartwig (approved):

PR #1297, opened 2025-08-19
ACTIN-2427 Derive latestWho and latestAsa from whoStatuses and asaScores

andreia-hartwig:


kduyvesteyn (approved):
Looks good to me. 

I probably wouldn't use `LocalDate.now()` inside test functions but no super strong opinion either way.

PR #1298, opened 2025-08-19
ACTIN-2390: Return drug and treatment listings from database

jbartletthmf:
Needed for populating a list of treatments in the feed review tool.


andreia-hartwig (approved):

jbartletthmf:
It turns out that I need the whole treatment object in the listing. Sorry for the mistake.

PR #1299, opened 2025-08-19
ACTIN-2427 ensure the same gson is used for fromJson and toJson function for ClinicalRecord, Trial and TreatmentMatch

andreia-hartwig:


common/src/main/kotlin/com/hartwig/actin/algo/serialization/TreatmentMatchJson.kt
```diff
@@ -38,20 +40,23 @@ object TreatmentMatchJson {
...
-        return gson.fromJson(json, TreatmentMatch::class.java)
+        return gsonBuilder().fromJson(json, TreatmentMatch::class.java)
     }
+
+    private fun gsonBuilder() = GsonBuilder().serializeNulls()
+        .enableComplexMapKeySerialization()
+        .serializeSpecialFloatingPointValues()
+        .registerTypeAdapter(LocalDateTime::class.java, GsonLocalDateTimeAdapter())
+        .registerTypeAdapter(LocalDate::class.java, GsonLocalDateAdapter())
+        .registerTypeHierarchyAdapter(Set::class.java, GsonSetAdapter<Any>())
```
jbartletthmf, line 55: Was it helpful to have a generic serializer to build on in these cases? It doesn't seem like an improvement to have to repeat all this code everywhere. If we wanted to add something like the set adapter in the future, we'd have to do it in a lot more places.

common/src/main/kotlin/com/hartwig/actin/util/json/EvaluationMessageAdapter.kt
```diff
@@ -1,27 +1,29 @@
...
             return
         }
-
-        val jsonObject = gson.toJsonTree(StaticMessage(value.toString())).asJsonObject
-        gson.toJson(jsonObject, out)
+        
+        val staticMsg = StaticMessage(value.toString())
+        out.beginObject()
+        out.name("message").value(staticMsg.message)
+        out.endObject()
```
jbartletthmf, line 21: Why was this change required? With this change, is there any need to create a `StaticMessage`?

common/src/test/kotlin/com/hartwig/actin/util/json/GsonSerializerTest.kt
```diff
@@ -1,57 +0,0 @@
...
-        val deserialized = gson.fromJson<List<String>>(gson.toJson(set), object : TypeToken<List<String>>() {}.type)
-        assertThat(deserialized).isEqualTo(listOf("2", "3", Country.NETHERLANDS.toString(), null))
-    }
-
-    @Test
-    fun `Should sort nulls last`() {
-        val set = setOf(Intent.PALLIATIVE, null, Intent.ADJUVANT)
-        val deserialized = gson.fromJson<List<Intent?>>(gson.toJson(set), object : TypeToken<List<Intent?>>() {}.type)
-        assertThat(deserialized).isEqualTo(listOf(Intent.ADJUVANT, Intent.PALLIATIVE, null))
-    }
```
jbartletthmf, line 43: Are these criteria still tested somewhere?

common/src/main/kotlin/com/hartwig/actin/util/json/EvaluationMessageAdapter.kt
andreia-hartwig: This version does the same as the previous but does not require the dependency on Gson.

common/src/main/kotlin/com/hartwig/actin/algo/serialization/TreatmentMatchJson.kt
andreia-hartwig: It wasn't used consistently, and because it doesn't offer all the type adaptors needed for all objects, some classes were creating a generic GsonBuilder() - for instance PatientRecordJson, TreatmentMatchJson or ClinicalGsonDeserializer. I will put back the GsonSerializer and use it instead.

jbartletthmf (approved):

PR #1300, opened 2025-08-20
ACTIN-2431: isMissingMolecularResultForEvaluation for Or() incorrectl…

cbruel:
…y merged


kduyvesteyn (commented):
You could maybe add a unit test for this, other than that it makes sense to me.

PR #1302, opened 2025-08-22
ACTIN-2421: Create CRC test data

cbruel:


kduyvesteyn (commented):
Haven't checked in detail (e.g. all lab values), some thoughts:
 - I'd go with PD-L1 <50%, think that's way more usual for a KRASpos CRC sample
 - Maybe remove the oncological history? Initial use case is for treatment-naive. Or maybe include a primary surgery for example? 
 - In the other samples, the `evidenceSource` is set to "External", no need to refer to CKB explicitly in test data (there is no CKB-dependency anywhere)

kduyvesteyn (approved):
Looks good to me!

I have patched the patient record json for the new molecular datamodel, and also corrected a typo in the treatment match (I think, pls check!). 

It would be nice to integrate things a bit further (make a runnable that uses the expected environment configuration, and add it to formal regression tests) but can be later (after report has been brushed up).

PR #1306, opened 2025-09-02
ACTIN-2402: Enhance treatment efficacy prediction report with SHAP 

beracira-hartwig:
[JIRA](https://hartwigmedical.atlassian.net/jira/software/projects/ACTIN/boards/45?jql=assignee%20%3D%20712020%3Ac6a1332f-01a6-48f0-9b52-ce4cdbff322b&selectedIssue=ACTIN-2402)

This is a rough sketch meant to see if everyone is okay with the approach. The idea is that the python service will also return the SHAP values together with prediction result in a json. And that json is parsed, then plotted in report generation. 

I already have the python component working as well. Will make a separate PR in the personalization repo. 

Sample plot generated from this code:
shap_plot


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/algo/TreatmentMatch.kt
```diff
@@ -10,6 +11,22 @@ data class TreatmentMatch(
...
     val maxMolecularTestAge: LocalDate? = null
 )
+
+data class TreatmentEfficacyPrediction(
+    @JsonProperty("survival_probs")
+    val survivalProbs: List<Double>,
+
+    @JsonProperty("shap_values")
+    val shapValues: Map<String, ShapDetail>
+) {
```
kduyvesteyn, line 24: What do these brackets do actually? Do they nest the data class or something? 

I don't think I'd do it this way out of consistency perspective (don't think its wrong though)

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/algo/TreatmentMatch.kt
```diff
@@ -10,6 +11,22 @@ data class TreatmentMatch(
     val trialMatches: List<TrialMatch>,
     val standardOfCareMatches: List<AnnotatedTreatmentMatch>?,
     val personalizedDataAnalysis: PersonalizedDataAnalysis?,
-    val survivalPredictionsPerTreatment: Map<String, List<Double>>?,
+    val survivalPredictionsPerTreatment: Map<String, TreatmentEfficacyPrediction>?,
     val maxMolecularTestAge: LocalDate? = null
 )
+
+data class TreatmentEfficacyPrediction(
```
kduyvesteyn, line 18: Common approach in ACTIN datamodel is to put every class in its own file. Would suggest to keep to that standard.

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/algo/TreatmentMatch.kt
```diff
@@ -10,6 +11,22 @@ data class TreatmentMatch(
...
+    @JsonProperty("survival_probs")
```
kduyvesteyn, line 19: This is tricky... I don't think we should expose non-standard json formatting downstream. This data class should be clean, but when we read it in from run_prediction we should map it (only there). Hope this makes sense!

common/src/test/resources/personalization/treatment_efficacy_prediction.json
kduyvesteyn, line 1: Auto-formatting should improve formatting of this file, can you check?

datamodel/src/main/kotlin/com/hartwig/actin/datamodel/algo/TreatmentMatch.kt
```diff
@@ -10,6 +11,22 @@ data class TreatmentMatch(
     val trialMatches: List<TrialMatch>,
     val standardOfCareMatches: List<AnnotatedTreatmentMatch>?,
     val personalizedDataAnalysis: PersonalizedDataAnalysis?,
-    val survivalPredictionsPerTreatment: Map<String, List<Double>>?,
+    val survivalPredictionsPerTreatment: Map<String, TreatmentEfficacyPrediction>?,
```
kduyvesteyn, line 14: I like the name of the class `TreatmentEfficacyPrediction`! Would also update the variable though (`efficacyPredictionPerTreatment`)

report/src/main/kotlin/com/hartwig/actin/report/pdf/chapters/PersonalizedEvidenceChapter.kt
```diff
@@ -117,9 +123,44 @@ class PersonalizedEvidenceChapter(private val report: Report, override val inclu
...
+                ggtitle("SHAP values for treatment: $treatmentName") +
+                ggsize(width = 1500, height = 800)
+
+        val tmpFile = createTempFile("shap_plot", ".svg")
+        ggsave(plot, tmpFile.absolutePathString())
+        val xObj = SvgConverter.convertToXObject(ByteArrayInputStream(tmpFile.readBytes()), document.pdfDocument)
+        return Image(xObj)
+    }
+
     private fun addSurvivalTable(document: Document) {
```
kduyvesteyn, line 157: The name of this function is no longer representative

kduyvesteyn (commented):
I like it so far!

On the graph:
 - I wouldn't go for the gradient colouring since the bar length is already indicative anyways (feels duplicated)
 - The feature values still feel awkward to me, they don't mean a lot.

As for datamodel, one thought that may be worthwhile is to add the treatment to `TreatmentEfficacyPrediction` so that you can make it a List instead of a Map in `TreatmentMatch` (consistent with e.g. `AnnotatedTreatmentMatch`

PR #1309, opened 2025-09-09
ACTIN-2504: ResistanceEvidenceMatcher performance improvement

kzuberihmf:
This addresses the sporadic timeouts on algo jobs. ResistanceEvidencematcher recomputes actionability matches in an inner loop, for the problematic case I estimate this was being rerun close to 10,000 times.

I've hoisted this computation out to class construction. Running the problematic case on ops vm the time goes from 22 mins to less than 30 seconds, with the treatment match json's being identical.

Note that we will now pay a bit extra at construction time vs if the matcher was never used, by i measured this to just a couple seconds, so don't think it's worth making this lazy.


kzuberihmf:
> 
> How did you narrow it down to this code?
> 

Once I was able to run the problematic case in ops, I hit it with 'kill -3' a few times to generate thread dumps to the console. They all showed execution stacks in the actionability matcher within resistance evidence matcher, so that was the main clue. Then did some remote debugging with break points set in that area to narrow it down a bit more.

Guess I could've profiled but not familiar with that tooling 😄.

> I'm curious if the `ActionabilityMatcher` could be made more efficient, but it's probably not worth the effort at this time.

Yeah good point, not so familiar with that code to want to attempt larger changes now either.

PR #1311, opened 2025-09-15
ACTIN-1494: Isolate environment configuration, clean up clinical configuration

kduyvesteyn:
Changes
 - `EnvironmentConfiguration` no longer exposed globally (reporter now only knows about report config etc)
 - Remove `ClinicalConfiguration`, an artefact from pre-feed-era.
 - Move `requestingHospital` into report configuration and rename to `hospitalOfReference`

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareApplication.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareApplication.kt
@@ -5,7 +5,7 @@ import com.hartwig.actin.PatientRecordJson
 import com.hartwig.actin.TreatmentDatabaseFactory
 import com.hartwig.actin.algo.calendar.ReferenceDateProviderFactory
 import com.hartwig.actin.algo.evaluation.RuleMappingResources
-import com.hartwig.actin.configuration.EnvironmentConfiguration
+import com.hartwig.actin.configuration.AlgoConfiguration
 import com.hartwig.actin.doid.DoidModel
 import com.hartwig.actin.doid.DoidModelFactory
 import com.hartwig.actin.doid.datamodel.DoidEntry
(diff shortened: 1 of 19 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareApplication.kt
```diff
@@ -53,8 +53,7 @@ class StandardOfCareApplication(private val config: StandardOfCareConfig) {
         val functionInputResolver = FunctionInputResolver(
             doidModel, icdModel, MolecularInputChecker.createAnyGeneValid(), treatmentDatabase, MedicationCategories.create(atcTree)
         )
-        val configuration = EnvironmentConfiguration.create(config.overridesYaml).algo
-        LOGGER.info(" Loaded algo config: $configuration")
+        val configuration = AlgoConfiguration.create(config.overridesYaml)
```
jbartletthmf, line 56: The reason for passing in the entire configuration was to support system-wide settings. For example, if we wanted to toggle SOC reporting, we could set an option in one place and have it apply to algo and reporting. I don't know if we have a use case like that though.

report/src/main/kotlin/com/hartwig/actin/report/ReporterApplication.kt
```diff
@@ -25,12 +25,11 @@ class ReporterApplication(private val config: ReporterConfig) {
         LOGGER.info("Loading treatment match results from {}", config.treatmentMatchJson)
         val treatmentMatch = TreatmentMatchJson.read(config.treatmentMatchJson)
 
-        val envConfig = EnvironmentConfiguration.create(config.overrideYaml)
-        LOGGER.info(" Loaded config: $envConfig")
+        val config = ReportConfiguration.create(config.overrideYaml)
```
jbartletthmf, line 28: Better to rename this `val` to avoid the naming conflict.

jbartletthmf (approved):

PR #1315, opened 2025-09-23
ACTIN-2558: Implement more protect logics into the actin algo

LiekeSchoenmaker:
SCOPE: 
- Drop ‘TML low' & 'TML high’ 
- Drop ‘gene disruption’


kduyvesteyn:
@LiekeSchoenmaker @ninajacobs The disruption evidence matching was removed, but the `DisruptionEvidence` class was left in place. This should be removed too

PR #1319, opened 2025-09-25
ACTIN-2552: Indirect evidence matching for hotspots

kzuberihmf:


common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/ActionabilityMatcher.kt
```diff
@@ -21,25 +22,99 @@ import org.apache.logging.log4j.Logger
 
 typealias MatchesForActionable = Map<Actionable, ActionabilityMatch>
 
-class ActionabilityMatcher(private val evidences: List<EfficacyEvidence>, private val trials: List<ActionableTrial>) {
+class ActionabilityMatcher(
+    private val evidences: List<EfficacyEvidence>,
+    private val trials: List<ActionableTrial>,
+    private val indirectEvidenceMatcher: IndirectEvidenceMatcher = IndirectEvidenceMatcher.empty()
```
kduyvesteyn, line 28: Would remove the default

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+                .filterNot { it.associatedWithDrugResistance() == true }
+                .groupBy { it.toKey() }
+
+            groupedByPosition
+                .filter { (_, hotspots) -> hotspots.size > 1 }
+                .forEach { (key, hotspots) ->
+                    hotspots.drop(1).forEach { duplicate ->
+                        logger.warn("KnownHotspot with key $key already exists in map; skipping duplicate for known hotspot $duplicate.")
+                    }
+                }
```
kduyvesteyn, line 78: Does this filter anything in practice? I'd think that every hotspot key when filtered on CKB is unique?

I guess you still want this code since datamodel doesn't enforce this but just interested in the answer! (and maybe if it really isn't supposed to happen you could turn the WARN into an exception?)

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+                .mapNotNull { evidence -> matchableEvidence(evidence, knownHotspotsByPosition) }
+                .groupBy({ it.first }, { it.second })
+                .mapValues { (_, evidences) -> evidences.toCollection(HashSet()) }
+
+            return IndirectEvidenceMatcher(treatmentByGeneEffect)
+        }
+
+        fun empty(): IndirectEvidenceMatcher = IndirectEvidenceMatcher(emptyMap())
+
+        private fun mapKnownHotspotsByPosition(serveRecord: ServeRecord): Map<HotspotKey, KnownHotspot> {
```
kduyvesteyn, line 65: Maybe rename for clarity since you only extract hotspots associated with resistance? e.g. `mapKnownResistantHotspotsByPosition`?

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        private fun matchableEvidence(
+            evidence: EfficacyEvidence,
+            knownHotspotsByVariant: Map<HotspotKey, KnownHotspot>
+        ): Pair<TreatmentKey, EfficacyEvidence>? {
+            return evidence.molecularCriterium().hotspots().firstOrNull()
+                ?.takeIf { ApplicabilityFiltering.isApplicable(it) }
+                ?.variants()?.firstOrNull()
+                ?.let { variant -> knownHotspotsByVariant[variant.toKey()] }
```
kduyvesteyn, line 90: Could there be a scenario where the first variant of a hotspot molecular criterium is not a known hotspot but the next one is? I guess you want to filter out evidence in case _any_ of the variants is associated with resistance? 

Maybe wrap line 90-92 into a separate function `takeIf { !hasVariantAssociatedWithDrugResistance(variant) }`?

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.hartwig.actin.molecular.interpretation.GeneAlterationFactory
+import com.hartwig.serve.datamodel.Knowledgebase
+import com.hartwig.serve.datamodel.ServeRecord
+import com.hartwig.serve.datamodel.efficacy.EfficacyEvidence
+import com.hartwig.serve.datamodel.molecular.hotspot.KnownHotspot
+import com.hartwig.serve.datamodel.molecular.hotspot.VariantAnnotation
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+
+data class TreatmentKey(
```
kduyvesteyn, line 17: Would rename to `GeneEffectKey` or `HotspotEffectKey`, this class isn't really related to treatments?

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+            return evidence.molecularCriterium().hotspots().firstOrNull()
+                ?.takeIf { ApplicabilityFiltering.isApplicable(it) }
+                ?.variants()?.firstOrNull()
+                ?.let { variant -> knownHotspotsByVariant[variant.toKey()] }
+                ?.takeIf { hotspot -> hotspot.associatedWithDrugResistance() != true }
+                ?.takeIf { GenericInhibitorFiltering.isGenericInhibitor(evidence.treatment()) }
+                ?.let { knownHotspot ->
+                    val proteinEffect = GeneAlterationFactory.convertProteinEffect(knownHotspot.proteinEffect())
+                    TreatmentKey(knownHotspot.gene(), proteinEffect) to evidence
+                }
```
kduyvesteyn, line 96: We should only match on `GAIN_OF_FUNCTION` (can include "predicted") and `LOSS_OF_FUNCTION` (can include "predicted"). Maybe that filtering could be added here?

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/curation/GenericInhibitorFiltering.kt
kduyvesteyn, line 1: 2 thoughts on this:
 - I would prefer to use the actual `drugClass` in CKB rather than the `drugClass` of the treatment approach (the treatment approach is a level of abstraction I am not a huge fan of, too abstract). We don't expose the drug classes of CKB treatments yet via SERVE though.
 - I went through all the SERVE drug classes and I believe nearly all of them are generic. My proposal would be to always consider a drug to be generic _unless its drug class contains a word that is resolvable to a HGVS protein effect_. I am not a clinician though so it is a somewhat risky statement :) This would be great to discuss with MCGI on Friday.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    // TODO: Actually this returns indirect but also direct hits, caller needs to filter out direct hits.
+    //  better name?
+    fun findIndirectEvidence(variant: Variant): Set<EfficacyEvidence> {
+        return findIndirectEvidence(variant.gene, variant.proteinEffect)
+    }
+
+    private fun findIndirectEvidence(gene: String, proteinEffect: ProteinEffect): Set<EfficacyEvidence> {
+        return treatmentByGeneEffectForHotspots[TreatmentKey(gene, proteinEffect)] ?: emptySet()
+    }
```
kduyvesteyn, line 42: I think you have all the input here to filter out direct hits? 

You can remove all the efficacy evidences for which the hotspot molecular criterium `HotspotKey` matches with the input variant `HotspotKey`?

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/IndirectEvidenceMatcher.kt
```diff
@@ -0,0 +1,107 @@
...
+    private val treatmentByGeneEffectForHotspots: Map<TreatmentKey, Set<EfficacyEvidence>>
+) {
+
+    // TODO: Actually this returns indirect but also direct hits, caller needs to filter out direct hits.
+    //  better name?
+    fun findIndirectEvidence(variant: Variant): Set<EfficacyEvidence> {
+        return findIndirectEvidence(variant.gene, variant.proteinEffect)
+    }
+
+    private fun findIndirectEvidence(gene: String, proteinEffect: ProteinEffect): Set<EfficacyEvidence> {
```
kduyvesteyn, line 40: You effectively reuse `HotspotMatching` in this class by pre-filtering and then matching on the key that is used by `HotspotMatching` but may be nice to consider reusing it? When I read this code for the first time, my question was "does it actually apply the default hotspot matching logic?"

kzuberihmf: Tried changing back to iterative matching with HotspotMatching instead of using the Map, but was slow. I've stuck with the map but put the key into hotspot matching and also used it there, so maybe its clearer that the same concept of match is being used (though not really sure this is an improvement).

kzuberihmf:
Also incorporated @pauldwolfe's utility to produce a tsv with detailed evidences, for detailed algorithm analysis & development over the next while. Separate main so invoked manually.

interpretation/src/main/kotlin/com/hartwig/actin/treatment/EvidenceScoringModel.kt
```diff
@@ -21,14 +21,25 @@ data class EvidenceScore(
...
+    ): EvidenceScore {
         val cancerTypeApplicability = when (treatment.cancerTypeMatch.applicability) {
             CancerTypeMatchApplicability.SPECIFIC_TYPE -> TumorMatch.PATIENT
             CancerTypeMatchApplicability.ALL_TYPES -> TumorMatch.ALL
             CancerTypeMatchApplicability.OTHER_TYPE -> TumorMatch.ANY
         }
-        val exactVariant = if (treatment.molecularMatch.sourceEvidenceType.isCategoryEvent()) VariantMatch.CATEGORY else VariantMatch.EXACT
-        val scoringMatch = ScoringMatch(cancerTypeApplicability, exactVariant)
+        val isCategoryEvent = treatment.molecularMatch.sourceEvidenceType.isCategoryEvent()
+        check(!(isIndirect && isCategoryEvent)) {
```
pauldwolfe, line 34: Could this ever happen in production without a programming error? What should someone do if this error occurs?

interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
```diff
@@ -21,12 +24,27 @@ data class TreatmentRankResult(val treatment: String, val scores: List<EvidenceS
...
+data class TreatmentEvidenceWithTarget(val treatmentEvidence: TreatmentEvidence, val target: String, val isIndirect: Boolean)
 data class DuplicateEvidenceGrouping(val treatment: String, val gene: String?, val tumorMatch: TumorMatch, val benefit: Boolean)
 
-class TreatmentRankingModel(private val scoringModel: EvidenceScoringModel) {
+class TreatmentRankingModel(
+    private val scoringModel: EvidenceScoringModel,
+    private val includeIndirectTreatmentEvidence: Boolean = true
+) {
 
     fun rank(record: PatientRecord): TreatmentEvidenceRanking {
```
pauldwolfe, line 35: suggestion, we could also put this functionality into TreatmentEvidenceRanking so which could both return the ranking (ordered, aggregated) or the results (unaggregated) and avoid having to expose the computeRankResults.  Although perhaps this is only temporary while we still need to generate the CSVs.

interpretation/src/main/kotlin/com/hartwig/actin/treatment/EvidenceScoringModel.kt
```diff
@@ -21,14 +21,25 @@ data class EvidenceScore(
 
 class EvidenceScoringModel(val config: ScoringConfig) {
 
-    fun score(treatment: TreatmentEvidence): EvidenceScore {
+    fun score(
+        treatment: TreatmentEvidence,
```
pauldwolfe, line 25: perhaps the isIndirect should be a property of the MolecularMatchDetails in the TreatmentEvidence?

interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
```diff
@@ -21,12 +24,27 @@ data class TreatmentRankResult(val treatment: String, val scores: List<EvidenceS
...
 }
 
-data class TreatmentEvidenceWithTarget(val treatmentEvidence: TreatmentEvidence, val target: String)
+data class TreatmentEvidenceWithTarget(val treatmentEvidence: TreatmentEvidence, val target: String, val isIndirect: Boolean)
 data class DuplicateEvidenceGrouping(val treatment: String, val gene: String?, val tumorMatch: TumorMatch, val benefit: Boolean)
 
-class TreatmentRankingModel(private val scoringModel: EvidenceScoringModel) {
+class TreatmentRankingModel(
+    private val scoringModel: EvidenceScoringModel,
+    private val includeIndirectTreatmentEvidence: Boolean = true
```
pauldwolfe, line 32: we can probably get rid of the feature toggle sooner than later, it's an experimental feature anyway

interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
```diff
@@ -60,22 +81,34 @@ class TreatmentRankingModel(private val scoringModel: EvidenceScoringModel) {
         }
 
     private fun treatmentEvidencesWithTargets(actionables: Sequence<Actionable>) =
-        actionables.flatMap {
-            it.evidence.treatmentEvidence.map { t ->
+        actionables.flatMap { actionable ->
+            val directEvidences = actionable.evidence.treatmentEvidence.asSequence().map { treatmentEvidence ->
```
pauldwolfe, line 85: Also related to comment on MolecularMatchDetails, I think in the long run it'll be nice to not have to deal with two collections (although I see the defensive value for now). Would be a good follow up task to assess impact of going back to one treatment evidence collection with a flag indicating if its direct/indirect.

pauldwolfe (dismissed):
A few comments/ideas/questions, but I think overall it looks good enough to evaluate with MCGI team.  We could schedule a session with Cam and Jens as soon as they are available I think.

interpretation/src/main/kotlin/com/hartwig/actin/treatment/TreatmentRankingModel.kt
kzuberihmf: Can make it private and still have the tsv helper use it think.

kzuberihmf: Moving to algo entered as ACTIN-2663 for future work.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/ActionabilityMatcher.kt
```diff
@@ -21,25 +22,100 @@ import org.apache.logging.log4j.Logger
 
 typealias MatchesForActionable = Map<Actionable, ActionabilityMatch>
 
-class ActionabilityMatcher(private val evidences: List<EfficacyEvidence>, private val trials: List<ActionableTrial>) {
+class ActionabilityMatcher(
+    private val evidences: List<EfficacyEvidence>,
+    private val trials: List<ActionableTrial>,
+    private val indirectEvidenceMatcher: IndirectEvidenceMatcher
```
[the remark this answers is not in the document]
kzuberihmf: Refactored.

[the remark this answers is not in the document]
kzuberihmf: @kduyvesteyn do you know why TreatmentMatcherApplication creates ActionabilityMatcher without going through the factory? After the suggested refactor, the indirect evidence matcher now produces a pile of warnings about duplicate hotspots, due to not filtering by evidence source as the ActionabilityMatcherFactory does.

kzuberihmf:
@kduyvesteyn @pauldwolfe now that the separate collection for indirect evidence is gone and there is not special feature flag for including it, any worries about unintended affects on non-mcgi reports?

E.g. Not very familiar with reports, but if looks like if we would have the molecular details chapter is switched on, that "Best evidence ..."  column could pick up an indirect evidence (I think). Not sure if there could be other ways for indirect evidence to get out.

common/src/main/kotlin/com/hartwig/actin/molecular/evidence/actionability/ActionabilityMatcher.kt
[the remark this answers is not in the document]
kduyvesteyn: Not sure, would even consider it a bug that it doesn't. Just pushed the change, good catch

PR #1322, opened 2025-09-30
ACTIN-2392: Changes to support comorbidity curation

jbartletthmf:
I had to update `Toxicity` to fix `withDefaultDate` and decided to seal `Comorbidity` at that time for ease of use.


datamodel/src/main/kotlin/com/hartwig/actin/datamodel/clinical/Comorbidity.kt
```diff
@@ -15,4 +15,79 @@ interface Comorbidity : Displayable {
...
     fun withDefaultDate(date: LocalDate): Comorbidity
+}
+
+data class Complication(
+    override val name: String?,
+    override val icdCodes: Set<IcdCode>,
+    override val year: Int? = null,
+    override val month: Int? = null,
+): Comorbidity {
+    override val comorbidityClass = ComorbidityClass.COMPLICATION
```
andreia-hartwig, line 26: aren't we removing complications?

PR #1323, opened 2025-10-03
ACTIN-2617: Updated section title and show T-number associated with pathology reports

andreia-hartwig:


kduyvesteyn (commented):
I would probably just always show the tissue ID, seems a bit arbitrary to sometimes leave it out. 

Also, assume you tested it somewhere?

kduyvesteyn (approved):
For @ninajacobs to comment on the question then

andreia-hartwig:
PR updated to always show the tissue ID.

ninajacobs (approved):

PR #1326, opened 2025-10-07
ACTIN-2632: Oversight in undetermined vs fail?

ninajacobs:


kduyvesteyn (approved):
Definitely an oversight! Good catch, this makes these functions consistent with the main `IsHomologousRecombinationDeficient`.

The good news is @ninajacobs was responsible for the oversight (Jan 2025, ACTIN-1669) ;) 

The tests are somewhat confusing btw as they defend on `hrdVariant()` to return a variant with reportable == false, but outside of scope of this change.

ninajacobs:
The logics are in place since 2024.. I mostly shouldn't have assigned 'isMissing..' part to a FAIL then - but indeed I could've catched then that we should've outputted an UNDETERMINED! It's fixed :)

PR #1501, opened 2026-03-31
ACTIN-3414: Add new rule HAS_HAD_CHEMORADIOTHERAPY_OF_TYPE_X_CHEMOTHERAPY_AND_AT_LEAST_Y_CYCLES

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -17,23 +16,15 @@ class CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles
 ) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val latestStart = record.oncologicalHistory.map { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
+        val treatmentHistoryEntries = record.oncologicalHistory
+        val latestStart = treatmentHistoryEntries.map { LocalDate.of(it.startYear ?: 0, it.startMonth ?: 1, 1) }
```
spdeleeuw: What is the rationale behind setting year to 0 when unknown?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
```diff
@@ -0,0 +1,97 @@
...
+    @Test
+    fun `Should fail if there is only radiotherapy`() {
+        val matchingTreatment =
+            TreatmentHistoryEntry(treatments = setOf(radiotherapy), treatmentHistoryDetails = TreatmentHistoryDetails(cycles = minCycles))
+        val record = TreatmentTestFactory.withTreatmentHistory(listOf(matchingTreatment))
+        assertResultForPatient(EvaluationResult.FAIL, matchingType, record)
+    }
+
+    @Test
+    fun `Should fail for matching treatment with insufficient cycles`() {
```
spdeleeuw, line 37: I think we should consider warning if the no. of cycles is insufficient, since that's a difficult data point.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
cbruel: Not my code and logic, but looked into it and I am also not sure about this logic either, thanks for catching. This class will currently pass for a treatment history entry with chemotherapy and radiotherapy with correct type and cycles when both start and stop date are null, though that seems more like an undetermined to me?

@jbartletthmf do you maybe remember why this was implemented this way? I notice this was discussed in this PR https://github.com/hartwigmedical/actin/pull/1029. 

I would say `enoughCyclesAndOngoingTreatment` should return null when dates are missing?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
cbruel: I see your point, but we evaluate it like this in many other classes too (HasHadPDFollowingTreatmentWithCategoryOfTypesAndCyclesOrWeeks, HasHadSystemicFirstLineTreatmentWithoutPdAndWithCycles, HasHadCombinedTreatmentNamesWithCycles, HasHadTreatmentWithDrugAndCycles). So then we should change this logic in all classes (in a different ticket)?

Or do you mean it's a difficult data point only for chemoradiotherapy and therefore this warn is only needed for this class?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
spdeleeuw: I guess it could pass if the treatment was the only one in the history (then stop date null could be considered as ongoing), but in other scenarios I think it should be undetermined in that case. 
It's also weird to not know the dates but to know the cycles, but it's theoretically possible with our datamodel.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCyclesTest.kt
spdeleeuw: No I mean we cannot really trust that the cycle data is 100% complete and correct IMO, so I guess we should change that more broadly to warn's instead of fails indeed.

cbruel: I've created a ticket to fix this: https://hartwigmedical.atlassian.net/browse/ACTIN-3443

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/CurrentlyGetsChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
cbruel: As discussed I removed that `Currently..` class since it isn't used anymore

spdeleeuw (approved):
LGTM

PR #1503, opened 2026-04-08
ACTIN-2840: Fix mismatching driver event and inclusion event issue

cbruel:
**Issue:** In e.g. `AnyGeneFromSetIsOverexpressed`: when a patient has a FGF4 amp, and thus potential FGF4 overexpression, we store "Potential FGF4 overexpression" in `inclusionEvents`. However, in `InterpretedCohortSummarizer.driverIsActionable()` we compare the `driver.event`s with the `inclusionEvent`s. Therefore there is no match and "FGF4 amp" is missing in "Potential trial events, considered no high driver"

**Solution** I've changed the `inclusionEvents` (and `exclusionEvents`) from a set of strings to a set of `MolecularEvent` (https://github.com/hartwigmedical/actin-datamodel/pull/27), in which we can store the `originalEvent` (`driver.event`) to match against in `InterpretedCohortSummarizer.driverIsActionable()` ánd the string that we would like to display in the "Molecular" column.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/AnyGeneFromSetIsOverexpressed.kt
```diff
@@ -15,16 +16,21 @@ class AnyGeneFromSetIsOverexpressed(
...
-            val result = geneIsAmplified.evaluate(test, ihcTests).result
-            result == EvaluationResult.PASS || result == EvaluationResult.WARN
-        }.map { it.key }
+        val amplifiedGenesWithEvents = genesToAmplification.mapValues { (_, geneIsAmplified) ->
+            geneIsAmplified.evaluate(test, ihcTests)
+        }.filter { (_, eval) ->
+            eval.result == EvaluationResult.PASS || eval.result == EvaluationResult.WARN
+        }.mapValues { (_, eval) ->
+            eval.inclusionMolecularEvents
+        }
```
jbartletthmf: I think it's cleaner to start a new line for each function call. I would also consider `filterValues` and using the implicit `it` here:

```suggestion
        val amplifiedGenesWithEvents = genesToAmplification.mapValues { (_, geneIsAmplified) -> geneIsAmplified.evaluate(test, ihcTests) }
        .filterValues { it.result == EvaluationResult.PASS || it.result == EvaluationResult.WARN }
        .mapValues { it.value.inclusionMolecularEvents }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasActivatingMutation.kt
```diff
@@ -81,7 +84,11 @@ class GeneHasActivatingMutation(
                     "$gene activating mutation(s): $variantsString " +
                             "together with potentially activating mutation(s) " +
                             concat(potentiallyActivatingWarnings.map { (event, type) -> "$event (${type.description})$inKinaseDomainString" }),
-                    inclusionEvents = activatingVariants + potentiallyActivatingWarnings.map { (event, _) -> event }
+                    inclusionEvents = activatingVariantsMolecularEvents + potentiallyActivatingWarnings.map { (event, _) ->
+                        MolecularEvent(
+                            event
+                        )
```
jbartletthmf: could be one line

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
```diff
@@ -63,6 +64,8 @@ object EvaluationFactory {
         )
     }
 
+    fun toMolecularEvent(events: Set<String>) = events.map(::MolecularEvent).toSet()
```
jbartletthmf: I'm not sure this belongs here. I would consider making it a top-level extension function, perhaps in `MolecularEventUtil.kt`:

```suggestion
    fun Collection<String>.toMolecularEvents() = map(::MolecularEvent).toSet()
```
so callers can simply do `eventStrings.toMolecularEvents()`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasSufficientCopyNumber.kt
jbartletthmf, line 72: With the suggested extension function, this could become:

```suggestion
        val fullAmplificationWithUnknownCopyNumber = evaluatedCopyNumbers[CopyNumberEvaluation.FULL_AMP_WITH_UNKNOWN_COPY_NUMBER]?.toMolecularEvents()
```
so we don't need an additional step/val below.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
```diff
@@ -205,7 +205,10 @@ object TrialGeneratorFunctions {
 
             listOfNotNull(
                 cohortString,
-                concat(cohort.molecularInclusionEvents - commonEvents, commonEvents.isEmpty() && (!allEventsEmpty || hidePrefix)),
+                concat(
+                    cohort.molecularInclusionEvents.map { it.display() }.toSet() - commonEvents,
```
jbartletthmf, line 209: When working with events in a set, is it correct that two events with the same original text but different display strings are considered separate elements? Does that hold for different original events with the same display string? Those cases may be unlikely, but the existing logic would only consider two events equal when both strings match.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
```diff
@@ -2,11 +2,12 @@ package com.hartwig.actin.algo.evaluation
 
 import com.hartwig.actin.datamodel.algo.Evaluation
 import com.hartwig.actin.datamodel.algo.EvaluationResult
+import com.hartwig.actin.datamodel.algo.MolecularEvent
 import com.hartwig.actin.datamodel.algo.StaticMessage
 
 object EvaluationFactory {
 
-    fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<String> = emptySet()): Evaluation {
+    fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<MolecularEvent> = emptySet()): Evaluation {
```
jbartletthmf: Instead of changing every caller to map its event strings to objects, we could overload the `pass` function to handle both strings and events, e.g. adding a separate function:
```kotlin
    fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<String> = emptySet()): Evaluation {
        return pass(message, recoverable, toMolecularEvent(inclusionEvents))
    }
```
When trying this locally I had to add `emptySet<MolecularEvent>()` in the call from `recoverablePass` to avoid ambiguity with both `pass` functions having the same name.

I think this is a reasonable convenience to add to our factory and it would help us avoid a lot of set conversions throughout algo. The main argument against this would be if we decide that it's important to always explicitly provide the event objects.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
cbruel: In this function we are only looking at the `MolecularEvent.display()` (since we are determining what strings to display, and make sure we don't display the same string twice when it's a molecular event for multiple cohorts), so we only compare the `.display()` strings here?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
cbruel: Thanks for this suggestion, very helpful! :)

cbruel: Update: ran into a few problems but eventually got a successful build, curious what you think :)
- After adding the extra pass function, I ran into ambiguity issues for all the `pass(message)` calls (for all the non molecular rules we don't pass any inclusionEvents; so in this case it's not clear which of the two pass() functions we want to call/use) -> so I decided to add another function  `fun pass(message: String, recoverable: Boolean = false)` to solve this
- However, this led to this problem (as described by chat gpt): `JVM signature clashes due to overloaded pass functions with default parameters and only differing by generic type (Set<MolecularEvent> vs Set<String>). The JVM erases generic types, so both signatures look the same.` -> which I solved by adding the `@JvmName("..") annotations`. Don't know if these annotations are common use, but this was the cleanest way to solve it, otherwise I had to rename the pass functions, which is very unwanted.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
jbartletthmf: That's true! I thought this was a relevant place for my comment, but I was wrong.

That said, we commonly handle these events in a set of `MolecularEvent` objects, so it's important to note that some set elements could differ only by original event or by display. Should we consider e.g. excluding the display override from the hash and equals functions so that each set element is known to have a unique `event`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
```diff
@@ -64,7 +76,21 @@ object EvaluationFactory {
         )
     }
 
-    fun toMolecularEvent(events: Set<String>) = events.map(::MolecularEvent).toSet()
+    @JvmName("passWithStringEvents")
```
jbartletthmf: Should be `warnWithStringEvents`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
jbartletthmf: The annotations make sense. Another option that avoids changing the function name would be to change the parameter ordering like e.g.
```kotlin
fun pass(message: String, inclusionEvents: Set<MolecularEvent>, recoverable: Boolean = false): Evaluation

fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<String>): Evaluation
``` 
Now both event types can be handled by functions with the same name, but the order of parameters means they have distinct signatures.

jbartletthmf (approved):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
cbruel: @jbartletthmf I tried this, before trying the annotations, but didn't work, but now it does :) probably did something slightly different. But doesn't matter, it works and I like this solution more than the annotations, so changed it!

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
cbruel: Not sure if I fully understand your comment. But I don't think we should`"so that each set element is known to have a unique event"`, since if we could have a set of molecular events with both `MolecularEvent("FGF4 amp", "FGF4 amp")` and `MolecularEvent("FGF4 amp", "Potential FGF4 overexpression")` and don't want to throw out any of those?

jbartletthmf: That's ok; I just wanted to check. I ask because cases like your example could result in output like:
"There are 2 inclusion events: FGF4 amp, FGF4 amp"
if we trust the set count and naively map the elements to `it.event`. It could also affect other set operations:
```kotlin
val fgf4Amp = MolecularEvent("FGF4 amp", "FGF4 amp")
// ...
val eventsExcludingFgf4Amp = result.inclusionEvents - fgf4Amp  // still contains an "FGF4 amp" event
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
```diff
@@ -18,17 +17,16 @@ object EvaluationFactory {
...
-    @JvmName("passWithStringEvents")
     fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<String> = emptySet()): Evaluation {
-        return pass(message, recoverable, inclusionEvents.toMolecularEvents())
+        return pass(message, inclusionEvents.toMolecularEvents(), recoverable)
     }
 
     fun pass(message: String, recoverable: Boolean = false): Evaluation {
-        return pass(message, recoverable, emptySet<MolecularEvent>())
+        return pass(message, emptySet(), recoverable)
     }
```
jbartletthmf: I think you can drop this function since it's equivalent to calling `fun pass(message: String, recoverable: Boolean = false, inclusionEvents: Set<String> = emptySet()): Evaluation` with the default empty parameter for events.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
```diff
@@ -18,17 +17,16 @@ object EvaluationFactory {
...
 
     fun recoverablePass(message: String): Evaluation {
-        return pass(message, true, emptySet<MolecularEvent>())
+        return pass(message, emptySet(),true)
```
jbartletthmf: Missing space, but you could also call the 2-parameter function for simplicity:
```suggestion
        return pass(message, true)
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationFactory.kt
jbartletthmf: Providing the type here is redundant.

jbartletthmf (approved):
Looks good. I suggested some minor simplifications that pass all tests locally.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
jbartletthmf: I went looking for examples in the code, but it seems that when we map a collection to `it.event`, we usually call `.toSet()` or `.distinct()` immediately after, so this may not be an issue.

cbruel: Thanks for checking! :)

PR #1504, opened 2026-04-10
Actin 2295: Replace scoreValue/scoreValuePrefix with scoreLowerBound/scoreUpperBound 

ayaahmedhmf:
## Notes
  - Each module's changes are in a separate commit to make reviewing easier but will all be squashed and merged together.
  - Build will fail until actin-datamodel v3.0.0 is deployed (see hartwigmedical/actin-datamodel#23).
  - This is the counterpart to hartwigmedical/actin-clinical#383 which covers the clinical/curation side.
  - Pending testing on test-framework

  ## Summary
  - Replace `scoreValue` and `scoreValuePrefix` with `scoreLowerBound` and `scoreUpperBound` in algo evaluation functions, database schema, and report formatting
  - Update `ValueComparison` with new bounds-based helpers (`evaluateBoundsVersusMinValue`, `evaluateBoundsVersusMaxValue`)
  - Update IHC classification, PD-L1 evaluation, and protein expression evaluation to use score bounds
  - Update database schema (`ihcTest` table) and `ClinicalDAO` to read/write new columns
  - Update `IhcTestInterpreter` to format score bounds as ranges in reports
  - Update example patient JSON fixtures
  - Bump actin-datamodel to 3.0.0


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
```diff
@@ -124,4 +127,10 @@ object PDL1EvaluationFunctions {
             else -> TestResult.UNKNOWN
         }
     }
+
+    private fun formatBounds(test: IhcTest) = when {
+        test.scoreLowerBound == test.scoreUpperBound -> "${test.scoreLowerBound}"
+        test.scoreLowerBound != null && test.scoreUpperBound != null -> "${test.scoreLowerBound}-${test.scoreUpperBound}"
+        else -> test.scoreLowerBound?.let { "> $it" } ?: test.scoreUpperBound?.let { "< $it" }
```
jbartletthmf: Should this be ">=" and "<="? I thought the range was inclusive.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/ProteinExpressionByIhcFunctions.kt
```diff
@@ -60,21 +60,23 @@ class ProteinExpressionByIhcFunctions(
...
-                evaluateVersusMinValue(scoreValue.roundToInt().toDouble(), ihcTest.scoreValuePrefix, referenceExpressionLevel.toDouble())
+                evaluateBoundsVersusMinValue(roundedScoreLowerBound?.toDouble(), roundedScoreUpperBound?.toDouble(), referenceExpressionLevel.toDouble())
             }
 
             IhcExpressionComparisonType.EXACT -> {
-                when (referenceExpressionLevel == scoreValue.roundToInt() && ihcTest.scoreValuePrefix.isNullOrEmpty()) {
+                when (roundedScoreLowerBound == roundedScoreUpperBound && roundedScoreLowerBound == referenceExpressionLevel) {
                     true -> EvaluationResult.PASS
                     false -> EvaluationResult.FAIL
                 }
```
jbartletthmf: Could use `if` instead of `when`.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasHER2ExpressionByIhcTest.kt
```diff
@@ -274,15 +274,41 @@ class HasHER2ExpressionByIhcTest {
...
+        val evaluation = negativeFunction.evaluate(
+            MolecularTestFactory.withIhcTests(listOf(ihcTest(scoreLowerBound = 0.0)))
+        )
+        assertMolecularEvaluation(EvaluationResult.PASS, evaluation)
+    }
+
     private fun ihcTest(
-        scoreValue: Double? = null,
+        scoreLowerBound: Double? = null,
+        scoreUpperBound: Double? = scoreLowerBound,
```
jbartletthmf: Defaulting this to `scoreLowerBound` is somewhat surprising. What do you think about adding a `score` parameter that is the default for both bounds?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/BreastCancerReceptorsEvaluatorTest.kt
```diff
@@ -96,4 +96,54 @@ class BreastCancerReceptorsEvaluatorTest {
...
+        )
+        assertThat(evaluation).isEqualTo(BreastCancerReceptorEvaluation.NEGATIVE)
+    }
+
+    @Test
+    fun `Should return LOW when evaluating a breast cancer patient with low ER IHC result`() {
+        val evaluation = breastCancerReceptorsEvaluator.evaluate(
+            setOf(DoidConstants.BREAST_CANCER_DOID),
+            listOf(IhcTest("HER2", scoreLowerBound = 1.0, scoreUpperBound = 1.0, scoreValueUnit = "+")),
+            ReceptorType.HER2
```
jbartletthmf, line 115: Should be ER according to test name?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasBreastCancerWithPositiveReceptorOfTypeTest.kt
```diff
@@ -335,9 +395,9 @@ class HasBreastCancerWithPositiveReceptorOfTypeTest {
     }
 
     private fun createIhcTest(
-        item: String, scoreText: String = "Score", scoreValue: Double = 50.0, scoreValueUnit: String = "Unit"
+        item: String, scoreText: String = "Score", scoreLowerBound: Double = 50.0, scoreUpperBound: Double = scoreLowerBound, scoreValueUnit: String = "Unit"
```
jbartletthmf: Again, I think it's a bit surprising to set the upper bound to the lower bound by default.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasBreastCancerWithPositiveReceptorOfTypeTest.kt
```diff
@@ -335,9 +395,9 @@ class HasBreastCancerWithPositiveReceptorOfTypeTest {
...
     ) = IhcTest(
-        item = item, scoreText = scoreText, scoreValue = scoreValue,
+        item = item, scoreText = scoreText, scoreLowerBound = scoreLowerBound, scoreUpperBound = scoreUpperBound,
         scoreValueUnit = scoreValueUnit, impliesPotentialIndeterminateStatus = false
```
jbartletthmf: If all arguments can't fit on the same line, then each should have its own line.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTripleNegativeBreastCancerTest.kt
```diff
@@ -202,10 +202,62 @@ class HasTripleNegativeBreastCancerTest {
...
+            ), setOf(DoidConstants.BREAST_CANCER_DOID)
+        )
+        val evaluation = function.evaluate(patient)
+        assertEvaluation(EvaluationResult.UNDETERMINED, evaluation)
+        assertThat(evaluation.undeterminedMessagesStrings()).containsExactly("Undetermined if triple negative breast cancer")
+    }
+
     private fun createIhcTest(
-        item: String, scoreText: String = "Score", scoreValue: Double = 50.0, scoreValueUnit: String = "Unit"
+        item: String, scoreText: String = "Score", scoreLowerBound: Double = 50.0, scoreUpperBound: Double = scoreLowerBound, scoreValueUnit: String = "Unit"
```
jbartletthmf: I'll stop repeating myself. Maybe we could provide this functionality for all test classes with a shared `TestIhcTestFactory`?

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
```diff
@@ -43,10 +43,17 @@ class IhcTestInterpreter {
...
-        } ?: ""
+        val formattedLowerValue = valueTest.scoreLowerBound?.let(Formats::twoDigitNumber).orEmpty()
+        val formattedUpperValue = valueTest.scoreUpperBound?.let(Formats::twoDigitNumber).orEmpty()
+
+        val formattedScore = when {
+            formattedLowerValue.isNotEmpty() && formattedUpperValue.isNotEmpty() -> {
+                if (formattedLowerValue == formattedUpperValue) formattedLowerValue else "$formattedLowerValue-$formattedUpperValue"
+            }
+            formattedLowerValue.isNotEmpty() -> "> $formattedLowerValue"
+            formattedUpperValue.isNotEmpty() -> "< $formattedUpperValue"
```
jbartletthmf: Are the bounds included in the range? If not, then cases where `lower == upper` imply an empty range.

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
```diff
@@ -43,10 +43,17 @@ class IhcTestInterpreter {
     }
 
     private fun formatValueBasedIhcTest(valueTest: IhcTest): String {
-        return valueTest.scoreValue?.let {
-            listOfNotNull(
-                "Score", valueTest.measure, valueTest.scoreValuePrefix, Formats.twoDigitNumber(it) + valueTest.scoreValueUnit.orEmpty()
-            ).joinToString(" ")
-        } ?: ""
+        val formattedLowerValue = valueTest.scoreLowerBound?.let(Formats::twoDigitNumber).orEmpty()
+        val formattedUpperValue = valueTest.scoreUpperBound?.let(Formats::twoDigitNumber).orEmpty()
```
jbartletthmf: Why `orEmpty()` here? It seems like it would be easier to just do null checks below since we never use the empty string.

common/src/main/kotlin/com/hartwig/actin/algo/evaluation/util/ValueComparison.kt
```diff
@@ -30,6 +30,16 @@ object ValueComparison {
         }
     }
 
+    fun evaluateBoundsVersusMinValue(lowerBound: Double?, upperBound: Double?, minValue: Double): EvaluationResult =
+        lowerBound?.takeIf { it >= minValue }?.let { PASS }
+            ?: upperBound?.takeIf { it < minValue }?.let { FAIL }
```
jbartletthmf: This makes sense if the range of included values includes the bounds. If the bounds are excluded, then an upper bound equal to the min value should also lead to FAIL since the true value could not equal the min.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
ayaahmedhmf: Yes good catch, will update.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/molecular/HasHER2ExpressionByIhcTest.kt
ayaahmedhmf: Yes, good point, Will change it.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/BreastCancerReceptorsEvaluatorTest.kt
ayaahmedhmf: Good catch!

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasTripleNegativeBreastCancerTest.kt
ayaahmedhmf: Yea will probably be better

common/src/main/kotlin/com/hartwig/actin/algo/evaluation/util/ValueComparison.kt
ayaahmedhmf: Yes, the assumption I was working with is the result bounds are inclusive, so `upper bound = min value` means it could be min value or less so `undetermined`, but I see your point that if the max is min, we can consider it is as a failure, depends on labs meaning I guess? @spdeleeuw what do u think? which one makes more sense?

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ayaahmedhmf: I actually agree, will change

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ayaahmedhmf: My assumption is they are included, waiting for @spdeleeuw to confirm.

common/src/main/kotlin/com/hartwig/actin/algo/evaluation/util/ValueComparison.kt
spdeleeuw: If a result describes e.g. PD-L1 1-10%, the described bounds (1 and 10%) are included in the range of possibilities. So when the trial requests:
- PD-L1 <1 -> failure since, lower bound (1) is above 1.
- PD-L1 ≤1 -> undetermined since score range is crossing the boundry.
- PD-L1 >1 -> undetermined
- PD-L1 ≥1 -> pass
- PD-L1 >10 -> failure
- PD-L1 <10 -> undetermined, since range is crossing boundry
- PD-L1 ≤10 -> pass

Does this answer your question?

ayaahmedhmf: Yes, thanks!

Previous version used to pass comparator from `scoreValuePrefix` ([example](https://github.com/hartwigmedical/actin/blob/1042d0034309f9f30583f365925f4b4ccb2a6777/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt#L34-L39)), right now we don't have a prefix and use bounds which we assume are inclusive, so we should also use >= and <= for comparison. @jbartletthmf You are right "upper bound equal to the min value should also lead to FAIL" .  I'll fix, thanks for flagging!

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ayaahmedhmf: > lower == upper imply an empty range.

It means it is a single value and not a range, e.g PD-L1: 5%. 

but I need to fix <,> to <= and >=

jbartletthmf: That makes sense, I just wasn't sure if the range was inclusive.

common/src/main/kotlin/com/hartwig/actin/algo/evaluation/util/ValueComparison.kt
jbartletthmf: Since the range is inclusive, upper bound equal to the min value should be UNDETERMINED. E.g. for an included upper bound of 10%, if a trial has a min value of 10% (must be >= 10%), the result is undetermined since it's possible that the underlying value is actually 10% and satisfies the requirement. This is different that if the range were exclusive and the underlying value were known to be < 10%. Also, if the trial requires the measure to be > 10%, then its min value is not 10% (it's something just a little bigger than 10%).

ayaahmedhmf: Yes valid point, Do we know is the trial criteria inclusive or exclusive? 

My understanding is that the [existing evaluation](https://github.com/hartwigmedical/actin/pull/1504/changes/BASE..1b2d35d127c108d094d7b0ce9dae2d9bf1c4d3c8#diff-109164535cb923d61ec0dd2768a212009a1a666e4ee845344edb4e858a7e456eR19) treats trial rules as inclusive so a trial requiring > 10% wouldn't happen unless the caller explicitly passes a minValue greater than 10, but the bounds functions themselves don't handle that distinction. Is that correct @jbartletthmf ?

If that's the case, then no change is needed.

jbartletthmf: I don't know if the trial criteria is inclusive or exclusive.

That's my understanding as well so I agree that no change is needed.

spdeleeuw: The eligibility rule constants include the bounds (see https://github.com/hartwigmedical/actin/tree/master/algo): 
- `PD_L1_SCORE_OF_AT_LEAST_X`
- `PD_L1_SCORE_OF_AT_MOST_X`

The trial criterion itself can be both: e.g. can be "greater than or equal to 50%", or "less than 50%". 
The former would be curated using PD_L1_SCORE_OF_AT_LEAST_X[50], the latter as PD_L1_SCORE_OF_AT_MOST_X[49]. However, since we take the in-/exclusion into account during curation, in the code we can assume the boundary is included?

ayaahmedhmf: Thanks! this answers my question, we can consider the trial inclusive.

ayaahmedhmf:
## What changed
  - Add `isLowerBoundInclusive`/`isUpperBoundInclusive` columns to database schema and DAO
  - Update `ValueComparison` to accept nullable inclusivity flags (`null` treated as inclusive)
  - Pass flags through in `PDL1EvaluationFunctions`, `ProteinExpressionByIhcFunctions`, and `IhcTestEvaluation`
  - Update `IhcTestInterpreter` to display `>` vs `>=` and `<` vs `<=` based on flags
  - Update `formatBounds` in PDL1 evaluation for correct operator display
  - Add test coverage for exclusive bounds in `ValueComparisonTest` and `IhcTestInterpreterTest`

Screenshot 2026-04-15 at 02 45 09

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/IhcTestEvaluation.kt
```diff
@@ -49,13 +49,13 @@ class IhcTestEvaluation(val filteredTests: Set<IhcTest>) {
             it.scoreText?.lowercase() in IhcTestEvaluationConstants.WILD_TYPE_TERMS && !it.impliesPotentialIndeterminateStatus
         }
 
-    private fun testValueAboveZero(ihcTest: IhcTest) = ihcTest.scoreValue?.let { scoreValue ->
-        ValueComparison.evaluateVersusMinValue(scoreValue, ihcTest.scoreValuePrefix, 0.0)
-    } == EvaluationResult.PASS
+    private fun testValueAboveZero(ihcTest: IhcTest) =
+        ValueComparison.evaluateBoundsVersusMinValue(ihcTest.scoreLowerBound, ihcTest.scoreUpperBound, 0.0,
+            ihcTest.isUpperBoundInclusive) == EvaluationResult.PASS
```
jbartletthmf: All parameters should either go on the same line, or one on each line. Here, I think you could do:

```suggestion
        ValueComparison.evaluateBoundsVersusMinValue(
            ihcTest.scoreLowerBound, ihcTest.scoreUpperBound, 0.0, ihcTest.isUpperBoundInclusive
        ) == EvaluationResult.PASS
```
This applies in other places too.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
```diff
@@ -124,4 +129,11 @@ object PDL1EvaluationFunctions {
             else -> TestResult.UNKNOWN
         }
     }
-}
\ No newline at end of file
+
+    private fun formatBounds(test: IhcTest) = when {
+        test.scoreLowerBound == test.scoreUpperBound -> "${test.scoreLowerBound}"
```
jbartletthmf: Is this the correct response when both are null? Is that assumed to never happen here?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
```diff
@@ -124,4 +129,11 @@ object PDL1EvaluationFunctions {
...
+        test.scoreLowerBound != null && test.scoreUpperBound != null -> "${test.scoreLowerBound}-${test.scoreUpperBound}"
+        else -> test.scoreLowerBound?.let { "${if (test.isLowerBoundInclusive ?: true) ">=" else ">"} $it" }
+            ?: test.scoreUpperBound?.let { "${if (test.isUpperBoundInclusive ?: true) "<=" else "<"} $it" }
```
jbartletthmf: I think this is easier to read as:
```suggestion
        test.scoreLowerBound != null -> ">${if (test.isLowerBoundInclusive == false) "" else "="} $it"
        else -> ">${if (test.isUpperBoundInclusive == false) "" else "="} $it"
```
I omitted `test.scoreUpperBound != null` in the last condition since at this point we know that exactly one of the bounds is non-null.

common/src/test/kotlin/com/hartwig/actin/util/ValueComparisonTest.kt
```diff
@@ -84,6 +86,116 @@ class ValueComparisonTest {
...
+    fun `evaluateBoundsVersusMinValue with exclusive lower bound above min should pass`() {
+        assertThat(ValueComparison.evaluateBoundsVersusMinValue(3.0, null, 2.0, null))
+            .isEqualTo(EvaluationResult.PASS)
+    }
+
+    @Test
+    fun `evaluateBoundsVersusMinValue with exclusive upper bound equal to min should fail`() {
+        assertThat(ValueComparison.evaluateBoundsVersusMinValue(null, 2.0, 2.0, false))
+            .isEqualTo(EvaluationResult.FAIL)
+    }
```
jbartletthmf, line 179: Should there be a test like this for an inclusive upper bound that results in undetermined? Likewise for evaluations against a max value.

jbartletthmf (commented):
Please refrain from force-pushing to large branches in review as it makes it harder to see what's been updated since the previous review. I was able to isolate the differences locally with git.

A few small comments:

ayaahmedhmf:
> Please refrain from force-pushing to large branches in review as it makes it harder to see what's been updated since the previous review. I was able to isolate the differences locally with git.

Thanks for flagging @jbartletthmf , In this PR I added the new changes to new commits (https://github.com/hartwigmedical/actin/pull/1504/changes/3ac8d0ebe8255a0723fb268c15a620830a9848ed and https://github.com/hartwigmedical/actin/pull/1504/changes/4667cb0ba4104bb36264f30309a2dd2005e2efc2) 

The force push was to merge conflicts with master and update, thought I makes it easier for reviewer to review latest commits instead of merging commits. But it seems it wasn't easy so I'll avoid that for new PRs.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
ayaahmedhmf: Yes it is assumed it would never happen, because we reach this method from this branch only  `if (roundedLower != null || roundedUpper != null) {`

But I do think even though this is the case I should cover this case here for clarity or for future bugs.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/PDL1EvaluationFunctions.kt
ayaahmedhmf: Agree, will change

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/IhcTestEvaluation.kt
ayaahmedhmf: Okay noted

common/src/test/kotlin/com/hartwig/actin/util/ValueComparisonTest.kt
ayaahmedhmf: Good catch, will add them.

jbartletthmf (approved):
> The force push was to merge conflicts with master and update, thought I makes it easier for reviewer to review latest commits instead of merging commits. But it seems it wasn't easy so I'll avoid that for new PRs.

Thanks! 🙂

ayaahmedhmf:
## No changes, just merging with master.

PR #1507, opened 2026-04-17
ACTIN-1605: Implement measured creatinine clearance evaluation

ayaahmedhmf:
Implement `HAS_MEASURED_CREATININE_CLEARANCE_OF_AT_LEAST_X`


cbruel:
Hey @ayaahmedhmf, this is correct but the goal of this ticket was to (also) implement the formula I sent you on slack (using the CREATININE_24U, CREATININE and URINE_VOLUME_24H). In most cases the "calculated" value won't be in the lab results, and we have to calculate it ourselves based on these 3 lab results and the formula. If the CREATININE_CLEARANCE_24H is already available in the lab results, we could use that of course, but if not, we should calculate it (if the 3 values are available).

ayaahmedhmf:
> Hey @ayaahmedhmf, this is correct but the goal of this ticket was to (also) implement the formula I sent you on slack (using the CREATININE_24U, CREATININE and URINE_VOLUME_24H). In most cases the "calculated" value won't be in the lab results, and we have to calculate it ourselves based on these 3 lab results and the formula. If the CREATININE_CLEARANCE_24H is already available in the lab results, we could use that of course, but if not, we should calculate it (if the 3 values are available).

Thanks for clarifying! Will update.

To double check, If `CREATININE_CLEARANCE_24H` is not in the lab results, we can calculate it using one of two paths depending on available data:

1. If we have `CREATININE_24U` (mmol/day) + `CREATININE` (µmol/L) we **wont** need `URINE_VOLUME_24H` we can calculate it: (CREATININE_24U x 1,000,000) / (CREATININE x 24 x 60) 
My understanding is since `CREATININE_24U` is the total daily amount, the `URINE_VOLUME_24H` is already mathematically included in this value.

2. If we dont have `CREATININE_24U` and have `CREATININE_URINE` (mmol/L) + `URINE_VOLUME_24H` (mL) + `CREATININE` (µmol/L):
we calculate it using: (CREATININE_URINE * URINE_VOLUME_24H x 1,000) / (CREATININE x 24 x 60 )

Does this look correct @cbruel ?

cbruel:
@ayaahmedhmf 
1. Yes, if `CREATININE_24U` is measured in mmol/day, we won't need the urine volume anymore.
2. Yes CREATININE_24U / URINE_VOLUME_24H = CREATININE_URINE

(btw also be aware of the date of the lab values here, we should only use 3 values in this calculation if they are all from the same date, otherwise it's of course not the same measurement!)

ayaahmedhmf:
## Changes

When a direct `CREATININE_CLEARANCE_24H` measurement is absent, calculate creatinine clearance from urine using two fallback methods:
                                                                                                                                                                                
- **Daily total** (CREATININE_24U + serum CREATININE):                                                                                                                        
- **Urine concentration** (CREATININE_URINE + URINE_VOLUME_24H + serum CREATININE):                                                                                           
                                                                                                                                                                                
- Fallback priority: direct measurement → daily total → urine concentration.                                                                                                    
- If the direct measurement is present and fails, fallbacks are not attempted.                                                                                                  
- Inputs with a comparator (`<`, `>`) yield UNDETERMINED since the direction of the calculated result would be ambiguous.                                                                                                                                         
- new `MultiLabEvaluationFunction`:  implementing both calculation methods                                                                                                                                       
- `MultiLabMeasurementEvaluator`: new evaluator that resolves a set of lab measurements (same-date or independent) before delegating to a `MultiLabEvaluationFunction`                                                                                 
- `MultiLabEvaluationFunction`:  new interface analogous to `LabEvaluationFunction` for multi-measurement evaluation                                                                                                                                            
- `LaboratoryRuleMapper` — wires `HAS_MEASURED_CREATININE_CLEARANCE_OF_AT_LEAST_X` to the fallback chain via `hasSufficientMeasuredCreatinineClearanceCreator()`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapper.kt
```diff
@@ -370,6 +390,14 @@ class LaboratoryRuleMapper(resources: RuleMappingResources) : RuleMapper(resourc
         return LabMeasurementEvaluator(measurement, function, minValidLabDate(), minPassLabDate(), highestFirst)
     }
 
+    private fun createMultiLabEvaluator(
+        measurements: Set<LabMeasurement>,
+        function: MultiLabEvaluationFunction,
+        requireSameDate: Boolean = true,
```
cbruel: `requireSameDate` is always true? So the else block in `MultiLabMeasurementEvaluator` line 25 will never be reached? Did you implement this for future cases? :) Since I don't think there will be any (you should always only compare lab values that were measured on the same day)

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+class LaboratoryRuleMapperTest {
+
+    private val factory = EvaluationFunctionFactory.create(RuleMappingResourcesTestFactory.create())
+    private val today = LocalDate.now()
+
+    private fun evaluatorFor(rule: EligibilityRule, vararg params: Parameter<*>): EvaluationFunction =
+        factory.create(EligibilityFunction(rule = rule.name, parameters = params.toList()))
+
+    // HAS_MEASURED_CREATININE_CLEARANCE_OF_AT_LEAST_X
```
cbruel: Would remove this line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+private const val ML_PER_LITER = 1000.0
+
+internal class HasSufficientMeasuredCreatinineClearance(
+    private val minCreatinineClearance: Double,
+    private val method: MeasuredCreatinineClearanceMethod
+): MultiLabEvaluationFunction {
+
+    override fun evaluate(record: PatientRecord, labMeasurements: Map<LabMeasurement, LabValue>): Evaluation {
+        val creatinine = labMeasurements[LabMeasurement.CREATININE]
+            ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.CREATININE.display()} value available")
```
cbruel: Would clarify this message to something like: "Measured creatinine clearance undetermined since no ${LabMeasurement.CREATININE.display()} value available"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+
+        if (labMeasurements.values.any { it.comparator.isNotEmpty() && it.comparator != "=" }) {
+            return EvaluationFactory.recoverableUndetermined(
+                "Cannot calculate creatinine clearance: comparator present on input value(s) makes the result direction ambiguous"
```
cbruel: Would rename this to: "Measured creatinine clearance undetermined: comparator present on input value(s) makes the result direction ambiguous"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+            )
+        }
+
+        val creatinineClearance = when (method) {
+            MeasuredCreatinineClearanceMethod.DAILY_TOTAL -> {
+                val creatinine24U = labMeasurements[LabMeasurement.CREATININE_24U]
+                    ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.CREATININE_24U.display()} value available")
```
cbruel: Also here add "Measured creatinine clearance undetermined since .."

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+                creatinine24U.value * MICROMOLES_PER_MILLIMOLE * ML_PER_LITER / (creatinine.value * MINUTES_PER_DAY)
+            }
+
+            MeasuredCreatinineClearanceMethod.URINE_CONCENTRATION -> {
+                val creatinineUrine = labMeasurements[LabMeasurement.CREATININE_URINE]
+                    ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.CREATININE_URINE.display()} value available")
```
cbruel: Same here

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+                val urineVolume = labMeasurements[LabMeasurement.URINE_VOLUME_24H]
+                    ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.URINE_VOLUME_24H.display()} value available")
```
cbruel: Same here

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+        val creatinine = labMeasurements[LabMeasurement.CREATININE]
+            ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.CREATININE.display()} value available")
+
+        if (labMeasurements.values.any { it.comparator.isNotEmpty() && it.comparator != "=" }) {
+            return EvaluationFactory.recoverableUndetermined(
+                "Cannot calculate creatinine clearance: comparator present on input value(s) makes the result direction ambiguous"
+            )
+        }
+
+        val creatinineClearance = when (method) {
```
cbruel, line 38: I believe you don't check the unit of the 3 lab values you are using to perform these calculations? (See `HasSufficientLabValue` and `HasLimitedLabValue`). If URINE_VOLUME_24H is for example given in dL, your calculations are not correct anymore?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+        }
+
+        val unit = LabMeasurement.CREATININE_CLEARANCE_24H.defaultUnit
+        val labValueString = labValue(LabMeasurement.CREATININE_CLEARANCE_24H, creatinineClearance, unit)
+        val refString = "$minCreatinineClearance ${unit.display()}"
+
+        return when (evaluateVersusMinValue(creatinineClearance, null, minCreatinineClearance)) {
+            EvaluationResult.PASS -> EvaluationFactory.recoverablePass("Calculated $labValueString exceeds min of $refString")
+            EvaluationResult.FAIL -> EvaluationFactory.recoverableFail("Calculated $labValueString below min of $refString")
+            else -> EvaluationFactory.recoverableUndetermined("Calculated creatinine clearance undetermined")
```
cbruel: "Measured creatinine clearance undetermined"

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+                val creatinineUrine = labMeasurements[LabMeasurement.CREATININE_URINE]
+                    ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.CREATININE_URINE.display()} value available")
+                val urineVolume = labMeasurements[LabMeasurement.URINE_VOLUME_24H]
+                    ?: return EvaluationFactory.recoverableUndetermined("No ${LabMeasurement.URINE_VOLUME_24H.display()} value available")
+                creatinineUrine.value * urineVolume.value * MICROMOLES_PER_MILLIMOLE / (creatinine.value * MINUTES_PER_DAY)
+            }
+        }
+
+        val unit = LabMeasurement.CREATININE_CLEARANCE_24H.defaultUnit
+        val labValueString = labValue(LabMeasurement.CREATININE_CLEARANCE_24H, creatinineClearance, unit)
```
cbruel: I wouldn't use the display value of the LabMeasurement.CREATININE_CLEARANCE_24H, so would remove this val and below just use
```
EvaluationResult.PASS -> EvaluationFactory.recoverablePass("Measured creatinine clearance exceeds min of $refString")
            EvaluationResult.FAIL -> EvaluationFactory.recoverableFail("Measured creatinine clearance below min of $refString")
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.hartwig.actin.datamodel.algo.EvaluationResult
+import com.hartwig.actin.datamodel.clinical.LabMeasurement
+import com.hartwig.actin.datamodel.trial.DoubleParameter
+import com.hartwig.actin.datamodel.trial.EligibilityFunction
+import com.hartwig.actin.datamodel.trial.Parameter
+import com.hartwig.actin.trial.input.EligibilityRule
+import java.time.LocalDate
+import org.junit.jupiter.api.Test
+
+class LaboratoryRuleMapperTest {
```
cbruel: I do understand you created a test class for this, but don't think it's really necessary? This logic should all be tested already in other test classes like `MultiLabMeasurementEvaluatorTest`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapper.kt
ayaahmedhmf: You're right, I created it for cases we don't need synced dates, but if there's no realistic scenario where we'd want to compare lab values across different dates. I'll remove it.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
ayaahmedhmf: The tests here validate the multi-step fallback ordering: direct measurement → daily-total calculation → urine-concentration calculation, which is logic specific to the `LaboratoryRuleMapper` wiring and isn't covered by `HasSufficientMeasuredCreatinineClearanceTest` or `MultiLabMeasurementEvaluatorTest` individually. Without this, a change in fallback ordering would go undetected.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
ayaahmedhmf: Unit validation is handled upstream by `MultiLabMeasurementEvaluator` line # 60 
```
LabEvaluation.isValid(it, measurement, minValidDate) 
```                            
  MultiLabMeasurementEvaluator uses this to filter candidate values so any lab value whose unit doesn't match measurement.defaultUnit is excluded from the map passed to `HasSufficientMeasuredCreatinineClearance`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.hartwig.actin.algo.evaluation.util.Format
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.algo.EvaluationResult
+import com.hartwig.actin.datamodel.algo.StaticMessage
+import com.hartwig.actin.datamodel.clinical.LabMeasurement
+import com.hartwig.actin.datamodel.clinical.LabValue
+import java.time.LocalDate
+
+internal class MultiLabMeasurementEvaluator(
```
jbartletthmf: Could `LabMeasurementEvaluator` be merged in here? We could keep some special logic for the n=1 case, but it looks like there's some overlap and having multiple evaluation paths could lead to drift.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+            .reduceOrNull { accumulator, dates -> accumulator intersect dates }
+            ?.filter { !it.isBefore(minValidDate) }
+            ?.sortedDescending()
+
+        return sharedDates?.firstNotNullOfOrNull { date ->
+            measurements.associateWith { measurement ->
+                interpretation.valuesOnDate(measurement, date)
+                    .firstOrNull { LabEvaluation.isValid(it, measurement, minValidDate) }
+            }
+                .takeUnless { it.values.any { it == null } }
```
jbartletthmf: `it` is shadowed

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
```diff
@@ -1,14 +1,57 @@
...
+private const val MICROMOLES_PER_MILLIMOLE = 1000.0
+private const val ML_PER_LITER = 1000.0
+
+internal class HasSufficientMeasuredCreatinineClearance(
+    private val minCreatinineClearance: Double,
+    private val method: MeasuredCreatinineClearanceMethod
+): MultiLabEvaluationFunction {
+
+    override fun evaluate(record: PatientRecord, labMeasurements: Map<LabMeasurement, LabValue>): Evaluation {
+        val creatinine = labMeasurements[LabMeasurement.CREATININE]
```
jbartletthmf, line 23: Should we also validate that creatinine is nonzero since we divide by it below?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+                    .toSet()
+            )
+        } else evaluation
+    }
+
+    private fun findSameDateValues(interpretation: LabInterpretation): Map<LabMeasurement, LabValue>? {
+        val sharedDates = measurements
+            .map { measurement ->
+                interpretation.allValues(measurement)
+                    ?.sortedByDescending { it.date }
```
jbartletthmf: Unnecessary sort?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+            ?.sortedDescending()
+
+        return sharedDates?.firstNotNullOfOrNull { date ->
+            measurements.associateWith { measurement ->
+                interpretation.valuesOnDate(measurement, date)
+                    .firstOrNull { LabEvaluation.isValid(it, measurement, minValidDate) }
+            }
+                .takeUnless { it.values.any { it == null } }
+                ?.mapValues { it.value!! }
+        }
```
jbartletthmf: I think this would be more efficient if we validated and found dates in the same pass:
```suggestion
        val validByDate = measurements.associateWith { measurement ->
            interpretation.allValues(measurement)
                ?.filter { LabEvaluation.isValid(it, measurement, minValidDate) }
                ?.groupBy { it.date }
        }

        return validByDate.values.takeUnless { it.any(Map<*,*>?::isNullOrEmpty) }
            ?.map { it!!.keys }
            ?.reduceOrNull(Set<LocalDate>::intersect)
            ?.maxOrNull()
            ?.let { date ->
                measurements.associateWith { measurement -> validByDate[measurement]!![date]!!.first() }
            }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearanceTest.kt
```diff
@@ -0,0 +1,124 @@
...
+                comparatorOverrides = mapOf(LabMeasurement.CREATININE to "<")
+            )
+        )
+    }
+
+    private fun evaluateFormula(
+        method: MeasuredCreatinineClearanceMethod,
+        vararg pairs: Pair<LabMeasurement, Double>,
+        comparatorOverrides: Map<LabMeasurement, String> = emptyMap(),
+        minValue: Double = 60.0,
```
jbartletthmf: It would be better to make `60.0` a constant.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearanceTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `Should pass for URINE_CONCENTRATION when calculated clearance is above minimum`() {
+        assertEvaluation(EvaluationResult.PASS, evaluateFormula(MeasuredCreatinineClearanceMethod.URINE_CONCENTRATION,
+            LabMeasurement.CREATININE_URINE to 5.0, LabMeasurement.URINE_VOLUME_24H to 1500.0, LabMeasurement.CREATININE to 70.0))
+    }
+
+    @Test
+    fun `Should fail for URINE_CONCENTRATION when calculated clearance is below minimum`() {
```
jbartletthmf, line 36: I find it hard to review these tests because the names don't clearly state the precondition (vs. e.g. `when INPUT_X is N times greater than INPUT_Y`). That might not be feasible for this type of calculation though. I'll leave it to @cbruel to assess test correctness unless there's something specific for me to review.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private val factory = EvaluationFunctionFactory.create(RuleMappingResourcesTestFactory.create())
+    private val today = LocalDate.now()
+
+    private fun evaluatorFor(rule: EligibilityRule, vararg params: Parameter<*>): EvaluationFunction =
+        factory.create(EligibilityFunction(rule = rule.name, parameters = params.toList()))
```
jbartletthmf: Private functions conventionally go at the bottom of the class.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluatorTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        assertEvaluation(EvaluationResult.UNDETERMINED, evaluator().evaluate(record))
+        verify { evaluationFunction wasNot called }
+    }
+
+    @Test
+    fun `Should degrade PASS to recoverable pass when same-date values are before minPassDate`() {
+        val oldDate = minPassDate.minusDays(1)
+        every { evaluationFunction.evaluate(any(), any()) } returns EvaluationFactory.recoverablePass("pass")
+        val record = recordWith(measurement1 to oldDate, measurement2 to oldDate)
+        assertEvaluation(EvaluationResult.PASS, evaluator().evaluate(record))
```
jbartletthmf: Should this test instead stub an unrecoverable PASS and assert that the evaluator makes it recoverable? It currently just asserts that the mocked PASS evaluation result is PASS.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluatorTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    private val recentDate = LocalDate.now()
+    private val minValidDate = recentDate.minusDays(90)
+    private val minPassDate = recentDate.minusDays(30)
+    private val measurement1 = LabMeasurement.entries[0]
+    private val measurement2 = LabMeasurement.entries[1]
+
+    private val evaluationFunction = mockk<MultiLabEvaluationFunction>()
+
+    @Test
+    fun `Should pass when all measurements are on the same date`() {
```
jbartletthmf: Maybe `Should propagate pass from function when all measurements are on the same date`? Otherwise it seems in conflict with the second test name.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
jbartletthmf: We don't typically write detailed tests for the RuleMappers, which may indicate a gap in our coverage, but we also try to avoid putting decision logic in those classes. Should the function returned by `hasSufficientMeasuredCreatinineClearanceCreator()` actually be in the implementation of `HasSufficientMeasuredCreatinineClearance`? We could introduce another class that evaluates for a specific `MeasuredCreatinineClearanceMethod` to keep the current separation of logic.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            findIndependentValues(interpretation)
+        } ?: return EvaluationFactory.recoverableUndetermined(
+            "No date found with all required lab values: ${measurements.joinToString { it.display() }}"
+        )
+
+        val evaluation = function.evaluate(record, selectedValues)
+
+        val oldestDate = selectedValues.values.minOf { it.date }
+        return if (evaluation.result == EvaluationResult.PASS && !oldestDate.isAfter(minPassDate)) {
```
jbartletthmf: Is `oldestDate` ever too old here? I thought those cases would be filtered out by the `findXValues` functions below.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
ayaahmedhmf: Good catch!
 `LabEvaluation.isValid()` only checks unit and date, not that the value is positive so a creatinine of 0 would yield a false PASS. I'll add a zero guard returning `UNDETERMINED`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
ayaahmedhmf: Thanks!  I'll update all the messages

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
ayaahmedhmf: Agreed, I'll remove it

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapper.kt
cbruel: Yes would remove it!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearance.kt
cbruel: Ah thanks for explaining!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
ayaahmedhmf: The functions below filter dates with **minValidDate** so `oldestDate` is always ≥
  `minValidDate`. But **minPassDate** is a stricter value that is valid (not stale) but old enough to warrant the "measurement occurred before minPassDate" on a PASS. So the check catches that middle band

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
ayaahmedhmf: Correct it is redundant.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
ayaahmedhmf: Will rename.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
ayaahmedhmf: Good suggestions, will apply

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasSufficientMeasuredCreatinineClearanceTest.kt
ayaahmedhmf: Agree, Looking at `HasSufficientDerivedCreatinineClearanceTest` I can add some inline comments showing the computed value (e.g. `// CrCl = x / (y * z) = 99 → above min of 60`). I understand that the codebase convention is to avoid comments, but I think this is the case where the "why" is non-obvious so comments will make the arithmetic traceable without bloating the test names.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
ayaahmedhmf: The pattern of using `Fallback` chains in the mapper is already established here (example at line 287) so the mapper acts as the composition root while the implementation classes stay focused on a single domain calculation. Moving the `Fallback` wiring into `HasSufficientMeasuredCreatinineClearance` would require it to know about `LabMeasurementEvaluator`, `MultiLabMeasurementEvaluator`, and date providers which belong in the mapper layer. 
But I do understand that you are trying to avoid exploding this class, so introducing a wrapper class here will help keeping the mapper simple, but would relocate those 5 line  into a new file without adding a new abstraction, testability benefit. I am happy to go either ways!

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluatorTest.kt
ayaahmedhmf: Good point

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluatorTest.kt
ayaahmedhmf: Good catch!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
ayaahmedhmf: Great point!                                                   
                                                                                                                                                                                
The overlap here is the  PASS block (checking `oldestDate` against `minPassDate`), duplicated identically in both classes. But the two evaluators have some differences                                                                                                                                                                    

One behavioural difference between the two classes is what happens when the most recent value for a measurement is invalid (wrong unit, too old)?         

- `LabMeasurementEvaluator`: takes `mostRecentValue()` and stops so if it's invalid, it returns an error ("Unexpected unit for creatinine", "Most recent measurement too old"). and doesn't fallback to older valid values.                                                                  
- `MultiLabMeasurementEvaluator`: is different, it iterates through shared dates and finds the most recent shared date where all measurements are simultaneously valid. The reason we do it is that 
it is very common that patient will have more serum creatinine measures (they sometimes take blood of a every day), then urine creatinine measures (which they probably only do once every few weeks)   (example by @cbruel )        

So I am thinking the merge could work by:                                                                                                                                               

1. Add a `stopAtMostRecent: Boolean = false` flag to `MultiLabMeasurementEvaluator`. When true: take the most recent value per measurement unconditionally, return an error for any invalid one rather than searching further. When false: current iterate-and-skip behaviour.                                                                                                            
2. LabMeasurementEvaluator becomes a thin delegation wrapper: stopAtMostRecent = true, passes through highestFirst.
                                                                                                                                                                                
The open question is what `stopAtMostRecent` should mean for n>1 and this is a separate question to @cbruel since you're closer to the clinical data.  Currently, for a single lab measurement (e.g. standard creatinine check), if the most recent value has the wrong unit , the system surfaces an error and stops, it does not fall back to an older valid value. I think the reason behind it is that the most recent reading is the authoritative one; if it's wrong, that's a data quality problem worth surfacing.                                                                                                                                              
                                                                                                                              
For the new measured creatinine clearance, we have multiple measurements that must be taken together (e.g. CREATININE_24U + CREATININE). Currently the system searches through time to find the most recent date where both were measured and both are valid.

So two possible policies for what to do when the most recent shared date has an invalid value (wrong unit on one of them):                                                       

1. keep searching: skip that date and look for the next shared date where all values are valid. Patient gets evaluated on older data with no warning about the bad recent reading.                                                                                                     
2. stop and surface: take the most recent shared date, and if any measurement on that date is invalid, return a specific error ("Unexpected unit for CREATININE_24U on 2026-04-10") rather than falling back.                                                                                                                                     
          
For example:                                                                                                                                                                  
  Date         CREATININE_24U    CREATININE                                                                                   
  2026-04-10   12 mmol/day ✓    85 mmol/L ✗ (wrong unit)
  2026-04-01   10 mmol/day ✓    70 µmol/mL ✓                                                                                                                                    
  - Option1 silently uses 2026-04-01, calculates clearance                                             
  - Option2 surfaces "Unexpected unit for creatinine on 2026-04-10" → UNDETERMINED                                                                                            
                                                                                                                                                                                
Which option better matches how you'd expect the system to behave?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapper.kt
ayaahmedhmf, line 371: @cbruel another question :) 

`LabMeasurementEvaluator` uses highestFirst to break ties when a patient has two creatinine values on the exact same date. For derived creatinine clearance the mapper passes `highestFirst = false`, meaning on a same-day tie it picks the lowest creatinine value, I assume because that gives the highest clearance estimate?                                                                                                                                                 

For the new measured creatinine clearance, if a patient has two CREATININE values on the same day (e.g. morning and afternoon draws), we can't tell which belongs to which 24h urine collection.Should we apply the same highestFirst = false logic (Per each measurment?)  for consistency but accepting that we still can't guarantee the paired values are from the same draw.

Is this same day duplicate sampling realistic for multiple measurements? If not, this is not an issue in practice and we can leave the current behaviour (no tiebreaker, picks first in list). 
If it is realistic, there's no clean solution at this level without time information in `LabValue`.

cbruel: @ayaahmedhmf thanks for the question. Same day duplicate sampling is not realistic for these measures, so agree it's safe to leave the current behaviour

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
jbartletthmf: I think it's right to ask @cbruel about this because I don't spend much time in this repo anymore.

On the topic of combining these classes, I'd just like to suggest for consideration the alternative of breaking down the two lab measurement evaluator classes into a shared class with the common logic (e.g. evaluating the selected measurement(s)) and two different classes for value selection that implement the specific handling. These classes could then be composed to provide the same behaviour with less duplication and potentially a simpler interface for testing.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LaboratoryRuleMapperTest.kt
jbartletthmf: A wrapper class would work just as well. My thoughts were: if this class is created by `hasSufficientMeasuredCreatinineClearanceCreator()` and mapped to the corresponding rule, shouldn't it be `HasSufficientMeasuredCreatinineClearance`? The existing `EvaluationFunction` only implements part of the logic that goes into this rule. The difference is just semantic so I'm happy with either approach.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/MultiLabMeasurementEvaluator.kt
cbruel: @ayaahmedhmf thanks for the question. I would keep it in line with the `LabMeasurementEvaluator` behaviour and go for option 2.

Haven't looked into detail into the code but I'm actually a bit surprised that wrong units lead to 'errors', since we could easily convert units (mmol/L -> micromol/mL), so wrong units shouldn't 'stop' us, but that's maybe work for another ticket? I know we have the `LabUnitConverter`, isn't that used in the `LabMeasurementEvaluator` class?

ayaahmedhmf: > Haven't looked into detail into the code but I'm actually a bit surprised that wrong units lead to 'errors', since we could easily convert units (mmol/L -> micromol/mL), so wrong units shouldn't 'stop' us, but that's maybe work for another ticket? I know we have the `LabUnitConverter`, isn't that used in the `LabMeasurementEvaluator` class?

You're right that `LabUnitConverter` exists and conversion could be done before rejecting wrong unit values. Currently isValid() filters them out. but we can attempt converting before rejecting. Fixing this properly means changing the selection layer which affects every lab rule, I can open a follow-up ticket for that.  

Also FYI: only 7 measurements have conversions defined: CREATININE, ALBUMIN, LYMPHOCYTES_ABS, HEMOGLOBIN, CALCIUM, TOTAL_BILIRUBIN, TESTOSTERONE. Everything else has no conversion factor and convert() returns null

ayaahmedhmf: > On the topic of combining these classes, I'd just like to suggest for consideration the alternative of breaking down the two lab measurement evaluator classes into a shared class with the common logic (e.g. evaluating the selected measurement(s)) and two dif

I like this approach as well

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabEvaluationFunction.kt
```diff
@@ -1,11 +1,16 @@
 package com.hartwig.actin.algo.evaluation.laboratory
 
-import com.hartwig.actin.datamodel.clinical.LabMeasurement
 import com.hartwig.actin.datamodel.PatientRecord
 import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.LabMeasurement
 import com.hartwig.actin.datamodel.clinical.LabValue
 
-interface LabEvaluationFunction {
+interface LabEvaluationFunction : MultiLabEvaluationFunction {
```
ayaahmedhmf: Ideally, `MultiLabEvaluationFunction` (the parent interface) should be the one named `LabEvaluationFunction`, but since that name is already taken, I’ve used MultiLabEvaluationFunction for now. Changing it now would affect 10+ files and make it harder for reviewers to spot the behavioral changes vs. the renaming. I can do the swap (and rename the old one to SingleLabValueEvFn) in the next commit so it’s easier to track.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/HasLimitedTotalBilirubinULNOrLimitedOtherMeasureULNDependingOnGilbertDisease.kt
```diff
@@ -29,7 +29,7 @@ class HasLimitedTotalBilirubinULNOrLimitedOtherMeasureULNDependingOnGilbertDisea
             LabMeasurement.TOTAL_BILIRUBIN to maxULNWithoutGilbertDisease
         }
 
-        return LabMeasurementEvaluator(applicableMeasure, HasLimitedLabValueULN(applicableULN), minValidLabDate, minPassLabDate).evaluate(
+        return LabMeasurementEvaluator(SingleLabValueSelector(applicableMeasure), HasLimitedLabValueULN(applicableULN), minValidLabDate, minPassLabDate).evaluate(
             record
         )
```
jbartletthmf: I think this would look a little cleaner with `evaluate` on a new line:
```suggestion
        return LabMeasurementEvaluator(SingleLabValueSelector(applicableMeasure), HasLimitedLabValueULN(applicableULN), minValidLabDate, minPassLabDate)
            .evaluate(record)
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabEvaluationFunction.kt
jbartletthmf: I think the swap is worthwhile 👍

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/SameDateLabValueSelector.kt
```diff
@@ -0,0 +1,36 @@
...
+                )
+            )
+
+        val selected = measurements.fold(mutableMapOf<LabMeasurement, LabValue>()) { acc, measurement ->
+            val value = interpretation.valuesOnDate(measurement, mostRecentSharedDate).firstOrNull()
+            if (!LabEvaluation.isValid(value, measurement, minValidDate)) {
+                return LabValueSelectionResult.NotFound(
+                    LabEvaluation.evaluateInvalidLabValue(measurement, value, minValidDate)
+                )
+            }
```
jbartletthmf, line 30: Is it right to fail if the first value on the most recent date is invalid, or would it be better to try the next value/previous date? `firstNotNullOfOrNull` could be used to test multiple values for validity.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/SameDateLabValueSelector.kt
```diff
@@ -0,0 +1,36 @@
...
+            acc.also { it[measurement] = value!! }
+        }
```
jbartletthmf: Folding with a mutable collection seems unnecessarily complex here. Aren't we just mapping `measurements` to the corresponding value? That could be done with `measurements.map { it to value }.toMap()` or more idiomatically:
```suggestion
        val selected = measurements.associateWith { measurement ->
            val value = interpretation.valuesOnDate(measurement, mostRecentSharedDate).firstOrNull()
            if (!LabEvaluation.isValid(value, measurement, minValidDate)) {
                return LabValueSelectionResult.NotFound(
                    LabEvaluation.evaluateInvalidLabValue(measurement, value, minValidDate)
                )
            }
            value!!
        }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/SameDateLabValueSelector.kt
ayaahmedhmf: The fail-fast behaviour here is based on the decision reached in this [thread](https://github.com/hartwigmedical/actin/pull/1507#discussion_r3172733062) (see Charlotte's comment), we treat the most recent shared date as authoritative. If the value on that date is invalid, surface the error rather than silently falling back to older data. This mirrors how `LabMeasurementEvaluator` handles single-lab invalid values.

jbartletthmf (approved):

PR #1510, opened 2026-04-20
ACTIN-3500: Add iText license key to ACTIN report

pauldwolfe:
Add dependencies and argument to the report config to optionally specify a license key.


ArnoHMF (approved):
Looks good to me.

PR #1511, opened 2026-04-21
ACTIN-3380: Remove unnecessary exclusions due to treatment-database

eljo-hartwig:
These exclusions were required due to the log4j dependencies of treatment-database, but are not required anymore as of this [PR](https://github.com/hartwigmedical/actin-treatment-database/pull/9)


nedleitch (approved):
LGTM, thanks Eljo!

PR #1513, opened 2026-04-24
ACTIN-3443: Warn instead of fail for insufficient cycles

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
```diff
@@ -19,23 +19,28 @@ class HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles(privat
...
+            treatmentMatches.any { (match, cycles) -> match == true && cycles == true } -> EvaluationFactory.pass(
                 "Had received chemoradiotherapy with $typeString chemotherapy and at least $minCycles cycles"
             )
 
-            null in treatmentMatches -> EvaluationFactory.undetermined(
+            treatmentMatches.any { (match, cycles) -> match == true && cycles == false } -> EvaluationFactory.warn(
+                "Had received chemoradiotherapy with $typeString chemotherapy but with less than $minCycles cycles"
+            )
+
+            treatmentMatches.any { (match, cycles) -> match == null || (match && cycles == null) } -> EvaluationFactory.undetermined(
```
spdeleeuw, line 40: (match && cycles == null) is never reached when match == null, so redundant?

spdeleeuw (approved):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadChemoradiotherapyWithSpecificChemotherapyTypeAndMinimumCycles.kt
cbruel: `(match && cycles == null)`  is `match == true && cycles == null` (since as you indicate match == null is already true for the first part, so match is always not null when this statement is evaluated) :)

PR #1519, opened 2026-05-01
ACTIN-3279: Review HasMinimumLanskyKarnofskyScore

cbruel:
Apart from a bug in one of the messages, I reviewed the logic in this class and it all works fine for AT_MOST and AT_LEAST WHO values, so not sure why this ticket was created? Does one of you remember?

https://hartwigmedical.atlassian.net/browse/ACTIN-3279?atlOrigin=eyJpIjoiNTRlY2U3OWNmZDEwNGJmOTg0NzdkZWMxNGI3MjUyYTEiLCJwIjoiaiJ9


ninajacobs:
Hmm I should have written it more clearly. I made this ticket after I added the WhoStatus.IsEqualTo / IsAtLeast / IsAtMost ranges of WhoStatusFunctions. The Krnofsky rule is currently the one one that still uses the asRange() function. I believe the ticket was about that the logics could be replaced by isAtLeast and isAtMost too, being allowed to remove asRange(). But will leave it up to you if it's worthwhile, it's probably not the most urgent thing if you say the logics are still correct!

ninajacobs (approved):
failing test ;)

cbruel:
@ninajacobs ah in my memory we still had a place in the code where we didn't properly evaluate the WHO <=2 inputs and thought that was this ticket, but guess that issue was already fixed then. Thanks for clarifying what this ticket is about, will have a look if we could use those functions

andreia-hartwig:
@cbruel I don't have any context on this ticket.

PR #1520, opened 2026-05-04
ACTIN-3233: Migrate from log4j to kotlin-logging

eljo-hartwig:
This PR migrates log4j to kotlin-logging for actin, this makes it consistent with actin-clinical which already uses kotlin-logging


algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareApplication.kt
```diff
@@ -92,7 +91,7 @@ fun main(args: Array<String>) {
         val config = StandardOfCareConfig.createConfig(DefaultParser().parse(options, args))
         StandardOfCareApplication(config).run()
     } catch (exception: ParseException) {
-        StandardOfCareApplication.LOGGER.warn(exception)
+        StandardOfCareApplication.logger.warn(exception) { exception.message ?: "" }
```
nedleitch: Aren't we just going to get the message twice?

algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareConfig.kt
```diff
@@ -5,10 +5,10 @@ import com.hartwig.actin.configuration.OVERRIDE_YAML_DESCRIPTION
 import com.hartwig.actin.util.ApplicationConfig
 import org.apache.commons.cli.CommandLine
 import org.apache.commons.cli.Options
-import org.apache.logging.log4j.Level
-import org.apache.logging.log4j.LogManager
-import org.apache.logging.log4j.Logger
-import org.apache.logging.log4j.core.config.Configurator
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.Logger
```
nedleitch: Would actually have expected to see a dependency on the SLF4J API here, not logback...

algo/src/main/kotlin/com/hartwig/actin/algo/TreatmentMatcherConfig.kt
```diff
@@ -85,13 +85,13 @@ data class TreatmentMatcherConfig(
 
         fun createConfig(cmd: CommandLine): TreatmentMatcherConfig {
             if (cmd.hasOption(LOG_DEBUG)) {
-                Configurator.setRootLevel(Level.DEBUG)
-                LOGGER.debug("Switched root level logging to DEBUG")
+                (LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME) as Logger).level = Level.DEBUG
+                logger.debug { "Switched root level logging to DEBUG" }
```
nedleitch: Wondering if we should make an (extension?) function for this to avoid duplicating it? It's a bit jarring.

common/src/main/kotlin/com/hartwig/actin/doid/serialization/DoidJson.kt
```diff
@@ -50,7 +50,7 @@ object DoidJson {
             graphsChecker.check(graph)
             val id: String = string(graph, "id")
             if (id == ID_TO_READ) {
-                LOGGER.debug(" Reading DOID entry with ID '{}'", id)
+                logger.debug { " Reading DOID entry with ID '$id'" }
```
nedleitch: Is the leading space intentional?

common/src/main/kotlin/com/hartwig/actin/util/DatamodelPrinter.kt
```diff
@@ -1,16 +1,15 @@
...
-import org.apache.logging.log4j.LogManager
-import org.apache.logging.log4j.Logger
+import io.github.oshai.kotlinlogging.KotlinLogging
 
 class DatamodelPrinter(private val indentation: Int) {
 
     fun print(line: String) {
-        LOGGER.info("{}{}", " ".repeat(indentation), line)
+        logger.info { "${" ".repeat(indentation)}$line" }
     }
```
nedleitch, line 9: Second time I've seen a similar block to this too, maybe a more-general function would be worthwhile?

nedleitch (commented):
Couple of small things here that could improve it even more but I won't hold up the merge for them.

algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareConfig.kt
nedleitch: Oh hang on, I'm remembering something about this, I think maybe the API doesn't allow setting, only the implementation does? Carry on if so...

algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareApplication.kt
eljo-hartwig: Yes most likely, I have removed these occurrences

algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareConfig.kt
eljo-hartwig: Yes this does seem to be the case, from my understanding in order to set a logging level with SLF4j it would have to be done either as a JVM flag or set as a system property before a logger is created

common/src/main/kotlin/com/hartwig/actin/doid/serialization/DoidJson.kt
eljo-hartwig: Leading space was already there, but good catch, I have removed it.

common/src/main/kotlin/com/hartwig/actin/util/DatamodelPrinter.kt
eljo-hartwig: I have created a function for indented debug logging, let me know if this is preferred. As this is the only indented info logging I have kept it as is.

algo/src/main/kotlin/com/hartwig/actin/algo/matchcomparison/EvaluationComparison.kt
```diff
@@ -47,9 +48,9 @@ object EvaluationComparison {
                 if (resultDifferences.isNotEmpty() || recoverableDifferences.isNotEmpty() || messageDifferences.isNotEmpty()
                     || functionDifferences.isNotEmpty()
                 ) {
-                    logDebug("Differences found for $criteriaId:", indent)
+                    logger.debugIndented("Differences found for $criteriaId:", indent)
```
nedleitch, line 51: Actually isn't this particular invocation just the same as `logger.debug`?

database/src/main/kotlin/com/hartwig/actin/database/algo/TreatmentMatchLoaderApplication.kt
```diff
@@ -37,7 +37,7 @@ fun main(args: Array<String>) {
     try {
         config = TreatmentMatchLoaderConfig.createConfig(DefaultParser().parse(options, args))
     } catch (exception: ParseException) {
-        TreatmentMatchLoaderApplication.logger.warn(exception) { exception.message ?: "" }
+        TreatmentMatchLoaderApplication.logger.warn { exception.toString() }
```
nedleitch: What I was expecting here actually was just `TreatmentMatchLoaderApplication.logger.warn(exception)`. Have you tested to see what you get out of this? I'm not sure you'll get the stacktrace will you?

nedleitch (changes requested):
I think these are nice improvements!

Annoying of me to bring it up now, but I forgot to mention that I think these should actually go in the `utils` module of `actin-tools` so we're not copy/pasting across projects. Do you think that is doable?

database/src/main/kotlin/com/hartwig/actin/database/algo/TreatmentMatchLoaderApplication.kt
eljo-hartwig: the syntax with round brackets is deprecated, so the one with curly braces is preferred, however when testing I don't think there is any difference between them or whether toString() is used. I have removed the toString().

algo/src/main/kotlin/com/hartwig/actin/algo/matchcomparison/EvaluationComparison.kt
eljo-hartwig: This indent can also be 2 when it is called in TreatmentMatchComparison in function determineCohortDifferences

nedleitch (approved):

PR #1521, opened 2026-05-05
ACTIN-2204: Warn on subclonal matches in GeneHasVariantInExonRangeOfType

ayaahmedhmf:
## Summary                                                                                                                                                                  
- Add `CANONICAL_REPORTABLE_SUBCLONAL` classification to `GeneHasVariantInExonRangeOfType`, mirroring `GeneHasVariantInCodon`                                                                                                        
- Subclonal matches (clonalLikelihood < 0.5) now emit WARN instead of  silently passing
- PASS branches (high-driver variant and high-driver exon skip) are guarded: they downgrade to WARN when a subclonal match coexists  

### Testing: 
Created a mocked patient with these cases: 
```                                                                                                                                                                                
- subclonal-only canonical match                                                                                                                                            
                                                                                                                                                                                 
- clonal HIGH driver coexisting with subclonal match                                                                                                                        
                                                                                                                                                                                
- exon skip fusion coexisting with subclonal canonical variant
                                                                                                                                
-  exon skip fusion coexisting with both subclonal and non-canonical              
``` 

|Before|After|
|-|-|
|<img width="724" height="425" alt="Screenshot 2026-05-05 at 15 45 34" src="https://github.com/user-attachments/assets/ed509c81-a7e6-463a-9751-5e2ca686136b" />|<img width="685" height="646" alt="Screenshot 2026-05-05 at 15 47 52" src="https://github.com/user-attachments/assets/f7100524-e2c7-4e18-999e-7a8b383c1c15" />|

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
@@ -22,11 +22,11 @@ class GeneHasVariantInExonRangeOfType(
 ) {
 
     private enum class VariantClassification {
-        CANONICAL_HIGH_DRIVER,
+        CANONICAL_REPORTABLE_HIGH_DRIVER,
         CANONICAL_REPORTABLE_SUBCLONAL,
         CANONICAL_REPORTABLE_NON_HIGH_DRIVER,
         CANONICAL_UNREPORTABLE,
-        REPORTABLE_OTHER,
+        NON_CANONICAL_REPORTABLE,
         NONE
     }
 
@@ -46,7 +46,7 @@ class GeneHasVariantInExonRangeOfType(
                         }
 
                         hasCanonicalEffectInExonRange && variant.isReportable && variant.driverLikelihood == DriverLikelihood.HIGH -> {
-                            VariantClassification.CANONICAL_HIGH_DRIVER
+                            VariantClassification.CANONICAL_REPORTABLE_HIGH_DRIVER
                         }
 
                         hasCanonicalEffectInExonRange && variant.isReportable -> {
@@ -58,14 +58,14 @@ class GeneHasVariantInExonRangeOfType(
                         }
 
                         variant.isReportable && variant.otherImpacts.any { hasEffectInExonRange(it.affectedExon, minExon, maxExon) } -> {
-                            VariantClassification.REPORTABLE_OTHER
+                            VariantClassification.NON_CANONICAL_REPORTABLE
                         }
 
                         else -> VariantClassification.NONE
                     }
                 }.mapValues { (_, variants) -> variants.map(Variant::event).toSet() }
-        val highDriverEvents = variantClassifications[VariantClassification.CANONICAL_HIGH_DRIVER]
-        val reportableOtherVariantMatches = variantClassifications[VariantClassification.REPORTABLE_OTHER]
+        val highDriverEvents = variantClassifications[VariantClassification.CANONICAL_REPORTABLE_HIGH_DRIVER]
+        val reportableOtherVariantMatches = variantClassifications[VariantClassification.NON_CANONICAL_REPORTABLE]
         val subclonalVariantMatches = variantClassifications[VariantClassification.CANONICAL_REPORTABLE_SUBCLONAL]
 
         val (reportableExonSkips, unreportableExonSkips) =
@@ -91,10 +91,7 @@ class GeneHasVariantInExonRangeOfType(
             }
 
             !highDriverEvents.isNullOrEmpty() -> {
-                val extensions = listOfNotNull(
-                    if (!reportableOtherVariantMatches.isNullOrEmpty()) "variant(s) in non-canonical transcript: ${concat(reportableOtherVariantMatches)}" else null,
-                    if (!subclonalVariantMatches.isNullOrEmpty()) "variant(s) in canonical transcript but subclonal likelihood of > ${percentage(1 - CLONAL_CUTOFF)}: ${concat(subclonalVariantMatches)}" else null
-                )
+                val extensions = buildWarnExtensions(reportableOtherVariantMatches, subclonalVariantMatches)
                 EvaluationFactory.warn(
                     "Variant(s) ${concat(highDriverEvents)} $baseMessage in canonical transcript together with ${concat(extensions)}",
                     inclusionEvents = highDriverEvents + reportableOtherVariantMatches.orEmpty() + subclonalVariantMatches.orEmpty()
@@ -102,10 +99,7 @@ class GeneHasVariantInExonRangeOfType(
             }
 
             highDriverExonSkipEvents.isNotEmpty() -> {
-                val extensions = listOfNotNull(
-                    if (!reportableOtherVariantMatches.isNullOrEmpty()) "variant(s) in non-canonical transcript: ${concat(reportableOtherVariantMatches)}" else null,
-                    if (!subclonalVariantMatches.isNullOrEmpty()) "variant(s) in canonical transcript but subclonal likelihood of > ${percentage(1 - CLONAL_CUTOFF)}: ${concat(subclonalVariantMatches)}" else null
-                )
+                val extensions = buildWarnExtensions(reportableOtherVariantMatches, subclonalVariantMatches)
                 EvaluationFactory.warn(
                     "Exon(s) skipped $baseMessage due to ${concat(highDriverExonSkipEvents)} together with ${concat(extensions)}",
                     inclusionEvents = highDriverExonSkipEvents + reportableOtherVariantMatches.orEmpty() + subclonalVariantMatches.orEmpty()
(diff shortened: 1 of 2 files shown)


ninajacobs (commented):
For some reason I can't leave comments to the PR, keep getting an error when trying to save the messages.

Nice, PR looks logically right, and thanks for making the images! 

May be good to ask someone additionally code-specific feedback, eg John? I wonder if now that we have so many distinct messages, it would be preferred to test the specific message as outcome rather than only PASS/WARN?

Some minor code suggestions:
- In the test may be good to add a private const val like SUBCLONAL_LIKELIHOOD and CLONAL_LIKELIHOOD and assign 0.3 and 0.6 respectively?
- "Should pass when reportable canonical high driver match has clonality above boundary" -> "Should pass when reportable canonical high driver clonal match"
- "Should warn when reportable canonical match is subclonal" -> Should warn when reportable canonical high driver match is subclonal
- ... similarly to above messages, would be good if you can review that the test names are phrased consistently, for clarity?

Two unrelated suggestions but to make the class clearer:
- CANONICAL_HIGH_DRIVER -> should we rename to CANONICAL_REPORTABLE_HIGH_DRIVER?
- REPORTABLE_OTHER -> should we rename to NON_CANONICAL_REPORTABLE?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -39,6 +41,10 @@ class GeneHasVariantInExonRangeOfType(
                 .groupBy { variant ->
                     val hasCanonicalEffectInExonRange = hasEffectInExonRange(variant.canonicalImpact.affectedExon, minExon, maxExon)
                     when {
+                        hasCanonicalEffectInExonRange && variant.isReportable && variant.clonalLikelihood?.let { it < CLONAL_CUTOFF } == true -> {
+                            VariantClassification.CANONICAL_REPORTABLE_SUBCLONAL
+                        }
```
jbartletthmf, line 46: Is it correct that subclonality is more important than any other classification, i.e. we want these variants in the subclonal collection over any other?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -72,30 +79,36 @@ class GeneHasVariantInExonRangeOfType(
         val highDriverExonSkipEvents = highDriverExonSkips.map { it.event }.toSet()
 
         return when {
-            !highDriverEvents.isNullOrEmpty() && reportableOtherVariantMatches.isNullOrEmpty() -> {
+            !highDriverEvents.isNullOrEmpty() && reportableOtherVariantMatches.isNullOrEmpty() && subclonalVariantMatches.isNullOrEmpty() -> {
```
jbartletthmf, line 82: If a clonal high-driver variant exists, shouldn't we still pass here regardless of whether there are non-canonical impacts or subclonal matches? A given variant can only go in one collection. I'd expect this condition to be:

```suggestion
            !highDriverEvents.isNullOrEmpty() -> {
```
but the condition previously excluded cases with high driver events and other reportable variants so I could be wrong.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -72,30 +79,36 @@ class GeneHasVariantInExonRangeOfType(
...
                 EvaluationFactory.pass(
                     "Variant(s) $baseMessage in canonical transcript",
                     inclusionEvents = highDriverEvents
                 )
             }
 
-            highDriverExonSkipEvents.isNotEmpty() && reportableOtherVariantMatches.isNullOrEmpty() -> {
+            highDriverExonSkipEvents.isNotEmpty() && reportableOtherVariantMatches.isNullOrEmpty() && subclonalVariantMatches.isNullOrEmpty() -> {
```
jbartletthmf, line 89: Why do we require `reportableOtherVariantMatches` and `subclonalVariantMatches` to be null or empty here?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -72,30 +79,36 @@ class GeneHasVariantInExonRangeOfType(
...
                 EvaluationFactory.pass("Exon(s) skipped $baseMessage", inclusionEvents = highDriverExonSkipEvents)
             }
 
             !highDriverEvents.isNullOrEmpty() -> {
+                val extensions = listOfNotNull(
+                    if (!reportableOtherVariantMatches.isNullOrEmpty()) "variant(s) in non-canonical transcript: ${concat(reportableOtherVariantMatches)}" else null,
```
jbartletthmf: Or:
```suggestion
                    reportableOtherVariantMatches?.ifEmpty { null }?.let { "variant(s) in non-canonical transcript: ${concat(it)}" },
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -72,30 +79,36 @@ class GeneHasVariantInExonRangeOfType(
...
+                    "Variant(s) ${concat(highDriverEvents)} $baseMessage in canonical transcript together with ${concat(extensions)}",
+                    inclusionEvents = highDriverEvents + reportableOtherVariantMatches.orEmpty() + subclonalVariantMatches.orEmpty()
                 )
             }
 
             highDriverExonSkipEvents.isNotEmpty() -> {
+                val extensions = listOfNotNull(
+                    if (!reportableOtherVariantMatches.isNullOrEmpty()) "variant(s) in non-canonical transcript: ${concat(reportableOtherVariantMatches)}" else null,
+                    if (!subclonalVariantMatches.isNullOrEmpty()) "variant(s) in canonical transcript but subclonal likelihood of > ${percentage(1 - CLONAL_CUTOFF)}: ${concat(subclonalVariantMatches)}" else null
+                )
```
jbartletthmf: This is done in two places and could be consolidated.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
```diff
@@ -181,6 +200,10 @@ class GeneHasVariantInExonRangeOfType(
             }
         }
     }
+
+    companion object {
+        private const val CLONAL_CUTOFF = 0.5
```
jbartletthmf, line 207: Could go at top of file before class declaration?

jbartletthmf (commented):
Based on Nina's feedback, I think a few of my comments are based on incorrect assumptions. It seems like the presence of a single subclonal variant is supposed to supersede all other potential classifications. If so, then the code looks correct.

Maybe I'm not the best person to review `actin` PRs 😛

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: Yes, that is based on my conversation with @ninajacobs,

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: Based on my understanding if they exist simultaneously, we can have a clonal **high-driver** alongside a separate **subclonal** variant, so downgrading to `WARN` to surface that.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: Same reasoning as the variant PASS branch, this is for cases where a high-driver exon skip coexists with a separate subclonal canonical variant.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: Good suggestion, will update.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: Agreed, will extract into a helper.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/GeneHasVariantInExonRangeOfType.kt
ayaahmedhmf: I was trying to be consistent with `GeneHasVariantInCodon` (and some other files) that use the same pattern, so keeping them consistent feels right. but I am fine with changing if you think otherwise

jbartletthmf (approved):

PR #1522, opened 2026-05-08
ACTIN-3580: Set isMissingMolecularResultForEvaluation = false for War…

spdeleeuw:
…nIf compound function


cbruel (approved):

cbruel (approved):
would add a test in `WarnIfTest` btw, so it's easier to trace back later why we have this behaviour :)

PR #1523, opened 2026-05-08
ACTIN-3586: Improve HasAtLeastCertainAge messaging

spdeleeuw:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasAtLeastCertainAge.kt
```diff
@@ -10,14 +10,14 @@ class HasAtLeastCertainAge(private val referenceYear: Int, private val minAge: I
     override fun evaluate(record: PatientRecord): Evaluation {
         val age = referenceYear - record.patient.birthYear
         return when {
-            age > minAge -> EvaluationFactory.pass("Patient is at least $minAge years old")
+            age > minAge -> EvaluationFactory.pass("Patient is older than $minAge")
 
             age == minAge -> EvaluationFactory.undetermined(
                 "Undetermined if patient with birth year ${record.patient.birthYear})"
                         + " is above $minAge years old"
```
cbruel: We could change this message as well for consistency with the other two messages: `"+ is older than $minAge"`?

cbruel (approved):

PR #1524, opened 2026-05-13
ACTIN-3607: Add clarification that no alterations were found in DNA f…

spdeleeuw:
…or undetermined message in AnyGeneFromSetIsOverexpressed

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/AnyGeneFromSetIsOverexpressed.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/AnyGeneFromSetIsOverexpressed.kt
@@ -1,12 +1,14 @@
 package com.hartwig.actin.algo.evaluation.molecular
 
 import com.hartwig.actin.algo.evaluation.EvaluationFactory
+import com.hartwig.actin.algo.evaluation.util.Format
 import com.hartwig.actin.algo.evaluation.util.Format.concat
 import com.hartwig.actin.datamodel.algo.Evaluation
 import com.hartwig.actin.datamodel.algo.EvaluationResult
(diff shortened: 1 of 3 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/AnyGeneFromSetIsOverexpressed.kt
```diff
@@ -29,8 +31,15 @@ class AnyGeneFromSetIsOverexpressed(
                 }.toSet()
             )
         } else {
+            val (genesTestedForAmplificationInDna, genesNotTestedInDna) =
```
cbruel: minor but would make names similar. so either something like: `genesTestedInDna, genesNotTestedInDna` or `genesTestedForAmplificationInDna, genesNotTestedForAmplificationInDna`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/molecular/AnyGeneFromSetIsOverexpressed.kt
```diff
@@ -29,8 +31,15 @@ class AnyGeneFromSetIsOverexpressed(
...
+                genes.partition { test.testsGene(it, specific(MolecularTestTarget.AMPLIFICATION, "Amplification of")) }
+            val dnaClarification = when {
+                genesTestedForAmplificationInDna.isEmpty() -> ""
+                genesNotTestedInDna.isEmpty() -> " (but no amplifications found in DNA)"
+                else -> " (no amplification in DNA for ${Format.concatWithCommaAndOr(genesTestedForAmplificationInDna)})"
```
cbruel: Would use `concat`, since genes should be concatenated with 'and' instead of 'or'

cbruel (approved):

PR #1527, opened 2026-05-13
ACTIN-3271: Add symptomatic CNS metastases rule

cbruel:
Not completely sure what you mean in your jira ticket description with `"If we transfer brain → CNS for known lesions, then we should also do this for active status."` ?


ninajacobs (approved):
Looks good!

I meant with that sentence that in actin-clinical we should check if 'hasSymptomaticBrainLesions' = true also leads to hasSymptomaticCnsLesions = true (i.e. we do the same with transferring 'symptomatic' from brain to CNS as for 'active' from brain to CNS)

PR #1529, opened 2026-05-18
ACTIN-3619: Make resolving atc codes case insensitive

eljo-hartwig:
PR to fix unresolved atc code due to case sensitivity.


spdeleeuw (approved):
Nice! I was planning on doing this later today and I'm glad you thought the same :)

PR #1530, opened 2026-05-22
ACTIN-3590: Normalize lab values to default unit at selection time

ayaahmedhmf:
### Changes

- Normalize lab values  to their measurement's `defaultUnit` at selection time (in `SingleLabValueSelector` and `SameDateLabValueSelector`), before being passed to evaluators
- Pass conversion notes (e.g. "albumin converted from: 4.0 g/dL to 40.0 g/L")  in evaluation messages for traceability

### Why
Following up on this previous [conversation](https://github.com/hartwigmedical/actin/pull/1507#discussion_r3173231506) about unit conversion in lab evaluation
- Previously we would return UNDETERMINED results when a patient's lab unit differs from the default unit eventhough a conversion path exists. 

Screenshot 2026-05-21 at 13 38 58

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
@@ -21,29 +21,21 @@ internal object LabUnitConverter {
         return labValue.value * conversionFactor
     }
 
-    data class NormalizedLabValue(val value: LabValue, val conversionNote: String? = null)
-
-    fun normalizeLabValue(measurement: LabMeasurement, labValue: LabValue): NormalizedLabValue? {
+    fun normalizeLabValue(measurement: LabMeasurement, labValue: LabValue): LabValue? {
         val targetUnit = measurement.defaultUnit
         val conversionFactor = LabUnitConversionTable.findConversionFactor(measurement, labValue.unit, targetUnit)
         return when {
-            labValue.unit == targetUnit -> NormalizedLabValue(labValue)
+            labValue.unit == targetUnit -> labValue
             conversionFactor == null -> {
                 logger.warn { "No conversion factor defined for ${measurement.display()} to go from '${labValue.unit.display()}' to '$targetUnit'" }
                 null
             }
-            else -> {
-                val converted = labValue.copy(
-                    value = labValue.value * conversionFactor,
-                    unit = targetUnit,
-                    refLimitLow = labValue.refLimitLow?.let { it * conversionFactor },
-                    refLimitUp = labValue.refLimitUp?.let { it * conversionFactor }
-                )
-                NormalizedLabValue(
-                    value = converted,
-                    conversionNote = "${measurement.display()} converted from: ${labValue.value} ${labValue.unit.display()} to ${converted.value} ${converted.unit.display()}"
-                )
-            }
+            else -> labValue.copy(
+                value = labValue.value * conversionFactor,
+                unit = targetUnit,
+                refLimitLow = labValue.refLimitLow?.let { it * conversionFactor },
+                refLimitUp = labValue.refLimitUp?.let { it * conversionFactor }
+            )
         }
     }
 }
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
@@ -20,4 +20,30 @@ internal object LabUnitConverter {
         }
         return labValue.value * conversionFactor
     }
+
+    data class NormalizedLabValue(val value: LabValue, val conversionNote: String? = null)
+
+    fun normalizeLabValue(measurement: LabMeasurement, labValue: LabValue): NormalizedLabValue? {
+        val targetUnit = measurement.defaultUnit
+        val conversionFactor = LabUnitConversionTable.findConversionFactor(measurement, labValue.unit, targetUnit)
+        return when {
+            labValue.unit == targetUnit -> NormalizedLabValue(labValue)
+            conversionFactor == null -> {
+                logger.warn { "No conversion factor defined for ${measurement.display()} to go from '${labValue.unit.display()}' to '$targetUnit'" }
+                null
+            }
+            else -> {
+                val converted = labValue.copy(
+                    value = labValue.value * conversionFactor,
+                    unit = targetUnit,
+                    refLimitLow = labValue.refLimitLow?.let { it * conversionFactor },
+                    refLimitUp = labValue.refLimitUp?.let { it * conversionFactor }
+                )
+                NormalizedLabValue(
+                    value = converted,
+                    conversionNote = "${measurement.display()} converted from: ${labValue.value} ${labValue.unit.display()} to ${converted.value} ${converted.unit.display()}"
+                )
+            }
+        }
+    }
 }
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabValueSelectionResult.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabValueSelectionResult.kt
@@ -13,11 +13,11 @@ internal fun normalizeAndValidate(
     measurement: LabMeasurement,
     labValue: LabValue?,
     minValidDate: LocalDate
-): LabUnitConverter.NormalizedLabValue? =
+): LabValue? =
     labValue?.let { LabUnitConverter.normalizeLabValue(measurement, it) }
-        ?.takeIf { LabEvaluation.isValid(it.value, measurement, minValidDate) }
+        ?.takeIf { LabEvaluation.isValid(it, measurement, minValidDate) }
 
 internal sealed class LabValueSelectionResult {
-    data class Found(val values: Map<LabMeasurement, LabValue>, val conversionNotes: List<String> = emptyList()) : LabValueSelectionResult()
+    data class Found(val values: Map<LabMeasurement, LabValue>) : LabValueSelectionResult()
     data class NotFound(val evaluation: Evaluation) : LabValueSelectionResult()
 }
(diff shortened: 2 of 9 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
```diff
@@ -20,4 +20,30 @@ internal object LabUnitConverter {
...
         return labValue.value * conversionFactor
     }
-}
\ No newline at end of file
+
+    data class NormalizedLabValue(val value: LabValue, val conversionNote: String? = null)
+
+    fun normalizeLabValue(measurement: LabMeasurement, labValue: LabValue): NormalizedLabValue? {
+        val targetUnit = measurement.defaultUnit
+        val conversionFactor = LabUnitConversionTable.findConversionFactor(measurement, labValue.unit, targetUnit)
```
jbartletthmf, line 26: Is it necessary to pass in the target unit here along with `measurement`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabValueSelectionResult.kt
```diff
@@ -9,7 +9,15 @@ internal interface LabValueSelector {
...
+    measurement: LabMeasurement,
+    labValue: LabValue?,
+    minValidDate: LocalDate
+): LabUnitConverter.NormalizedLabValue? =
+    labValue?.let { LabUnitConverter.normalizeLabValue(measurement, it) }
+        ?.takeIf { LabEvaluation.isValid(it.value, measurement, minValidDate) }
+
 internal sealed class LabValueSelectionResult {
-    data class Found(val values: Map<LabMeasurement, LabValue>) : LabValueSelectionResult()
+    data class Found(val values: Map<LabMeasurement, LabValue>, val conversionNotes: List<String> = emptyList()) : LabValueSelectionResult()
```
jbartletthmf: Should `values` be a `Map<LabMeasurement, NormalizedLabValue>` instead so each conversion note is kept with the relevant value?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/SameDateLabValueSelector.kt
```diff
@@ -21,16 +20,15 @@ internal class SameDateLabValueSelector(
...
-                return LabValueSelectionResult.NotFound(
-                    LabEvaluation.evaluateInvalidLabValue(measurement, value, minValidDate)
-                )
-            }
-            value!!
+            val normalized = normalizeAndValidate(measurement, value, minValidDate)
+                ?: return LabValueSelectionResult.NotFound(LabEvaluation.evaluateInvalidLabValue(measurement, value, minValidDate))
+            normalized.conversionNote?.let { conversionNotes.add(it) }
+            normalized.value
         }
```
jbartletthmf: This could be done without a mutable collection:

```kotlin
  val normalized = measurements.associateWith { measurement ->
      val value = interpretation.valuesOnDate(measurement, mostRecentSharedDate).firstOrNull()
      normalizeAndValidate(measurement, value, minValidDate)
          ?: return LabValueSelectionResult.NotFound(LabEvaluation.evaluateInvalidLabValue(measurement, value, minValidDate))
  }
  return LabValueSelectionResult.Found(
      values = normalized.mapValues { (_, n) -> n.value },
      conversionNotes = normalized.values.mapNotNull { it.conversionNote }
  )

```

jbartletthmf (approved):
Minor comments

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
```diff
@@ -20,4 +20,30 @@ internal object LabUnitConverter {
...
+            else -> {
+                val converted = labValue.copy(
+                    value = labValue.value * conversionFactor,
+                    unit = targetUnit,
+                    refLimitLow = labValue.refLimitLow?.let { it * conversionFactor },
+                    refLimitUp = labValue.refLimitUp?.let { it * conversionFactor }
+                )
+                NormalizedLabValue(
+                    value = converted,
+                    conversionNote = "${measurement.display()} converted from: ${labValue.value} ${labValue.unit.display()} to ${converted.value} ${converted.unit.display()}"
```
cbruel: I would remove the `to ${converted.value} ${converted.unit.display()}` part since this will already be displayed in the total message? The total message (as I look at the example picture you added) contains quite some duplicate information (mentioning hemoglobin twice and 8.2 mmol/L twice). 
Ideally the message should be “Hemoglobin 8.2 mmol/L (converted from 13.2 g/dL) below min of 9.0 mmol/L” (like it is for HasLimitedLabValue), but looking at your code I think this will be complex to change and this is probably why you chose to implement it this way?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
ayaahmedhmf: Yes as `findConversionFactor` seems like a general-purpose lookup, so it can't assume the destination is
always `defaultUnit`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
ayaahmedhmf: Yes I agree there are some duplicates but this is because the conversion note is produced at selection time ("this value was in g/dL and was converted to g/L"), while the evaluation message is produced at evaluation time ("value X is below min Y"). also we could have cases where we evaluate 3 lab values for example `albumin` is converted but `creatinine` is the one that fails so the eval message will only mention creatinine, so the conversion note is the only place albumin's name and destination unit appears.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabValueSelectionResult.kt
ayaahmedhmf: My concern  is that it would require every `LabEvaluationFunction` implementor to unwrap `.value` who have no ownership of the conversion concern. The current state keeps that isolated to `LabMeasurementEvaluator`, which is the only place that both has access to the selection result and needs to append the note. Callers trust whatever the selector passes and shouldn't be aware of whether it was converted.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/SameDateLabValueSelector.kt
ayaahmedhmf: Good point, agree.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/laboratory/LabUnitConverter.kt
cbruel: - I see your point, agree we can't remove the part in that case. But could we please fix the rounding of the decimal number in the conversion note? We shouldn't decimal numbers like 8.19192 on a clinical report :)

- But more importantly, looking at your code in `evaluationWithConversionNotes`, am I correct the conversion note is a separate `EvaluationMessage` in passMessages/warnMessages/failMessages/.. ? If this is the case (sorry if I'm wrong!) I don't really like this. These messages should always be shown one after the other on the report and because e.g. `failMessages` is a Set, this could maybe lead to issues where the messages are not shown one after the other? 
But apart from that, it's just not a separate message, it should be in one string. If there is no way to do this properly, I would suggest to remove the conversion note (sorry that I said to you before we should add one, but this was in the belief that the message would look similar to the ones we produce in `HasLimitedLabValue`)

ayaahmedhmf: Thanks for the follow-up!

  **Rounding** agree, 

  **Separate message**

You're right, it's a separate `EvaluationMessage` added to `passMessages`/`failMessages`/etc.
but Kotlin's [`Set`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/set-of.html) preserves insertion order so the evaluation message will always appear before the conversion note.

The "one string" is what I'd prefer too, and it works cleanly in `HasLimitedLabValue` because
  conversion and evaluation happen in the same place. Here they're decoupled bec conversion
  happens at selection time while evaluation happens later in the `LabEvaluationFunction`. 

Making this a one string is possible but too complex as it would require passing conversion notes into **every** `LabEvaluationFunction`  implementor so each can phrase it in context. That's 12 evaluators to update and new evaluators could silently drop notes so it seems fragile and can get out of sync easily. 

Also thinking about the multi-value case (e.g. creatinine + creatinine_24U), inlining could also be wrong? the evaluator produces "Measured creatinine clearance below min of 60 mL/min", but the conversion note might be for creatinine_24U so appending it to the clearance message implies a connection that isn't there.  Or are you suggesting we only add the note if the converted measurement is the one the evaluation message directly refers to, and drop it otherwise? if so then this as well requires each evaluator to know which of its inputs were converted and decide whether to include the note. Doable but non-trivial and relies on every current and future evaluator getting it right.
  
In this case I think listing all conversions that happened after the evaluation msg adds a value for clinical traceability, but I will trust your judgement here if you think it doesn't :). 
wdyt?

cbruel: Adding the note indeed adds traceability, but weighting all the pros and cons (thanks a lot for thinking about this thoroughly and providing me these pros and cons), I think it's better that we leave it out.

I take the responsibility, you can add it to the jira ticket ;)
Sorry for all the extra work you did to add this, but I think it's good that we took some time to evaluate what is best in the end here :)

ayaahmedhmf: Totally agree, it is a well spent work :) 

>I take the responsibility, you can add it to the jira ticket ;)

haha

cbruel (approved):
thanks for making these changes Aya :)

PR #1532, opened 2026-05-28
ACTIN-3641: Show warnings with isMissingMolecularResultForEvaluation first

cbruel:


report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
```diff
@@ -167,7 +168,7 @@ object TrialGeneratorFunctions {
     fun contentForTrialCohortList(
         cohortsForTrial: List<InterpretedCohort>,
         includeFeedback: Boolean,
-        feedbackFunction: (InterpretedCohort) -> Set<String>,
+        feedbackFunction: (InterpretedCohort) -> Set<StaticMessageWithIsMissingMolecularResultForEvaluation>,
```
jbartletthmf: It seems like we only need `isMissingMolecularResultForEvaluation` for this use case to sort the results. Instead of changing the type of all feedback functions, we could just pair the messages with the `isMissing` flag for this case, e.g.
```kotlin
findCommonMembersInCohorts(cohortsForTrial) { cohort ->
    feedbackFunction(cohort).map { it to it.isMissingMolecularResultForEvaluation }.toSet()
}
```
or, to ensure that we base uniqueness on messages rather than the molecular result flag, just provide `feedbackFunction` when finding the common/specific feedback and use a separate lookup for sorting (untested code):
```kotlin
val missingMolecularResultByMessage = cohortsForTrial.associate(feedbackFunction, InterpretedCohort::isMissingMolecularResultForEvaluation)
val sorted = messages.sortedWith(
    compareByDescending<String> { missingMolecularResultByMessage[it] ?: false }.thenBy { it }
)
```

jbartletthmf: The latter approach would also eliminate the need for `sorted.map { it.message }.toSet()` because you'd already have a unique list of messages.

cbruel: Thanks for reviewing this PR :) 

We can't use the `InterpretedCohort::isMissingMolecularResultForEvaluation`, since this is defined on the 'cohort level'. So if any of the evaluations for that cohort has `Evaluation.isMissingMolecularResultForEvaluation` = true, this will be set to true. So this `InterpretedCohort::isMissingMolecularResultForEvaluation` will be equal for all messages of that cohort, and the sorting doesn't work.

What we need to know here is on the 'message level' if the evaluation corresponding to that message has `isMissingMolecularResultForEvaluation = true`, because in that case we want to show that message first.

Hope I did understand your feedback correctly, otherwise please let me know :)

jbartletthmf: Understood: you need additional information linked to each message. Maybe we could make use of the `EvaluationMessage` interface here so we can more easily incorporate other message properties in the future? For example:
```kotlin
class MessageWithIsMissingMolecularResultForEvaluation(
    val message: String,
    val isMissingMolecularResultForEvaluation: Boolean
) : EvaluationMessage {

    override fun combineBy(): String {
        return message
    }

    override fun combine(other: EvaluationMessage): EvaluationMessage {
        return this
    }

    override fun toString() = message

    override fun equals(other: Any?) = (other as? EvaluationMessage)?.toString() == message

    override fun hashCode() = message.hashCode()
}
```
I overrode `equals` and `hashCode` to only consider the message when checking for equality. We could compare it with other `EvaluationMessage`s with something like:
```kotlin
        compareByDescending<EvaluationMessage> {
            (it as? MessageWithIsMissingMolecularResultForEvaluation)?.isMissingMolecularResultForEvaluation ?: false
        }
            .thenComparing(EvaluationMessage::toString)
```
This would allow us to use the more general interface, which isn't strictly necessary as long as all of these messages can be converted to your new class. Overriding `equals` and `hashCode` may help with using set operations for uniqueness.

cbruel: Thanks a lot for thinking along here!

I added the `MessageWithIsMissingMolecularResultForEvaluation`. As you commented on the actin-datamodel PR I added the class in the actin code instead of actin-datamodel (so you can ignore that actin-datamodel PR from now on, it's not used anymore, will remove when done)

interpretation/src/main/kotlin/com/hartwig/actin/report/interpretation/InterpretedCohortFactory.kt
```diff
@@ -133,17 +130,31 @@ object InterpretedCohortFactory {
...
+                    MessageWithIsMissingMolecularResultForEvaluation(
+                        failMessage.toString(),
+                        evaluation.isMissingMolecularResultForEvaluation
+                    )
+                }
+            }
             .toSet()
     }
+
+    private fun Collection<Any>.toMessageSet(evaluation: Evaluation): Set<MessageWithIsMissingMolecularResultForEvaluation> =
```
jbartletthmf: Should be `EvaluationMessage` instead of `Any`?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/trial/TrialGeneratorFunctions.kt
```diff
@@ -225,16 +230,32 @@ object TrialGeneratorFunctions {
...
 
+    private fun concatFeedback(
+        input: Set<MessageWithIsMissingMolecularResultForEvaluation>,
+        replaceEmptyWithNone: Boolean = true,
+        separator: String = Formats.COMMA_SEPARATOR
+    ): String {
+        val sorted = input.sortedWith(compareByDescending<EvaluationMessage> {
+            (it as? MessageWithIsMissingMolecularResultForEvaluation)?.isMissingMolecularResultForEvaluation ?: false
+        }.thenComparing(EvaluationMessage::toString))
+        val joinedString = sorted.map { it.toString() }.toSet().joinToString(separator)
```
jbartletthmf: `toSet` not needed?

interpretation/src/main/kotlin/com/hartwig/actin/report/interpretation/InterpretedCohortFactory.kt
```diff
@@ -134,13 +133,28 @@ object InterpretedCohortFactory {
...
+    private fun extractFails(evaluations: Map<Eligibility, Evaluation>): Set<MessageWithIsMissingMolecularResultForEvaluation> {
         return evaluations.values.filter { it.result == EvaluationResult.FAIL && !it.recoverable }
-            .flatMap(Evaluation::failMessagesWithIsMissing)
+            .flatMap { evaluation ->
+                evaluation.failMessages.map { failMessage ->
+                    MessageWithIsMissingMolecularResultForEvaluation(
+                        failMessage.toString(),
+                        evaluation.isMissingMolecularResultForEvaluation
+                    )
+                }
```
jbartletthmf: Could use `toMessageSet`?

interpretation/src/main/kotlin/com/hartwig/actin/report/interpretation/InterpretedCohortFactory.kt
```diff
@@ -133,17 +130,31 @@ object InterpretedCohortFactory {
...
             .toSet()
     }
+
+    private fun Collection<Any>.toMessageSet(evaluation: Evaluation): Set<MessageWithIsMissingMolecularResultForEvaluation> =
+        this.map {
+            MessageWithIsMissingMolecularResultForEvaluation(
+                it.toString(),
+                evaluation.isMissingMolecularResultForEvaluation
+            )
+        }.toSet()
```
jbartletthmf: How about this instead?
```suggestion
    private fun Evaluation.toMessageSet(
        getMessages: (Evaluation) -> Set<EvaluationMessage>
    ): Set<MessageWithIsMissingMolecularResultForEvaluation> = getMessages(this)
        .map { MessageWithIsMissingMolecularResultForEvaluation(it.toString(), isMissingMolecularResultForEvaluation) }
        .toSet()
```
Then you could call it with `evaluation.toMessageSet(Evaluation::failMessages)`.

jbartletthmf (approved):

PR #1536, opened 2026-06-03
ACTIN-3611: Propagate toxicity datamodel changes in actin repo

spdeleeuw:


common/src/test/kotlin/com/hartwig/actin/algo/evaluation/toxicity/ToxicityFunctionsTest.kt
```diff
@@ -18,7 +18,9 @@ class ToxicityFunctionsTest {
     private val ehrTox = Toxicity(
         name = "tox",
         icdCodes = setOf(IcdCode("code")),
-        evaluatedDate = referenceDate.minusMonths(2),
+        year = 2024,
```
andreia-hartwig: Perhaps you can use the referenceDate (referenceDate.year, referenceDate.month and referenceDate.dayOfMonth) here so you don't need to repeat the same numbers.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/clinical/ClinicalDataFunctions.kt
```diff
@@ -0,0 +1,12 @@
+package com.hartwig.actin.report.pdf.tables.clinical
+
+object ClinicalDataFunctions {
+
+    fun toDateString(maybeYear: Int?, maybeMonth: Int?, maybeDay: Int? = null): String? =
```
andreia-hartwig: Why maybeYear and not just year?

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/clinical/ClinicalDataFunctions.kt
```diff
@@ -0,0 +1,12 @@
...
+        when {
+            maybeYear == null -> null
+            maybeMonth == null -> "$maybeYear"
+            maybeDay == null -> "$maybeMonth/$maybeYear"
+            else -> "$maybeDay/$maybeMonth/$maybeYear"
```
andreia-hartwig: I would follow the ISO 8601 standard here and use format YYYY-MM-DD.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/clinical/ClinicalDataFunctions.kt
```diff
@@ -0,0 +1,12 @@
+package com.hartwig.actin.report.pdf.tables.clinical
+
+object ClinicalDataFunctions {
```
andreia-hartwig: I would drop Clinical from the name since the function is purely date based.

report/src/main/kotlin/com/hartwig/actin/report/pdf/tables/clinical/ClinicalDataFunctions.kt
spdeleeuw: I used the existing function present in ClinicalSummaryGenerator as base for this function. The properties were named `maybe...` there, which seemed fine to me to emphasise the nullability. Happy to rename to year/month/day though.
Or is your question on why `year` is nullable at all?

andreia-hartwig: My comment is about the name only. `maybeYear` can raise the question of is the number representing a year or something else (which I don't think is the intention), and if we used `maybe` to convey that non existent, that is redundant since we already have it as a nullable parameter. 

I understand that this is existing code but since we're creating a new class dedicated for this code, it might be a good opportunity to rename the parameters for clarity.

andreia-hartwig (approved):

PR #1538, opened 2026-06-11
ACTIN-2739: Replace Gson with Jackson

dkhartwig:


common/src/main/kotlin/com/hartwig/actin/clinical/serialization/ComorbidityDeserializer.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.hartwig.actin.datamodel.clinical.OtherCondition
+
+object ComorbidityDeserializer : JsonDeserializer<Comorbidity>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Comorbidity {
+        val node: JsonNode = parser.codec.readTree(parser)
+        val className = node.get("comorbidityClass")?.takeUnless { it.isNull }?.asText()
+            ?: throw JsonMappingException.from(parser, "Missing 'comorbidityClass' field in comorbidity: $node")
+        return try {
+            val target = if (className == "COMPLICATION") {
```
andreia-hartwig: I think this was kept for backwards compatibility after a data model change, I would double check if this is still needed.

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/TreatmentDeserializer.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.JsonNode
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentClass
+
+object TreatmentDeserializer : JsonDeserializer<Treatment?>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Treatment? {
+        val node: JsonNode = parser.codec.readTree(parser)
+        if (node.isNull) return null
```
andreia-hartwig: I would avoid having multiple return statements.

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/WhoStatusDeserializer.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.JsonNode
+import com.hartwig.actin.datamodel.clinical.WhoStatus
+import com.hartwig.actin.datamodel.clinical.WhoStatusPrecision
+import java.time.LocalDate
+
+object WhoStatusDeserializer : JsonDeserializer<WhoStatus>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): WhoStatus {
```
andreia-hartwig: I believe this was also kept for backwards compatibility after a data model change. I would double check if still needed.

common/src/main/kotlin/com/hartwig/actin/doid/serialization/DoidJson.kt
```diff
@@ -27,26 +26,32 @@ import com.hartwig.actin.util.json.Json.string
...
 
     const val ID_TO_READ = "http://purl.obolibrary.org/obo/doid.owl"
 
     const val DOID_URL_PREFIX = "http://purl.obolibrary.org/obo/DOID_"
 
     fun readDoidOwlEntry(doidJson: String): DoidEntry {
-        val reader = JsonReader(FileReader(doidJson))
-        reader.isLenient = true
-        val rootObject: JsonObject = JsonParser.parseReader(reader).asJsonObject
+        val rootObject = mapper.createParser(File(doidJson)).use { parser ->
```
andreia-hartwig: This could be a utility function to split the reading and validation of the json file from the actual doid logic.

common/src/main/kotlin/com/hartwig/actin/util/json/ComplexKeyMapModule.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.fasterxml.jackson.databind.SerializationConfig
+import com.fasterxml.jackson.databind.SerializerProvider
+import com.fasterxml.jackson.databind.deser.Deserializers
+import com.fasterxml.jackson.databind.jsontype.TypeDeserializer
+import com.fasterxml.jackson.databind.jsontype.TypeSerializer
+import com.fasterxml.jackson.databind.ser.Serializers
+import com.fasterxml.jackson.databind.type.MapType
+
+/**
+ * Replacement of Gson's enableComplexMapKeySerialization()
```
andreia-hartwig: We generally do not use comments.

common/src/main/kotlin/com/hartwig/actin/util/json/SortedSetSerializer.kt
```diff
@@ -0,0 +1,33 @@
...
+        }
+        gen.writeEndArray()
+    }
+
+    private fun compare(a: Any, b: Any): Int {
+        @Suppress("UNCHECKED_CAST")
+        val comparable = a as? Comparable<Any>
+        if (comparable != null) {
+            try {
+                return comparable.compareTo(b)
```
andreia-hartwig: I would avoid having multiple return statements.

common/src/test/kotlin/com/hartwig/actin/util/json/EligibilityFunctionAdapterTest.kt
```diff
@@ -12,27 +12,38 @@ import com.hartwig.actin.datamodel.trial.SystemicTreatmentParameter
...
+            addDeserializer(EligibilityFunction::class.java, EligibilityFunctionDeserializer)
+            if (includeTreatment) {
+                addDeserializer(Treatment::class.java, TreatmentDeserializer)
+            }
+        }
+    )
 
     @Test
-    fun `Should deserialize eligibility function without modification`() {
+    fun `Should round-trip eligibility function`() {
```
andreia-hartwig: This name could be a bit more explicit and say something like `Should serialize and deserialize eligibility function`.

common/src/test/kotlin/com/hartwig/actin/util/json/EligibilityFunctionAdapterTest.kt
```diff
@@ -12,27 +12,38 @@ import com.hartwig.actin.datamodel.trial.SystemicTreatmentParameter
...
 import com.hartwig.actin.trial.input.EligibilityRule
+import com.hartwig.actin.trial.serialization.EligibilityFunctionDeserializer
+import com.hartwig.actin.trial.serialization.EligibilityFunctionSerializer
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
-class EligibilityFunctionDeserializerTest {
+class EligibilityFunctionAdapterTest {
+
+    private fun mapper(includeTreatment: Boolean = false) = ActinObjectMapper.create().registerModule(
```
andreia-hartwig: I would put the private function after the tests.

common/src/test/kotlin/com/hartwig/actin/util/json/StrictEnumDeserializerTest.kt
```diff
@@ -1,39 +1,29 @@
...
 
     @Test
     fun `Should deserialize valid enum value`() {
-        assertThat(gson.fromJson(JsonPrimitive("Brachytherapy"), RadiotherapyType::class.java))
+        assertThat(mapper.readValue("\"Brachytherapy\"", RadiotherapyType::class.java))
             .isEqualTo(RadiotherapyType.BRACHYTHERAPY)
     }
 
-    @Test
-    fun `Should deserialize nulls for nullable fields`() {
```
andreia-hartwig, line 24: This test has been removed, it null no longer a valid use case?

common/src/test/resources/algo/patient.treatment_match.json
andreia-hartwig, line 1: This change should not require any updates to the existing JSON files. We should be able to read the exact same files with Jackson that we previously read with Gson; otherwise, we can't guarantee backward compatibility.

andreia-hartwig (changes requested):
This change should not require any updates to the existing JSON files. We should be able to read the exact same files with Jackson that we previously read with Gson, otherwise, we can't guarantee backward compatibility.

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/TreatmentDeserializer.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentClass
+
+object TreatmentDeserializer : JsonDeserializer<Treatment?>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Treatment? {
+        val node: JsonNode = parser.codec.readTree(parser)
+        return when {
+            node.isNull -> null
+            else -> {
```
jbartletthmf: Why `when` instead of `if`?

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/TreatmentDeserializer.kt
```diff
@@ -0,0 +1,31 @@
...
+                val className = node.get("treatmentClass")?.takeUnless { it.isNull }?.asText()
+                    ?: throw JsonMappingException.from(parser, "Missing 'treatmentClass' field in treatment: $node")
+                try {
+                    val target = TreatmentClass.valueOf(className).treatmentClass as Class<*>
+                    parser.codec.treeToValue(node, target) as Treatment
```
jbartletthmf: Does it help to have explicit null handling here vs. e.g.
```suggestion
                try {
                    val target = node.get("treatmentClass")?.takeUnless { it.isNull }?.asText()
                        ?.let { TreatmentClass.valueOf(it).treatmentClass as Class<*> }
                    parser.codec.treeToValue(node, target) as Treatment
```

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/TreatmentDeserializer.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.JsonNode
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentClass
+
+object TreatmentDeserializer : JsonDeserializer<Treatment?>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Treatment? {
+        val node: JsonNode = parser.codec.readTree(parser)
```
jbartletthmf: Is there any reason to specify type here?

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/DrugDeserializer.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    private fun requiredField(node: JsonNode, field: String, parser: JsonParser): JsonNode {
+        return node.get(field)?.takeUnless { it.isNull }
+            ?: throw JsonMappingException.from(parser, "Missing required field '$field' in drug: $node")
+    }
+
+    private fun <T> readTree(parser: JsonParser, node: JsonNode?, type: TypeReference<T>): T? {
+        if (node == null || node.isNull) return null
+        return parser.codec.readValue(parser.codec.treeAsTokens(node), type)
```
jbartletthmf:
```suggestion
        return node.takeUnless { it.isNull }?.let { parser.codec.readValue(parser.codec.treeAsTokens(it), type) }
```

common/src/main/kotlin/com/hartwig/actin/clinical/serialization/ComorbidityDeserializer.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.fasterxml.jackson.databind.JsonNode
+import com.hartwig.actin.datamodel.clinical.Comorbidity
+import com.hartwig.actin.datamodel.clinical.ComorbidityClass
+
+object ComorbidityDeserializer : JsonDeserializer<Comorbidity>() {
+
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Comorbidity {
+        val node: JsonNode = parser.codec.readTree(parser)
+        val className = node.get("comorbidityClass")?.takeUnless { it.isNull }?.asText()
+            ?: throw JsonMappingException.from(parser, "Missing 'comorbidityClass' field in comorbidity: $node")
```
jbartletthmf: I don't see why this is handled explicitly outside of try/catch. Do we not get a useful exception message in that case?

common/src/main/kotlin/com/hartwig/actin/trial/serialization/EligibilityFunctionAdapter.kt
```diff
@@ -0,0 +1,157 @@
...
+        if (value == null || value.isNull) return ""
+        return when {
+            value.isValueNode -> value.asText()
+            value.isArray -> value.joinToString(";") { it.asText() }
+            else -> value.toString()
+        }
+    }
+}
+
+object EligibilityFunctionSerializer : JsonSerializer<EligibilityFunction>() {
```
jbartletthmf, line 117: Why do we need a custom serializer for this data class? What happens if we leave it to Jackson?

common/src/main/kotlin/com/hartwig/actin/util/json/StrictEnumDeserializer.kt
```diff
@@ -1,15 +1,16 @@
...
+class StrictEnumDeserializer<T : Enum<T>>(private val enumType: Class<T>) : JsonDeserializer<T>() {
 
-    override fun deserialize(jsonElement: JsonElement, type: Type, context: JsonDeserializationContext): T {
-        val nameToFind = jsonElement.asString.uppercase()
-        return enumType.enumConstants.firstOrNull { it.name == nameToFind }
-            ?: throw IllegalArgumentException("Unknown enum value for type ${enumType.simpleName}: $jsonElement")
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): T {
+        val raw = parser.text?.trim()
+        val normalized = raw?.uppercase()
+        return enumType.enumConstants.firstOrNull { it.name == normalized }
```
jbartletthmf: It would be more efficient to use a map like `private val lookup = enumType.enumConstants.groupBy { it.name }` for fast lookups.

common/src/main/kotlin/com/hartwig/actin/util/json/ComplexKeyMapModule.kt
```diff
@@ -0,0 +1,112 @@
...
+            }
+            tree.forEach { entry ->
+                if (!entry.isArray || entry.size() != 2) {
+                    throw JsonMappingException.from(parser, "Expected 2-element [key, value] array for complex key map entry, got: $entry")
+                }
+                val key: Any? = context.readTreeAsValue(entry.get(0), keyType)
+                val value: Any? = context.readTreeAsValue(entry.get(1), valueType)
+                result[key] = value
+            }
+            return result
```
jbartletthmf: We should be able to do this without a mutable collection:
```suggestion
            return tree.associate { entry ->
                if (!entry.isArray || entry.size() != 2) {
                    throw JsonMappingException.from(parser, "Expected 2-element [key, value] array for complex key map entry, got: $entry")
                }
                context.readTreeAsValue(entry.get(0), keyType) to context.readTreeAsValue(entry.get(1), valueType)
            }
```

common/src/main/kotlin/com/hartwig/actin/util/json/Json.kt
```diff
@@ -1,107 +1,131 @@
...
+    private val logger = KotlinLogging.logger {}
+
+    fun readSingleObjectFromFile(path: String, mapper: ObjectMapper = ActinObjectMapper.create()): ObjectNode {
+        return mapper.createParser(File(path)).use { parser ->
+            val root = parser.readValueAsTree<JsonNode>() as? ObjectNode
+                ?: throw IllegalStateException("Root of JSON file is not a JSON object: $path")
+            if (parser.nextToken() != null) {
+                logger.warn { "More data found in $path after reading main JSON object!" }
+            }
+            root
```
jbartletthmf: I think this could be:
```suggestion
            (parser.readValueAsTree<JsonNode>() as? ObjectNode)?.also {
                if (parser.nextToken() != null) {
                    logger.warn { "More data found in $path after reading main JSON object!" }
                }
            } ?: throw IllegalStateException("Root of JSON file is not a JSON object: $path")
```

common/src/main/kotlin/com/hartwig/actin/util/json/Json.kt
```diff
@@ -1,107 +1,131 @@
...
-import com.google.gson.JsonObject
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.node.ArrayNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+import io.github.oshai.kotlinlogging.KotlinLogging
+import java.io.File
 import java.time.LocalDate
 
 object Json {
```
jbartletthmf: Outside of `DoidJson` (which is rather convoluted but probably not worth rewriting at this time), do we really need this class? The only other usages I see are to `Json.string` and `Json.integer`, which could easily be inlined. Should we just move the necessary functions into `DoidJson` as private helpers? I don't think we want the approach used here to proliferate.

common/src/test/kotlin/com/hartwig/actin/util/json/ClinicalAdaptersTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    fun `Should round-trip OtherCondition via Comorbidity polymorphic deserializer`() {
+        val condition: Comorbidity = OtherCondition(name = "Hypertension", icdCodes = emptySet(), year = 2024, month = 6)
+        val roundtripped = mapper.readValue(mapper.writeValueAsString(condition), Comorbidity::class.java)
+        assertThat(roundtripped).isEqualTo(condition)
+    }
+
+    @Test
+    fun `Should dispatch Toxicity via Comorbidity polymorphic deserializer`() {
```
jbartletthmf: Why `dispatch` vs. `round-trip`?

common/src/test/kotlin/com/hartwig/actin/util/json/EligibilityFunctionAdapterTest.kt
```diff
@@ -48,11 +49,17 @@ class EligibilityFunctionDeserializerTest {
...
+        assertThat(mapper.readValue(mapper.writeValueAsString(function), EligibilityFunction::class.java)).isEqualTo(function)
     }
+
+    private fun mapper(includeTreatment: Boolean = false) = ActinObjectMapper.create().registerModule(
+        SimpleModule().apply {
+            addSerializer(EligibilityFunction::class.java, EligibilityFunctionSerializer)
+            addDeserializer(EligibilityFunction::class.java, EligibilityFunctionDeserializer)
+            if (includeTreatment) {
+                addDeserializer(Treatment::class.java, TreatmentDeserializer)
+            }
```
jbartletthmf: Why isn't this included in all cases?

common/src/test/kotlin/com/hartwig/actin/util/json/LocalDateAdapterTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+class LocalDateAdapterTest {
+
+    @Test
+    fun `Should serialize and deserialize local date`() {
+        val date = LocalDate.of(2022, 10, 20)
+        val mapper = ActinObjectMapper.create()
+
+        assertThat(mapper.writeValueAsString(date)).isEqualTo("\"2022-10-20\"")
+        assertThat(mapper.readValue(mapper.writeValueAsString(date), LocalDate::class.java)).isEqualTo(date)
```
jbartletthmf: could extract `val serialized = mapper.writeValueAsString(date)`

common/src/main/kotlin/com/hartwig/actin/trial/serialization/EligibilityFunctionAdapter.kt
dkhartwig: I ran a quick test removing the custom serializer and letting Jackson serialize natively, then round-tripped through the existing deserializer. It fails on the first non-structured parameter whose value isn't a plain string/list and the deserializer (which expects a string for non‑structured types) throws IllegalArgumentException: Could not resolve string to a treatment type.

common/src/main/kotlin/com/hartwig/actin/util/json/Json.kt
dkhartwig: Good point. Thanks.

PR #1542, opened 2026-06-18
ACTIN-3765: Add HAS_HAD_AT_MOST_X_SYSTEMIC_TREATMENT_LINES_IN_THE_ADVANCED_OR_

spdeleeuw:
...METASTATIC_SETTING and HAS_HAD_AT_MOST_X_SYSTEMIC_TREATMENT_LINES_IN_THE_METASTATIC_SETTING

Intended logic is comparable to HAS_HAD_SYSTEMIC_TREATMENT_IN_THE_ADVANCED_OR_METASTATIC_SETTING

Treatment lines are first filtered by intent: lines with clearly non-metastatic intent (curative/adjuvant/neoadjuvant) are excluded from the count. The remaining lines are evaluated against the configured maximum:
  - PASS: no included lines, or total included lines ≤ maximum
  - FAIL: palliative-intent lines alone exceed the maximum
  - FAIL: total included lines exceed maximum + 1 (even under the most lenient interpretation of uncertain intents, the maximum is exceeded)
  - UNDETERMINED: total included lines exceed the maximum by exactly 1 (one uncertain-intent line could be outside the metastatic setting. That line is given benefit of the doubt)

The +1 buffer exists because intents like MAINTENANCE, INDUCTION, or null can occur in both metastatic and non-metastatic settings. A single uncertain line should not cause a hard fail.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting.kt
```diff
@@ -17,44 +17,43 @@ class HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting(
...
-            priorSystemicTreatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
+            SystemicTreatmentAnalyser.partitionByIntent(priorSystemicTreatments, intentsToIgnore)
         val (recentPotentiallyCorrectIntentTreatments, nonRecentPotentiallyCorrectIntentTreatments) =
             partitionRecentTreatments(includedIntentTreatments, false)
         val (recentPotentiallyCorrectIntentTreatmentsIncludingUnknown, _) =
             partitionRecentTreatments(includedIntentTreatments, true)
         val potentiallyCorrectIntentTreatmentsWithUnknownStopDate = includedIntentTreatments.filter { it.stopYear() == null }
         val palliativeIntentTreatments = priorSystemicTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) == true }
-        val messageEnding = "$settingDescription setting"
-        val wrongIntentMessage = "${Format.concatItemsWithOr(intentsToIgnore)} intent"
```
spdeleeuw, line 34: I think it was incorrect to use this message in the pass messages below, that's why I removed it and rephrased the messages. The actual "wrong" (or excluded) intent treatments are not included into the counts towards at least 1 line of advanced/metastatic setting, so they don't belong in the pass messages.

common/src/main/kotlin/com/hartwig/actin/trial/input/EligibilityRule.kt
```diff
@@ -185,6 +185,8 @@ enum class EligibilityRule(val input: List<Parameter.Type>) {
     HAS_HAD_SYSTEMIC_THERAPY_WITH_ANY_INTENT_X(FunctionInput.MANY_INTENTS),
     HAS_HAD_SYSTEMIC_TREATMENT_IN_METASTATIC_SETTING(FunctionInput.NONE),
     HAS_HAD_SYSTEMIC_TREATMENT_IN_ADVANCED_OR_METASTATIC_SETTING(FunctionInput.NONE),
+    HAS_HAD_AT_MOST_X_SYSTEMIC_TREATMENT_LINES_IN_THE_METASTATIC_SETTING(FunctionInput.ONE_INTEGER),
```
cbruel, line 188: README should be adjusted

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -21,6 +22,13 @@ object SystemicTreatmentAnalyser {
         UNKNOWN
     }
 
+    fun partitionByIntent(
```
cbruel, line 25: add new test to `SystemicTreatmentAnalyserTest`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.hartwig.actin.algo.evaluation.EvaluationFactory
+import com.hartwig.actin.algo.evaluation.EvaluationFunction
+import com.hartwig.actin.algo.evaluation.util.Format.concat
+import com.hartwig.actin.datamodel.PatientRecord
+import com.hartwig.actin.datamodel.algo.Evaluation
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
+import com.hartwig.actin.datamodel.clinical.treatment.history.Intent
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
+
+class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
```
cbruel, line 11: I would expect the exact same logic in this class as in `HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting` (only as extra counting the lines), so I don't really understand why you e.g. don't determine `recentPotentiallyCorrectIntentTreatments` and have the `referenceDate` as input (why would for this new class the date not be important anymore?)?
That being said I would also think it's easy to refactor this into 1 class (or at least place all common logic into 1 class), since there is so much overlapping logic? Only the messages are different?


(probably overlooking something obvious here..)

spdeleeuw: Good point. The referenceDate absence is an inconsistency. The recency heuristic (treatment within 6 months is considered to be in the metastatic setting) is about classifying whether a treatment was in the setting at all, so it should apply equally to counting lines as to checking existence. I'll add it to align the logic in both classes.

On keeping them as separate classes: they answer opposite questions. `HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting` passes when the patient has had treatment in the specified setting (result flips from FAIL to PASS as evidence accumulates), while the max-lines evaluation class passes when they haven't exceeded the limit (result flips from PASS to FAIL). Concretely: a palliative treatment is a PASS trigger in the setting-check but a FAIL trigger in the max-lines check if it puts the patient over the limit. Merging them would require parameterizing the evaluation direction, which is more complexity than two focused classes with shared classification helpers.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -0,0 +1,82 @@
...
+                    createMessage(
+                        "Has had at most $maximumLines systemic treatment line(s) in $settingMessage",
+                        includedIntentTreatments
+                    )
+                )
+        }
+    }
+
+    private fun createMessage(string: String, treatments: List<TreatmentHistoryEntry>): String {
+        return "$string (${concat(treatments.map { it.treatmentDisplay() })})"
```
cbruel: messages can become quite long when you are going to include all these treatments, so not sure if we should add this?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSettingTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        val evaluation = function.evaluate(record)
+        assertEvaluation(EvaluationResult.UNDETERMINED, evaluation)
+        assertThat(evaluation.undeterminedMessagesStrings()).containsExactly(
+            "Uncertain whether maximum of 2 systemic treatment line(s) in metastatic setting is exceeded" +
+                    " (3 lines likely in metastatic setting, setting unclear for some) (Treatment)"
+        )
+    }
+
+    @Test
+    fun `Should fail when non-palliative treatments with unknown stop date push probable count above maximum plus one buffer`() {
```
cbruel, line 99: Do we want unknown stop date treatments to 'push' count to a fail?

cbruel (approved):
Nice!
I made some changes to the test file while testing/reviewing and felt free to commit it, since felt like a waste to remove it :) 
Approved with 2 comments that need I attention I think.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSettingTest.kt
spdeleeuw: Yes, I don't see why not? They should count towards the line count right, and also we have the one line buffer in place.

PR #1543, opened 2026-06-22
ACTIN-3779: Add IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X_WITH_INTENT_Y

ninajacobs:


common/src/main/kotlin/com/hartwig/actin/trial/input/FunctionInput.kt
```diff
@@ -30,6 +30,7 @@ object FunctionInput {
     val MANY_TREATMENT_CATEGORIES = listOf(Parameter.Type.MANY_TREATMENT_CATEGORIES)
     val ONE_SPECIFIC_TREATMENT = listOf(Parameter.Type.TREATMENT)
     val ONE_SPECIFIC_TREATMENT_ONE_INTEGER = listOf(Parameter.Type.TREATMENT, Parameter.Type.INTEGER)
+    val ONE_SPECIFIC_TREATMENT_ONE_INTENT = listOf(Parameter.Type.TREATMENT, Parameter.Type.INTENT)
```
cbruel, line 33: I don't think it's really necessary to create this new `Parameter.Type.INTENT`, we could just use the existing`Parameter.Type.MANY_INTENTS` (and name rule `IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X_WITH_ANY_INTENT_Y`, which would even make the rule more generic).
But since you already did the work for that actin-datamodel PR, it's ok to leave it like this, leave it up to you.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -29,12 +30,13 @@ class IsEligibleForOnLabelTreatment(
...
     private val doidModel: DoidModel,
-    private val minTreatmentDate: LocalDate
+    private val minTreatmentDate: LocalDate,
+    private val intent: Intent? = null
 ) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
         val standardOfCareEvaluator = standardOfCareEvaluatorFactory.create()
-        val treatmentDisplay = treatment.display()
+        val treatmentDisplay = intent?.let { "${intent.name.lowercase()} ${treatment.display()}" } ?: treatment.display()
```
cbruel, line 38: we could use `intent.display()`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -29,12 +30,13 @@ class IsEligibleForOnLabelTreatment(
...
         val isNsclc = DoidEvaluationFunctions.isOfDoidType(doidModel, record.tumor.doids, DoidConstants.LUNG_NON_SMALL_CELL_CARCINOMA_DOID)
 
         return when {
```
cbruel, line 42: This will also be included in the pass message `EvaluationFactory.pass("Eligible for on-label treatment $treatmentDisplay")`. So no matter what the input intent is, we pass? In other words: if you are eligible for e.g. osimertinib as NSCLC patient, you are eligible for adjuvant / curative / palliative osimertinib etc. Is that the logic you intended?

algo/README.md
```diff
@@ -198,6 +198,7 @@ The following criterion algorithms ('rules') are available:
...
...INTENT_Y                                                          | Idem as IS_ELIGIBLE_FOR_ON_LABEL_TREATMENT_X, only showing the intent in the message                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |                                                                                                                                                                                                                                                                                                                                                           |
```
cbruel: readme should be a bit more specific I think about this rule never resolving to pass but to undetermined?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
```diff
@@ -29,19 +29,16 @@ class IsEligibleForOnLabelTreatment(
     private val treatment: Treatment,
     private val standardOfCareEvaluatorFactory: StandardOfCareEvaluatorFactory,
     private val doidModel: DoidModel,
-    private val minTreatmentDate: LocalDate
+    private val minTreatmentDate: LocalDate,
+    private val intent: Intent? = null
 ) : EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
```
cbruel, line 36: minor but I would either remove or add "treatment" after "on-label" for all messages in this class (last two messages are different). I think I would prefer removing it, to keep messages as short as possible

cbruel: and since you use "Undetermined if patient is eligible for on-label $treatmentDisplay", multiple times, I would create a val for this string

cbruel (approved):
looks good and agree with your comments! :) only minor comments

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsEligibleForOnLabelTreatment.kt
ninajacobs: check! Agreed

PR #1544, opened 2026-06-23
ACTIN-3786: Add HAS_LIVER_AND_OR_LYMPH_NODE_AND_OR_LUNG_METASTASES_ONLY

ninajacobs:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLiverAndOrLymphNodeAndOrLungMetastasesOnly.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.hartwig.actin.datamodel.algo.Evaluation
+
+class HasLiverAndOrLymphNodeAndOrLungMetastasesOnly : EvaluationFunction {
+
+    override fun evaluate(record: PatientRecord): Evaluation {
+        with(record.tumor) {
+
+            val hasLiverAndOrLymphNodeAndOrLungMetastases = hasLiverLesions == true || hasLymphNodeLesions == true || hasLungLesions == true
+            val hasNoMetastasesOutsideLiverAndLymphNodeAndLung =
+                hasCnsLesions == false && hasBrainLesions == false && hasBoneLesions == false && otherLesions?.isEmpty() == true &&
```
cbruel: Like in `HasSpecificMetastasesOnly` I think we should make use of `confirmedCategoricalLesionList()` and `suspectedCategoricalLesionList()` here. We shouldn't have to keep the code in this class up to date if we add a new lesion type to our datamodel

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLiverAndOrLymphNodeAndOrLungMetastasesOnly.kt
```diff
@@ -0,0 +1,47 @@
...
+                        hasSuspectedCnsLesions != true && hasSuspectedBrainLesions != true && hasSuspectedBoneLesions != true &&
+                        otherSuspectedLesions?.isNotEmpty() == false
+
+            val hasNoLiverAndNoLymphNodeAndNoLungMetastases =
+                hasLiverLesions == false && hasLymphNodeLesions == false && hasLungLesions == false
+            val hasMetastasesOutsideLiverAndLymphNodeAndLung =
```
cbruel: same here

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/tumor/HasLiverAndOrLymphNodeAndOrLungMetastasesOnlyTest.kt
```diff
@@ -0,0 +1,143 @@
...
+                    hasLiverLesions = false,
+                    hasLymphNodeLesions = false,
+                    hasLungLesions = false
+                )
+            )
+        )
+    }
+
+    @Test
+    fun `Should fail when patient has target lesion but also bone metastases`() {
```
cbruel: I would prefer to just name "liver lesion" in the test name instead of "target lesion" (same for messages below), but no super strong preference

cbruel (approved):

PR #1546, opened 2026-06-25
ACTIN-3795: Fix bug with adding medications to other oncological history

cbruel:
I noticed for a report this week that `HasHadSomeTreatmentsWithCategory` evaluation was incorrect. Too many lines were counted, since patient had chemotherapy mentioned in the oncologicalHistory, but the medications also contained these chemotherapies, which were all counted as a separate line because of using `createTreatmentHistoryEntriesFromMedications` instead of MedicationToTreatmentConverter.convert() (which removes 'duplicates'). 
Replaced all occurrences of `createTreatmentHistoryEntriesFromMedications` with `MedicationToTreatmentConverter.convert()`

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/medication/MedicationToTreatmentConverter.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/medication/MedicationToTreatmentConverter.kt
@@ -12,10 +12,14 @@ import java.time.YearMonth
 
 object MedicationToTreatmentConverter {
 
+    fun convertAndCombine(medications: List<Medication>?, treatmentHistory: List<TreatmentHistoryEntry>): List<TreatmentHistoryEntry> {
+        return treatmentHistory + convert(medications ?: emptyList(), treatmentHistory)
+    }
+
     fun convert(medications: List<Medication>, treatmentHistory: List<TreatmentHistoryEntry>): List<TreatmentHistoryEntry> {
         val treatmentsByDrug = createTreatmentHistoryEntryPerDrugMap(treatmentHistory)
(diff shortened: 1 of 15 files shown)


common/src/main/kotlin/com/hartwig/actin/medication/MedicationToTreatmentConverter.kt
jbartletthmf, line 1: The changes look good, but do we still need this functionality now that we support re-classifying medications as treatment history entries before clinical record creation? We could probably improve LLM support for reclassifying, but I wonder to what extent this problem still exists.

jbartletthmf (approved):

common/src/main/kotlin/com/hartwig/actin/medication/MedicationToTreatmentConverter.kt
cbruel: @jbartletthmf thanks for reviewing and sharing your thoughts :) 
I'm not aware of the possibility to reclassify medications as treatment history? (maybe you are mistaken this with surgeries which can be reclassified as treatment history?). Also, we currently only show medications in the feed UI that require curation (either dosage curation or name curation for medication with missing ATC code) (would be too much noise to show them all), so for example for this patient I didn't see any medications in the UI.
Not sure if I'm in favor of adding this reclassification, because it would put an extra burden on reviewing, since a patient can have quite a big list of medications

jbartletthmf: You're right, I was thinking of surgeries. Sorry.

PR #1549, opened 2026-06-29
ACTIN-3802: Assume PD when subsequent treatment line present and stop…

spdeleeuw:
…Reason not explicitly Tox or treatment pause of >6 months between treatment and subsequent line

This PR adds PD inference from treatment history.                                                                                                                                                                                                                                                                                                   When a treatment entry has no explicit stop reason, PD is inferred if any of the following hold:                                                                                                                                                                                                                                                                                                                                                                                       
- Stop reason is PD → confirmed PD _(existing logic)_                                                                                                                                                                                                                         
- Best response is PD (treatmentResultedInPD only) → confirmed PD _(existing logic)_                                                                                                                                                                                          
- Stop reason is non-null and not PD → confirmed **not** PD _(existing logic)_                                                                                                                                                                                                    
- Subsequent systemic line started within 26 weeks of this entry's stop → inferred PD                                                                                                                                                                      
- Another systemic treatment started during this entry and outlasted it → inferred PD
- Treatment duration exceeded 26 weeks → inferred PD _(existing logic)_                                                                                                                                                                                                         
- Otherwise → undetermined (null) _(existing logic)_                                                                                                                                                                                                                          

The 26-week threshold is the same for both the gap-to-next-line and the duration checks, and acts as "if the patient was on treatment long enough before switching or stopping, it was likely due to progression."                                                                                                                                                                                                                                                                                             Inference based on treatment gap is skipped (returns null) when the stop date is missing, since the gap to the next line cannot be computed

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/clinical/interpretation/ProgressiveDiseaseFunctions.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/clinical/interpretation/ProgressiveDiseaseFunctions.kt
@@ -1,41 +1,68 @@
 package com.hartwig.actin.clinical.interpretation
 
 import com.hartwig.actin.calendar.DateComparison
+import com.hartwig.actin.datamodel.clinical.treatment.Treatment
 import com.hartwig.actin.datamodel.clinical.treatment.history.StopReason
 import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
 import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentResponse
-
-const val MIN_WEEKS_TO_ASSUME_STOP_DUE_TO_PD = 26L // half year
+import java.time.YearMonth
 
 object ProgressiveDiseaseFunctions {
 
-    fun treatmentResultedInPD(treatment: TreatmentHistoryEntry, hasSubsequentLine: Boolean? = null): Boolean? {
+    private const val MIN_WEEKS_TO_ASSUME_STOP_DUE_TO_PD = 26L // half year
+
+    fun treatmentResultedInPD(treatment: TreatmentHistoryEntry, treatmentHistory: List<TreatmentHistoryEntry> = emptyList()): Boolean? {
         val bestResponse = treatment.treatmentHistoryDetails?.bestResponse
         return when {
             bestResponse == TreatmentResponse.PROGRESSIVE_DISEASE -> true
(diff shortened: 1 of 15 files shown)


cbruel:
As discussed on slack, putting it here for look-up later:
* I think a "treatment holiday" (i.e. stable disease and therefore stop for X amount of weeks/months with treatment and continue when it progresses) could last less than 26 weeks, and therefore assuming PD for this treatment is incorrect, because treatment was not stopped because of PD. Therefore I think we should this logic with an oncologist and maybe lower the 26 weeks value to e.g. only 2 months. I think 6 months I quite a long time to not give a pt any treatment while having progressive disease

common/src/main/kotlin/com/hartwig/actin/clinical/interpretation/ProgressiveDiseaseFunctions.kt
```diff
@@ -1,39 +1,72 @@
...
+        val stopMonth = entry.stopMonth()
+        val entryStartYear = entry.startYear
+        val others = history.filter { it !== entry && it.allTreatments().any(Treatment::isSystemic) }
 
-            stopReason != null -> false
+        val hasSubsequent = if (stopYear == null) null else {
+            val entryStop = YearMonth.of(stopYear, stopMonth ?: 12)
+            others.mapNotNull { other -> other.startYear?.let { year -> other to year } }.any { (other, otherStartYear) ->
+                val otherStart = YearMonth.of(otherStartYear, other.startMonth ?: 1)
+                when {
```
cbruel: I think it would be easier to read if you create a function here, I just committed a suggestion (was easier than typing it here), but if you disagree, please feel free to change / revert ofcourse

cbruel (approved):
If you approve my review changes, I approve yours :)

common/src/main/kotlin/com/hartwig/actin/clinical/interpretation/ProgressiveDiseaseFunctions.kt
spdeleeuw: I like it. I refactored a little bit since it changed the logic of my original implementation. An entry without start date would slip through the subsequenLineSuggestsPD function, while we cannot evaluate subsequency in that case. Also I added logic if both have stop date unknown but there is still a subsequent line it still runs the inferred PD function (this would happen often in practice since stop date is often null)

PR #1556, opened 2026-07-17
ACTIN-3884: Add HAS_SOFT_TISSUE_METASTASES

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadCombinedTreatmentNamesWithCycles.kt
```diff
@@ -78,18 +78,21 @@ class HasHadCombinedTreatmentNamesWithCycles(
...
-                when (it.treatmentHistoryDetails?.cycles) {
+                when (val cycles = it.treatmentHistoryDetails?.cycles) {
                     null -> EvaluationResult.UNDETERMINED
-                    in minCycles..maxCycles -> EvaluationResult.PASS
-                    else -> EvaluationResult.WARN
+                    else -> if (cycles >= minCycles && (maxCycles == null || cycles <= maxCycles)) {
+                        EvaluationResult.PASS
+                    } else {
+                        EvaluationResult.WARN
+                    }
```
spdeleeuw: You could keep this one when-statement, like this:
```
.groupBy {
    when (val cycles = it.treatmentHistoryDetails?.cycles) {
        null -> EvaluationResult.UNDETERMINED
        in minCycles..(maxCycles ?: Int.MAX_VALUE) -> EvaluationResult.PASS
        else -> EvaluationResult.WARN
    }
}
```

common/src/main/kotlin/com/hartwig/actin/molecular/interpretation/TumorOriginInterpreter.kt
```diff
@@ -13,8 +12,7 @@ private const val MAX_PREDICTIONS_TO_DISPLAY = 3
 class TumorOriginInterpreter(val hasSufficientQuality: Boolean?, val predictedTumorOrigin: PredictedTumorOrigin?) {
 
     fun hasConfidentPrediction(): Boolean {
-        val roundedLikelihood = predictedTumorOrigin?.likelihood()?.let { (it * 100).roundToInt() / 100.0 }
-        return hasSufficientQuality == true && roundedLikelihood?.let { it >= LIKELIHOOD_CONFIDENCE_THRESHOLD } == true
+        return hasSufficientQuality == true && predictedTumorOrigin?.likelihood()?.let { it >= LIKELIHOOD_CONFIDENCE_THRESHOLD } == true
```
spdeleeuw, line 15: Is this related to the changes in this PR?

cbruel: thanks for catching, this shouldn't be here indeed!

cbruel: oh no wait, this is weird, this is a change I reverted yesterday after merging, but apparently didn't go well. the change you see here is correct at least (this is the code how it was before I made changes to it yesterday), so will leave it here

spdeleeuw: Might have to do with on which commit this PR is based. Anyhow, if this is the correct version of this code then this PR is approved

cbruel: I think the problem is that i committed but didn't push my revert yesterday, so I notice master doesn't contain the correct version now. So will be fixed with this PR :)

PR #1558, opened 2026-07-20
ACTIN-3867: Allow for DOID codes

cbruel:
Same as for ICD titles/code, since this merge (https://github.com/hartwigmedical/actin-trial/pull/238) we are now allowed to also input DOID codes instead of only terms during trial curation, we should correctly ingest this in algo.


spdeleeuw (approved):
LGTM

PR #1562, opened 2026-07-31
ACTIN-3944 Review report for Hartwig Trial Search positioning

andreia-hartwig:
We now have 3 properties files, which are selected based on the type of report to generate. 

Since there are small differences between TM and TM RUO, the `trial_matching_ruo_overrides.properties` was designed to work as an override on `trial_matching.properties`. Avoiding having to keep the messages duplicated across both file. 

I opted to keep `personalisation.properties` and `trial_matching.properties` fully separate since these diverge a lot more. 

Additionally this PR introduces a few formatting adjustments to align with the OncoAct report: 
- Added the HMF logo
- Adjusted the header sizes
- Adjusted table context text color
- Adjusted the table header cell layout
- Adjusted table sub title weight

Two bugs found in the report were also fixed: 
- Rows in the table showing international trials, would split between pages mid row.
- Page breaking between title and table. With the implemented fix, the tradeoff is that the table header will repeat on every page the table spans, which can be a positive consequence. 

Depends on PR: https://github.com/hartwigmedical/actin-infrastructure/pull/425


common/src/main/kotlin/com/hartwig/actin/configuration/ReportConfiguration.kt
```diff
@@ -68,6 +69,7 @@ data class ReportConfiguration(
 
         fun extended(): ReportConfiguration {
             return ReportConfiguration(
+                reportType = ReportType.PERSONALISATION,
```
cbruel: I think the extended report is something else than the personalisation report, so not sure if reportType should be set to `ReportType.PERSONALISATION` in this extended() function?

cbruel (commented):
one small comment/thought while going through these files

common/src/main/kotlin/com/hartwig/actin/configuration/ReportConfiguration.kt
andreia-hartwig: @cbruel I selected Personalization between it's uses the following two settings: 
- standardOfCareSummaryType = ReportContentType.COMPREHENSIVE,
- efficacyEvidenceChapterType = EfficacyEvidenceChapterType.COMPLETE

These are ACTIN 2.0 specific right?

cbruel: @andreia-hartwig for personalisation `efficacyEvidenceChapterType` is `STANDARD_OF_CARE_ONLY` (see `ExampleFunctions.createPersonalizationReportConfiguration`. This `extended()` is to create an 'extended' report (we don't use that one in practice anymore), but it just contains more information for some of the sections, to provide the user more information. But extended() should be similar to the default ReportConfiguration, just everything set to COMPREHENSIVE / COMPLETE. So I would remove `reportType = ReportType.PERSONALISATION` (and thus use the default set in ReportConfiguration above)

andreia-hartwig: @cbruel This function should indeed not be choosing the report type. I've changed the extended function to simply update the relevant fields from an existing ReportConfiguration.

andreia-hartwig:
This PR is now up to date with master, removing the report's extended mode.

PR #1561, opened 2026-07-29
ACTIN-3927 Manage labels used in report with property files

andreia-hartwig:
With this PR I introduce ReportLabels as the single access point for all user-facing strings in the report PDF, backed by .properties files per ReportIntendedUse. This is preparation to enable showing different copy depending on the intended use. 

All hardcoded strings were moved to the properties file with no copy changes, this is a structural refactor only. 
The regression test works without changes with this new setup.

PR #1564, opened 2026-08-04
ACTIN-3947 Manage labels created by algo with property files

andreia-hartwig:
With this PR I introduce EvaluationLabels as the single access point for all user-facing labels generated by algo rules, backed by .properties files per ReportIntendedUse. This is preparation to enable showing different copy depending on the intended use.

Hardcoded strings were moved to the properties file with no copy changes, this is a structural refactor only.

This PR follows the same strategy for handling labels as followed in the report module [here](https://github.com/hartwigmedical/actin/pull/1561).

Ideally, the algo module would not actually produce labels but keys & parameters that would then be converted to labels when generating the report. That would represent a bigger change and if we ever want to do it, it will be easier when having the labels centralised.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/bloodtransfusion/RequiresRegularHematopoieticSupport.kt
```diff
@@ -11,25 +12,31 @@ import com.hartwig.actin.medication.AtcTree
...
-        val medications = record.medications ?: return MEDICATION_NOT_PROVIDED
+        val medications = record.medications ?: return medicationNotProvided(medicationLabels)
         val filteredMedications = medications
             .filter { activeBetweenDates(it) }
             .filter { it.atc?.chemicalSubGroup in resolvedCategories }
             .map { it.name }
         return if (filteredMedications.isNotEmpty()) {
-            EvaluationFactory.pass("Has received recent hematopoietic support (${concatLowercaseWithCommaAndAnd(filteredMedications)})")
+            EvaluationFactory.pass(
+                bloodTransfusionLabels.requiresRegularHematopoieticSupportPassMedication(concatLowercaseWithCommaAndAnd(filteredMedications))
```
jbartletthmf, line 36: long line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/cardiacfunction/EcgMeasureEvaluationFunctions.kt
```diff
@@ -1,37 +1,41 @@
 package com.hartwig.actin.algo.evaluation.cardiacfunction
 
+import com.hartwig.actin.algo.evaluation.EvaluationLabels
 import com.hartwig.actin.algo.evaluation.cardiacfunction.EcgMeasureEvaluationFunction.ThresholdCriteria
 import com.hartwig.actin.datamodel.clinical.Ecg
 
 object EcgMeasureEvaluationFunctions {
 
-    fun hasLimitedQtcf(maxQtcf: Double): EcgMeasureEvaluationFunction {
+    fun hasLimitedQtcf(maxQtcf: Double, labels: EvaluationLabels.CardiacFunction): EcgMeasureEvaluationFunction {
```
jbartletthmf, line 9: It would be nice if we could achieve this without having to explicitly pass the labels object around everywhere. I'm not sure what that would look like though, as we probably don't want to add a full DI framework just to support labels. One lighter-weight version of this would be to have a singleton `EvaluationLabels` object that can be configured once and then accessed by everyone, e.g.:
```kotlin
object EvaluationLabels {
    private lateinit var configuration: String

    private val labels: Map<String, String> by lazy {
        check(::configuration.isInitialized) {
            "LabelResolver has not been configured"
        }
       load(configuration)
    }

    fun configure(configuration: String) {
        check(!::configuration.isInitialized) {
            "LabelResolver has already been configured"
        }
        this.configuration = configuration
    }

    // ...
}
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
```diff
@@ -2,35 +2,34 @@ package com.hartwig.actin.algo.evaluation.general
...
         val patientWho = who?.let { who.asText() }
 
         return when {
-            who == null -> EvaluationFactory.undetermined(
-                "Undetermined if WHO status is within requested max WHO $maximumWHO (WHO missing)"
-            )
+            who == null -> EvaluationFactory.undetermined(labels.hasMaximumWhoStatusUndeterminedMissing(maximumWHO))
 
-            evaluation == EvaluationResult.PASS -> EvaluationFactory.pass("WHO $patientWho is below requested max WHO $maximumWHO")
+            evaluation == EvaluationResult.PASS -> EvaluationFactory.pass(labels.hasMaximumWhoStatusPass(patientWho!!, maximumWHO))
```
jbartletthmf, line 23: The pass/fail/recoverableFail/undetermined versions here have me thinking. Would it make sense to define an interface (`EvaluationMessageFactory`?) with functions returning strings like `pass()`, `fail()`, etc., and when you call something like `labels.hasMaximumWhoStatus(patientWho!!, maximumWHO)`, it returns an object that implements that interface? Then you might be able to just pass in the `EvaluationResult`, recoverability, and `labels.hasMaximumWhoStatus(patientWho!!, maximumWHO)` to the `EvaluationFactory` and let it take care of selecting the right string. This would save us from having to pass in the same parameters to a slightly different `labels` function for each outcome.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/EvaluationLabels.kt
```diff
@@ -0,0 +1,3641 @@
...
+            get("treatment.has.had.progression.following.latest.treatment.line.fail")
+
+        fun hasHadProgressionFollowingLatestTreatmentLineRecoverableUndetermined() =
+            get("treatment.has.had.progression.following.latest.treatment.line.recoverable.undetermined")
+
+        fun hasHadSomeTreatmentsWithCategoryPass(minTreatmentLines: Int, category: String) =
+            format("treatment.has.had.some.treatments.with.category.pass", minTreatmentLines, category)
+
+        fun hasHadSomeTreatmentsWithCategoryUndetermined(minTreatmentLines: Int, category: String) =
+            format("treatment.has.had.some.treatments.with.category.undetermined", minTreatmentLines, category)
```
jbartletthmf, line 2919: Should the word separations be more meaningful here? E.g.
```suggestion
            format("treatment.hasHadSomeTreatmentsWithCategory.undetermined", minTreatmentLines, category)
```

algo/src/main/resources/evaluation/research_use_only.properties
jbartletthmf, line 1: Is this file meaningfully different than non_medical at this point? Just curious.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/comorbidity/HasChildPughScoreTest.kt
```diff
@@ -1,16 +1,20 @@
...
+import com.hartwig.actin.configuration.ReportIntendedUse
 import com.hartwig.actin.datamodel.algo.EvaluationResult
 import com.hartwig.actin.icd.TestIcdFactory
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
 class HasChildPughScoreTest {
-    private val function = HasChildPughScore(TestIcdFactory.createTestModel(), listOf("A", "B"))
+    private val function = HasChildPughScore(
+        TestIcdFactory.createTestModel(), listOf("A", "B"), EvaluationLabels.load(ReportIntendedUse.RESEARCH_USE_ONLY).comorbidity
```
jbartletthmf, line 16: Does all this loading slow down the build? If so, can we share a global instance or use a simple mock?

jbartletthmf (commented):
I didn't read every line 😛 

It looks OK, but it would be nice if we could avoid some of the overhead...

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
jbartletthmf: This could actually be a step towards deferring label resolution to reporting, as you suggested in the description.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/cardiacfunction/EcgMeasureEvaluationFunctions.kt
andreia-hartwig: I actually had an initial version where EvaluationLabels was a singleton but I changed it to be consistent with the implementation in the report module as well. 
On one side having to explicitly pass the labels object around is not ideal but I generally don't like using singleton like that. 

Do you think it's worth changing? In which case I would probably change report as well.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/general/HasMaximumWHOStatus.kt
andreia-hartwig: That sounds like an interesting improvement but I would prefer to keep it out of the scope of this change.

algo/src/main/resources/evaluation/research_use_only.properties
andreia-hartwig: Right now they are the same.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/comorbidity/HasChildPughScoreTest.kt
andreia-hartwig: I did not notice an increase of build time, but I didn't really measure it. It's just loading a properties file.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/cardiacfunction/EcgMeasureEvaluationFunctions.kt
jbartletthmf: Yes, I would prefer a singleton for ease of use for a simple dependency like this. I didn't think the overhead was that bad in the `report` module (there's much less code there) but not a bad idea to change both for consistency.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/comorbidity/HasChildPughScoreTest.kt
jbartletthmf: It's reading and parsing a large file from disk for every test that depends on labels. That could have an impact over many tests, but happy to hear that you didn't notice anything. With the singleton approach, maybe we can instantiate it once for all tests for efficiency.

PR #1566, opened 2026-08-07
ACTIN-3975: Bug fix in HAS_HAD_TREATMENT_NAME_X_WITHIN_Y_WEEKS evaluation

ninajacobs:
We saw a case where the treatment of interest is administered in the last line (without stop date), but we confidently evaluate that the treatment would have been stopped X weeks ago, which I believe is wrong.

FYI as I understand:
- If there is a stop date, the case was handled in treatmentSinceMinDate by evaluating stop date directly.
- If there is a start date, the case is handled in treatmentSinceMinDate - by evaluating if start date occurred after min date (then you are sure).
- If stop date and start date unknown, the case is implemented in treatmentSinceMinDate too.

However, the case if there is a start date, but the start date occurred _before_ min date, _but_ the treatment is the last treatment line is not handled. This is added in this PR.

Of note: I think this case is still not handled perfectly if we look for a treatment of many weeks ago, and we may want to look back multiple treatment lines (then you would actually want to derive a max stop date from the treatment occurring after the treatment of interest). But this requiring some more substantial refactoring I believe. We use this rule for HAS_HAD_TREATMENT_NAME_X_WITHIN_Y_WEEKS so only a number of weeks, hence we won’t need to look back across multiple lines I expect, so out of scope for now?

### DIFF
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
@@ -16,38 +16,33 @@ object TreatmentVersusDateFunctions {
     ): Evaluation {
         val matchingTreatments = record.oncologicalHistory
             .mapNotNull { entry -> TreatmentHistoryEntryFunctions.portionOfTreatmentHistoryEntryMatchingPredicate(entry, predicate) }
+        val matchingPortionOfMostRecentTreatment = record.oncologicalHistory.filter { it.startYear != null }
+            .maxWithOrNull(TreatmentHistoryEntryStartDateComparator())
+            ?.let { TreatmentHistoryEntryFunctions.portionOfTreatmentHistoryEntryMatchingPredicate(it, predicate) }
+
 
         return when {
             matchingTreatments.any { treatmentSinceMinDate(it, minDate, false) } -> {
                 EvaluationFactory.pass("Treatment $predicateDescription administered since ${Format.date(minDate)}")
             }
 
-            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } -> {
-                EvaluationFactory.undetermined(
-                    "Treatment $predicateDescription administered with unknown date hence " +
-                            "undetermined if administered since ${Format.date(minDate)}"
-                )
-            }
-
-            matchingTreatments.any { isMostRecentTreatmentWithoutStopDate(it, record.oncologicalHistory) == true } -> {
+            matchingPortionOfMostRecentTreatment?.let { stopDateCouldBeSinceMinDate(it, minDate) } == true -> {
                 EvaluationFactory.undetermined(
                     "Undetermined if treatment $predicateDescription may have been administered since " +
                             "${Format.date(minDate)} because it is the last treatment line and stop date missing"
                 )
             }
 
-            matchingTreatments.any { isMostRecentTreatmentWithoutStopDate(it, record.oncologicalHistory) == null } -> {
-                EvaluationFactory.undetermined(
-                    "Undetermined if treatment $predicateDescription may have been administered since " +
-                            "${Format.date(minDate)} because it is undetermined if treatment may be the last treatment line"
-                )
+            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } -> {
+                EvaluationFactory.undetermined("Treatment $predicateDescription administered with unknown date hence " +
+                        "undetermined if administered since ${Format.date(minDate)}")
             }
 
             matchingTreatments.isNotEmpty() -> {
                 EvaluationFactory.fail("All treatments $predicateDescription administered before ${Format.date(minDate)}")
             }
 
-            else -> EvaluationFactory.fail("No treatments $predicateDescription in prior history")
+            else -> EvaluationFactory.fail("No treatments $predicateDescription in history")
         }
     }
 
@@ -66,14 +61,7 @@ object TreatmentVersusDateFunctions {
             ?: includeUnknown
     }
 
-    private fun isMostRecentTreatmentWithoutStopDate(entry: TreatmentHistoryEntry, history: List<TreatmentHistoryEntry>): Boolean? {
-        val (historyWithoutDates, historyWithDates) = history.partition { it.startYear == null }
-        val mostRecentHistoryEntryWithDate = historyWithDates.maxWithOrNull(TreatmentHistoryEntryStartDateComparator())
-
-        return when {
-            historyWithoutDates.contains(entry) -> null
-            mostRecentHistoryEntryWithDate?.stopYear() != null -> false
-            else -> mostRecentHistoryEntryWithDate == entry
-        }
+    private fun stopDateCouldBeSinceMinDate(entry: TreatmentHistoryEntry, minDate: LocalDate): Boolean {
+        return DateComparison.isAfterDate(minDate, entry.stopYear(), entry.stopMonth()) != false
     }
 }
--- a/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
+++ b/algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
@@ -18,17 +18,44 @@ object TreatmentVersusDateFunctions {
             .mapNotNull { entry -> TreatmentHistoryEntryFunctions.portionOfTreatmentHistoryEntryMatchingPredicate(entry, predicate) }
 
         return when {
-            matchingTreatments.any { treatmentSinceMinDate(it, minDate, false) } ->
+            matchingTreatments.any { treatmentSinceMinDate(it, minDate, false) } -> {
                 EvaluationFactory.pass("Treatment $predicateDescription administered since ${Format.date(minDate)}")
+            }
 
-            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } ->
-                EvaluationFactory.undetermined("Treatment $predicateDescription administered with unknown date")
+            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } -> {
+                EvaluationFactory.undetermined("Treatment $predicateDescription administered with unknown date hence " +
+                        "undetermined if administered since ${Format.date(minDate)}")
+            }
 
-            matchingTreatments.isNotEmpty() ->
+            matchingTreatments.any {
+                isMostRecentTreatment(
+                    it,
+                    record.oncologicalHistory
+                ) == true
+            } -> {
+                EvaluationFactory.undetermined(
+                    "Undetermined if treatment $predicateDescription may have been administered since " +
+                            "${Format.date(minDate)} because it is the last treatment line and stop date missing"
+                )
+            }
+
+            matchingTreatments.any {
+                isMostRecentTreatment(
+                    it,
+                    record.oncologicalHistory
+                ) == null
+            } -> {
+                EvaluationFactory.undetermined(
+                    "Undetermined if treatment $predicateDescription may have been administered since " +
+                            "${Format.date(minDate)} because it is undetermined if treatment may be the last treatment line"
+                )
+            }
+
+            matchingTreatments.isNotEmpty() -> {
                 EvaluationFactory.fail("All treatments $predicateDescription administered before ${Format.date(minDate)}")
+            }
 
-            else ->
-                EvaluationFactory.fail("No treatments $predicateDescription in prior history")
+            else -> EvaluationFactory.fail("No treatments $predicateDescription in prior history")
         }
     }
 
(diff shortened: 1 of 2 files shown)


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -18,17 +18,44 @@ object TreatmentVersusDateFunctions {
...
 
-            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } ->
-                EvaluationFactory.undetermined("Treatment $predicateDescription administered with unknown date")
+            matchingTreatments.any { treatmentSinceMinDate(it, minDate, true) } -> {
+                EvaluationFactory.undetermined("Treatment $predicateDescription administered with unknown date hence " +
+                        "undetermined if administered since ${Format.date(minDate)}")
+            }
 
-            matchingTreatments.isNotEmpty() ->
+            matchingTreatments.any {
```
spdeleeuw: This should all fit on one line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -18,17 +18,44 @@ object TreatmentVersusDateFunctions {
...
+                    record.oncologicalHistory
+                ) == true
+            } -> {
+                EvaluationFactory.undetermined(
+                    "Undetermined if treatment $predicateDescription may have been administered since " +
+                            "${Format.date(minDate)} because it is the last treatment line and stop date missing"
+                )
+            }
+
+            matchingTreatments.any {
```
spdeleeuw: This should all fit on one line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -18,17 +18,44 @@ object TreatmentVersusDateFunctions {
...
-            matchingTreatments.isNotEmpty() ->
+            matchingTreatments.any {
+                isMostRecentTreatment(
+                    it,
+                    record.oncologicalHistory
+                ) == true
+            } -> {
+                EvaluationFactory.undetermined(
+                    "Undetermined if treatment $predicateDescription may have been administered since " +
+                            "${Format.date(minDate)} because it is the last treatment line and stop date missing"
```
spdeleeuw, line 32: `...and stop date missing` -> this isn't actually checked in `fun isMostRecentTreatment`?

I think before in case of "last line + known stop date" it would `FAIL` and now this could be `UNDETERMINED` with the message containing "stop date missing". That seems like an unintended regression?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -7,9 +7,26 @@ import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory
 import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatment
 import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatmentHistoryEntry
 import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
+import org.assertj.core.api.Assertions.assertThat
 import java.time.LocalDate
 import org.junit.jupiter.api.Test
 
+private const val YEARS_TO_SUBTRACT = 3
+
+private val NOW = LocalDate.now()
```
spdeleeuw: We tend not to use `LocalDate.now()` since that could lead to unexpected behaviour from one day on the other, so we usually just pick a specific date as the reference.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -7,9 +7,26 @@ import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory
...
+private val TARGET_DATE = NOW.minusYears(1)
+private val RECENT_DATE = TARGET_DATE.plusYears(1)
```
spdeleeuw, line 18: These constants read a bit awkward. Now = now, target is now minus 1, recent is target + 1 (so recent is actually the same as now, older is target minus 1 thus older is now minus 2 years...). It feels that this could be simplified a bit?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -19,72 +36,98 @@ abstract class TreatmentVersusDateFunctionsTestAbstract {
...
+    fun `Should pass when treatment history includes matching treatment within range`() {
+        val treatmentHistory =
+            listOf(NON_MATCHING_RECENT_TREATMENT, matchingOlderTreatment(), matchingTreatment(RECENT_DATE.year, RECENT_DATE.monthValue))
         assertEvaluation(EvaluationResult.PASS, function().evaluate(TreatmentTestFactory.withTreatmentHistory(treatmentHistory)))
     }
 
     @Test
-    fun shouldReturnUndeterminedWhenMatchingTreatmentHasUnknownYear() {
-        val treatmentHistory = listOf(NON_MATCHING_TREATMENT, olderTreatment(), matchingTreatment(null, 10))
+    fun `Should be undetermined when matching treatment has unknown stop year`() {
```
spdeleeuw: You're testing a treatment with unknown stopYear **and unknown startYear** here, so the title should be updated, but even better to test both missing properties in separate tests?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -19,72 +36,98 @@ abstract class TreatmentVersusDateFunctionsTestAbstract {
...
-            function().evaluate(TreatmentTestFactory.withTreatmentHistory(listOf(NON_MATCHING_TREATMENT)))
+            function().evaluate(TreatmentTestFactory.withTreatmentHistory(listOf(NON_MATCHING_RECENT_TREATMENT)))
         )
     }
 
     @Test
-    fun shouldFailWhenMatchingTreatmentIsOlderByYear() {
-        val treatmentHistory = listOf(NON_MATCHING_TREATMENT, olderTreatment())
+    fun `Should fail when matching treatment is older by year`() {
+        val treatmentHistory = listOf(NON_MATCHING_RECENT_TREATMENT, matchingOlderTreatment())
```
spdeleeuw: This test is failing because the most recent treatment is non-matching, not because the matching treatment is too old... If you remove the `NON_MATCHING_RECENT_TREATMENT` it goes from `FAIL` to `UNDETERMINED`.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -19,72 +36,98 @@ abstract class TreatmentVersusDateFunctionsTestAbstract {
...
-    }
-
-    @Test
-    fun shouldPassWhenPriorTreatmentHasUnknownStopDateButStartDateInRange() {
+    fun `Should pass when prior treatment has unknown stop date but start date within range`() {
         val treatmentHistory = listOf(
-            NON_MATCHING_TREATMENT,
+            NON_MATCHING_RECENT_TREATMENT,
             matchingTreatment(NOW.minusYears(YEARS_TO_SUBTRACT.toLong()).year, null),
             matchingTreatment(NOW.year, null, RECENT_DATE.year, RECENT_DATE.monthValue)
```
spdeleeuw: Stop year is set to NOW.year here, so the naming of the test seems off (stop date is partly known).

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
spdeleeuw, line 34: I know this was pre-existing but it reads a bit awkward that the stop-year/month arguments come first. Also, this fits on one line.

spdeleeuw (commented):
I think there are still several issues in the class and tests. For instance: this test returns undetermined but it should return a FAIL (so the test itself fails). 

```
@Test
    fun `Should fail when matching treatment is most recent line but stop date known and before min date`() {
        val treatmentHistory = listOf(matchingTreatment(startYear = OLDER_DATE.minusYears(1).year, startMonth = 3, stopYear = OLDER_DATE.year, stopMonth = OLDER_DATE.monthValue))
        assertEvaluation(EvaluationResult.FAIL, function().evaluate(TreatmentTestFactory.withTreatmentHistory(treatmentHistory)))
    }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
ninajacobs: check. with new code changes this test actually fails!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -74,13 +66,14 @@ object TreatmentVersusDateFunctions {
             ?: includeUnknown
     }
 
-    private fun isMostRecentTreatment(entry: TreatmentHistoryEntry, history: List<TreatmentHistoryEntry>): Boolean? {
+    private fun isMostRecentTreatmentWithoutStopDate(entry: TreatmentHistoryEntry, history: List<TreatmentHistoryEntry>): Boolean? {
```
spdeleeuw: The double negation is not ideal ("not without stop date"), but I'm okay with leaving it as is.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -13,8 +13,8 @@ import org.junit.jupiter.api.Test
 
 private const val YEARS_TO_SUBTRACT = 3
 
-private val NOW = LocalDate.now()
-private val TARGET_DATE = NOW.minusYears(1)
+private val CURRENT_DATE = LocalDate.of(2025, 6, 1)
+private val TARGET_DATE = CURRENT_DATE.minusYears(1)
 private val RECENT_DATE = TARGET_DATE.plusYears(1)
```
spdeleeuw, line 18: `RECENT_DATE` == `CURRENT_DATE` right? So why not drop `RECENT_DATE`?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTestAbstract.kt
```diff
@@ -82,21 +82,33 @@ abstract class TreatmentVersusDateFunctionsTestAbstract {
     fun `Should pass when prior treatment has unknown stop date but start date within range`() {
         val treatmentHistory = listOf(
             NON_MATCHING_RECENT_TREATMENT,
-            matchingTreatment(NOW.minusYears(YEARS_TO_SUBTRACT.toLong()).year, null),
-            matchingTreatment(NOW.year, null, RECENT_DATE.year, RECENT_DATE.monthValue)
+            matchingTreatment(CURRENT_DATE.year, CURRENT_DATE.monthValue, RECENT_DATE.year, RECENT_DATE.monthValue)
```
spdeleeuw: RECENT == CURRENT as mentioned above. I think it's clearer if you use only one of these. Also, the test name says unknown stop date, but actually stopYear and stopMonth are populated?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -74,13 +66,14 @@ object TreatmentVersusDateFunctions {
...
         val (historyWithoutDates, historyWithDates) = history.partition { it.startYear == null }
```
spdeleeuw: ...with/without**Start**Dates ?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -74,13 +66,14 @@ object TreatmentVersusDateFunctions {
...
+        val mostRecentHistoryEntryWithDate = historyWithDates.maxWithOrNull(TreatmentHistoryEntryStartDateComparator())
 
-        return if (historyWithoutDates.isNotEmpty() && historyWithDates.maxWithOrNull(TreatmentHistoryEntryStartDateComparator()) == entry) {
-            null
-        } else {
-            historyWithDates.maxWithOrNull(TreatmentHistoryEntryStartDateComparator()) == entry
+        return when {
+            historyWithoutDates.contains(entry) -> null
```
spdeleeuw: What if a treatment has no start date, but a known stop date which is before `minDate`. That should FAIL then right (which I think it does on the version on `master` branch), but now this would return undetermined because of this first when clause?

spdeleeuw (commented):
I think there's still a bug in here. A treatment without start date but with a stop date before the min date now is undetermined but should actually fail. Also when the stop date matches min date but the month is unknown, it now fails but I think that should return undetermined. I've pushed a refactor in a branch forked from yours: https://github.com/hartwigmedical/actin/pull/1569 . There are two extra test cases and all existing tests pass. Let me know what you think.
The idea is to revert the question: first "what is the most recent line" and second "are you a matching treatment?", instead of the other way around.

PR #1567, opened 2026-08-07
ACTIN-3976: Add rule CURRENTLY_GETS_TREATMENT_WITH_ANY_DRUG_X

ninajacobs:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
```diff
@@ -34,31 +36,40 @@ class HasHadTreatmentWithDrugFromSetAsMostRecent(private val drugsToMatch: Set<D
...
-                EvaluationFactory.pass("Has received $matchingDrugDisplay as most recent treatment")
+                if (!requireCurrentAdministration) {
+                    EvaluationFactory.pass("Has received $matchingDrugDisplay as most recent treatment")
+                } else {
+                    EvaluationFactory.undetermined("Has received $matchingDrugDisplay as most recent treatment but unknown if currently still administered")
+                }
             }
 
             matchingDrugsInUnknownTreatmentLines.isNotEmpty() || matchingDrugsInMostRecentLineWithDate.isNotEmpty() -> {
                 val drugList = Format.concatItemsWithAnd(matchingDrugsInUnknownTreatmentLines + matchingDrugsInMostRecentLineWithDate)
```
spdeleeuw, line 59: You could consider to inline all these since they're only used once, within the scope of this when-statement

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
```diff
@@ -34,31 +36,40 @@ class HasHadTreatmentWithDrugFromSetAsMostRecent(private val drugsToMatch: Set<D
...
 
             possibleTrialMatch(if (history.size == 1) history.first() else mostRecentTreatmentEntry) -> {
-                EvaluationFactory.undetermined("Undetermined if treatment received in previous trial included ${Format.concatItemsWithOr(drugsToMatch)}")
+                val currentlyDisplay = if (requireCurrentAdministration) " and unknown if currently still administered" else ""
+                EvaluationFactory.undetermined(
+                    "Undetermined if treatment received in previous trial included ${
+                        Format.concatItemsWithOr(
+                            drugsToMatch
+                        )
+                    }$currentlyDisplay"
```
spdeleeuw: If this doesn't fit on one line, maybe on two? 
```
"Undetermined if treatment received in previous trial included + 
    ${Format.concatItemsWithOr(drugsToMatch)}$currentlyDisplay"
```

spdeleeuw (commented):
I think I'm missing the stop date in the evaluation here. We could return a fail for the `requireCurrentAdministration` case when it's the most recent line but a stop date is known? Or am I overlooking something?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
ninajacobs: Thanks, but. I think I like it for readability to keep them as is :)

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
```diff
@@ -30,17 +30,23 @@ class HasHadTreatmentWithDrugFromSetAsMostRecent(private val drugsToMatch: Set<D
...
 
         return when {
             matchingDrugsInMostRecentLine.isNotEmpty() -> {
                 val matchingDrugDisplay = Format.concatItemsWithAnd(matchingDrugsInMostRecentLine)
                 if (!requireCurrentAdministration) {
                     EvaluationFactory.pass("Has received $matchingDrugDisplay as most recent treatment")
+                } else if (mostRecentMatchingEntryHasStopDate) {
+                    EvaluationFactory.fail("Does not currently receive $matchingDrugDisplay (treatment has stopped)")
                 } else {
                     EvaluationFactory.undetermined("Has received $matchingDrugDisplay as most recent treatment but unknown if currently still administered")
```
spdeleeuw: I think `when` reads a bit more naturally with >2 if's, but up to you if you want to change:
```
when (requireCurrentAdministration) {
                    false -> EvaluationFactory.pass("Has received $matchingDrugDisplay as most recent treatment")
                    true -> {
                        if (mostRecentMatchingEntryHasStopDate) {
                            EvaluationFactory.fail("Does not currently receive $matchingDrugDisplay (treatment has stopped)")
                        } else {
                            EvaluationFactory.undetermined("Has received $matchingDrugDisplay as most recent treatment but unknown if currently still administered")
                        }
                    }
                }
```

spdeleeuw (approved):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
ninajacobs: I agree!

PR #1568, opened 2026-08-07
ACTIN-3978: Bug fix - should consider only DrugTreatment

ninajacobs:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadTreatmentWithDrugFromSetAsMostRecent.kt
```diff
@@ -8,20 +8,23 @@ import com.hartwig.actin.datamodel.algo.Evaluation
...
 import com.hartwig.actin.datamodel.clinical.treatment.Treatment
 import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
 
 class HasHadTreatmentWithDrugFromSetAsMostRecent(private val drugsToMatch: Set<Drug>, private val requireCurrentAdministration: Boolean) :
     EvaluationFunction {
 
     override fun evaluate(record: PatientRecord): Evaluation {
-        val history = record.oncologicalHistory
+        val relevantHistory = record.oncologicalHistory.filterNot { entry ->
+            entry.allTreatments().isNotEmpty() && entry.allTreatments().all { it is Radiotherapy }
```
spdeleeuw: We could also consider to use `.filterIsInstance<DrugTreatment>()` (also see TreatmentHistoryAnalysis line 35). Since this class is evaluating the most recent drug treatment, why would we exclude radiotherapy but include other non-systemic treatments (e.g. surgery)?

ninajacobs: { as discussed, a surgery may indicate the end of a treatment line, but we will prefer false positives over false negatives }

PR #1569, opened 2026-08-07
Actin 3975 simon

spdeleeuw:


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctions.kt
```diff
@@ -16,33 +16,28 @@ object TreatmentVersusDateFunctions {
...
+                        "undetermined if administered since ${Format.date(minDate)}")
             }
 
-            matchingTreatments.any { isMostRecentTreatmentWithoutStopDate(it, record.oncologicalHistory) == true } -> {
+            matchingPortionOfMostRecentTreatment?.let { stopDateCouldBeSinceMinDate(it, minDate) } == true -> {
                 EvaluationFactory.undetermined(
                     "Undetermined if treatment $predicateDescription may have been administered since " +
                             "${Format.date(minDate)} because it is the last treatment line and stop date missing"
                 )
             }
```
ninajacobs, line 34: Minor - would it be even better to swap the line-29 evaluation and line-34 evaluation? I imagine in the scenario where the treatment of interest has no date AND is the last line, the highest chance of 'success' is if the treatment occurred in the last line.

other minor: I think we could still improve this rule by making some split between surgeries (no time period in which it is 'done' but rather only one moment) vs treatments that take a longer time but may prefer to keep it out of scope..

ninajacobs (approved):
Looks good! I think we should go for this solution?

PR #1571, opened 2026-08-11
ACTIN-3989: Don't evaluate CRC SOC for sarcoma

cbruel:


algo/src/main/kotlin/com/hartwig/actin/algo/soc/StandardOfCareEvaluator.kt
ninajacobs, line 35: Did you consider changing COLORECTAL_CANCER_DOID to COLORECTAL_CARCINOMA_DOID as fix?

ninajacobs: if you think it may be tricky because sometimes we recieve "colorectal cancer" only as input, then I think your fix is good. Though maybe you should also exclude 'rectum sarcoma'?

ninajacobs: and colon sarcoma :)

cbruel: indeed think that could get tricky to use colorectal carcinoma. I would say there is no need to also exclude colon sarcoma and rectum sarcoma specifically, since we always assign the sarcoma doid in case of a type of sarcoma by the LLM :)

cbruel: but we could add it just in case of course

ninajacobs: yeah I think it makes sense to add them too, then we don't depend on the LLM logics. But understood!

ninajacobs (approved):

PR #1572, opened 2026-08-12
ACTIN-3990: Resolve to undetermined for sarcoma and NET

cbruel:
In case any term in X is in the DOID list of "Main cancer types" and a sample configured DOID or one of its parents matches X but Sarcoma (DOID: 1115) or Neuroendocrine tumor (DOID: 169) is also configured, resolve to `UNDETERMINED`.


algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/tumor/PrimaryTumorLocationBelongsToDoid.kt
```diff
@@ -30,10 +32,22 @@ class PrimaryTumorLocationBelongsToDoid(
             val doidTermsTumorBelongsTo = Format.concat(doidsToTerms(doidsTumorBelongsTo))
             val potentialAdenoSquamousMatches = isPotentialAdenoSquamousMatch(tumorDoids!!, doidsToMatch)
             val undeterminedUnderMainCancerTypes = isUndeterminedUnderMainCancerType(tumorDoids, doidsToMatch)
+            val undeterminedTumorTypes = tumorDoids.intersect(UNDETERMINED_TUMOR_TYPES)
```
ninajacobs: Having a bit of difficulty with the intuition around variable name, wonder if undeterminedTumorTypes -> "tumorTypesWithSocByTumorType" and undeterminedTumorTypeMatchesOnlyMainCancerType -> "hasTumorTypeWithSocByTumorType" may be clearer?

Up to you if it's really clearer though. I can see it is longer...

ninajacobs (approved):

PR #1573, opened 2026-08-13
ACTIN-4003: Keep PD-L1 IHC results that are both recent but with different measure in report

ninajacobs:
Fixes problem in https://hartwigmedical.atlassian.net/browse/ACTIN-4003:

### DIFF
--- a/report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
+++ b/report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
@@ -14,10 +14,10 @@ class IhcTestInterpreter {
     private val interpretationBuilder = IhcTestInterpretationBuilder()
 
     fun interpret(ihcTests: List<IhcTest>): List<IhcTestInterpretation> {
-        val latestIhcTestsByItem = ihcTests.groupBy { Pair(it.item, it.measure) }.map { (_, tests) ->
+        val latestIhcTestsByItemAndMeasure = ihcTests.groupBy { Pair(it.item, it.measure) }.map { (_, tests) ->
             tests.maxByOrNull { it.measureDate ?: LocalDate.MIN } ?: tests.first()
         }
-        latestIhcTestsByItem.forEach(::interpret)
+        latestIhcTestsByItemAndMeasure.forEach(::interpret)
         return interpretationBuilder.build()
     }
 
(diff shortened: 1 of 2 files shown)


report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
```diff
@@ -14,7 +14,7 @@ class IhcTestInterpreter {
     private val interpretationBuilder = IhcTestInterpretationBuilder()
 
     fun interpret(ihcTests: List<IhcTest>): List<IhcTestInterpretation> {
-        val latestIhcTestsByItem = ihcTests.groupBy { it.item }.map { (_, tests) ->
+        val latestIhcTestsByItem = ihcTests.groupBy { Pair(it.item, it.measure) }.map { (_, tests) ->
```
cbruel: minor but update val name to `latestIhcTestsByItemAndMeasure`?

cbruel: good to be aware that measure is also often null, so if you have one test with measure = null and one with measure = CPS, on the same date, both will be kept, but I think that's desired behaviour

cbruel (approved):

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
```diff
@@ -14,7 +14,7 @@ class IhcTestInterpreter {
...
             tests.maxByOrNull { it.measureDate ?: LocalDate.MIN } ?: tests.first()
         }
         latestIhcTestsByItem.forEach(::interpret)
```
cbruel: bit unrelated to this PR but is there are reason why for ihc results with a scoreText we don't show the measure on the report (line 35), only for results with a score (line 36) we show the measure? Didn't look into it in detail, but I believe you could now get something like:
PD-L1 negative
PD-L1 negative

(while the first is TPS and second is CPS on the same date)

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ninajacobs: thanks, adding a test for it!

report/src/main/kotlin/com/hartwig/actin/report/interpretation/IhcTestInterpreter.kt
ninajacobs: good point, I agree. Making a PR

cbruel (approved):
Nice!

PR #1574, opened 2026-08-13
ACTIN-3992: Review implementation of treatmentSinceMinDate function

ninajacobs:
In treatmentSinceMinDate we previously checked stopYear() and stopMonth(). These functions return stopYear and stopMonth if available, or maxStopYear and maxStopMonth if available. 

The latter is problematic for this function, because there is a difference in interpretation between stopDate and maxStopDate - namely for stopDate we can evaluate treatmentSinceMinDate with certainty, whereas with maxStopDate we cannot (stop date is maximum, but could be earlier). This means that we cannot be certain that a treatment will have been administered after a date if we rely on maxStopDate.

Key of this PR:
- Consider now certainTreatmentSinceMinDate and potentialTreatmentSinceMinDate.
    - certainTreatmentSinceMinDate: Evaluates stopYear/stopMonth (i.e. excluding maxStopYear/maxStopMonth) and startYear/startMonth. If this boolean is true, we know for _certain_ that the patient had treatment since that date.
    - potentialTreatmentSinceMinDate: Evaluates stopYear()/stopMonth() (including maxStopYear/maxStopMonth). If all dates are null, this function returns also true. Hence if this boolean is true, there has been potentially treatment since that date (note that this one also includes all certain treatments).

In other classes I replaced treatmentSinceMinDate(includeUnknown = true) with potentialTreatmentSinceMinDate and treatmentSinceMinDate(includeUnknown = false) with certainTreatmentSinceMinDate.

Specific for HasHadSomeTreatmentsWithCategoryAndTypeWithIntents:
- This led to failing test in line 163 but I believe this test should actually be undetermined rather than fail.
- Also added two other tests for clarity + some message building improvement.

Specific for HasHadSystemicTherapyWithAnyIntentTest
- Led to failing test but again I believe this test should actually be undetermined rather than fail. When evaluating within weeks, e.g. when a patient has had treatment within 10 weeks, then if the start date is older, it is still possible that the treatment continued into the within 10 weeks window.

Finally: With similar logics I also reviewed treatmentBeforeMaxDate and split it into certainTreatmentBeforeMaxDate and potentialTreatmentBeforeMaxDate. I think our interpretation of HAS_HAD_SYSTEMIC_THERAPY_WITH_ANY_INTENT_X_AT_LEAST_Y_WEEKS_AGO was off before (see changing tests). If treatment occurred after start date, we know the treatment hasn't happened at least X weeks ago, but if we only have stop date and stop date is before or after it, we don't know.


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTest.kt
```diff
@@ -0,0 +1,234 @@
+package com.hartwig.actin.algo.evaluation.treatment
+
+import com.hartwig.actin.datamodel.clinical.TreatmentTestFactory.treatmentHistoryEntry
+import com.hartwig.actin.algo.evaluation.treatment.TreatmentVersusDateFunctions.certainTreatmentSinceMinDate
+import com.hartwig.actin.datamodel.clinical.treatment.Drug
+import com.hartwig.actin.datamodel.clinical.treatment.DrugTreatment
+import com.hartwig.actin.datamodel.clinical.treatment.TreatmentCategory
+import com.hartwig.actin.datamodel.clinical.treatment.history.TreatmentHistoryEntry
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
```
spdeleeuw: This should be `org.junit.jupiter.api.Test` (JUnit 5 instead of JUnit 4 you're using now with this import), otherwise these tests aren't executed with the build.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/IsNotParticipatingInAnotherInterventionalTrial.kt
```diff
@@ -25,6 +24,15 @@ class IsNotParticipatingInAnotherInterventionalTrial(
                 EvaluationFactory.warn("Recent trial treatment - undetermined if patient is participating in another interventional trial")
             }
 
+            record.oncologicalHistory.any {
+                it.isTrial && TreatmentVersusDateFunctions.potentialTreatmentSinceMinDate(
```
spdeleeuw: Might fit on one line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -20,10 +20,12 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
...
         val nonPalliativeIncludedTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) != true }
-        val (recentUncertainTreatments, _) =
-            SystemicTreatmentAnalyser.partitionRecentTreatments(nonPalliativeIncludedTreatments, referenceDate.minusMonths(6), true)
+        val (recentAndPotentiallyRecentTreatments, _) = SystemicTreatmentAnalyser.potentialRecentTreatments(
+            nonPalliativeIncludedTreatments,
+            referenceDate.minusMonths(6)
+        )
         val settingMessage = "$settingDescription setting"
-        val probableCount = palliativeIntentTreatments.size + recentUncertainTreatments.size
+        val probableCount = palliativeIntentTreatments.size + (recentAndPotentiallyRecentTreatments).toSet().size
```
spdeleeuw: I don't think you need the parentheses around `recentAndPotentiallyRecentTreatments`

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -29,12 +29,18 @@ object SystemicTreatmentAnalyser {
...
     }
 
-    fun partitionRecentTreatments(
+    fun certainRecentTreatments(
         treatments: List<TreatmentHistoryEntry>,
         minDate: LocalDate,
-        includeUnknown: Boolean
     ): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> {
-        return treatments.partition { TreatmentVersusDateFunctions.treatmentSinceMinDate(it, minDate, includeUnknown) }
+        return treatments.partition { TreatmentVersusDateFunctions.certainTreatmentSinceMinDate(it, minDate) }
```
spdeleeuw: Naming feels odd here: the function partitions, but you can't see that from the function name which is confusing at the call site. `fun partitionByIntent` directly above keeps the verb and has the same `Pair<List, List>` return type, so these two now read inconsistently with their own neighbour.
`.second` is also not the negation of the name: negating the whole predicate gives `uncertainRecentTreatments`, not `certainUnrecentTreatments`. So the second list holds certainly-older entries *and* unknown-date ones.
I'd suggest to rename this function to  `partitionByCertainlyRecent`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -29,12 +29,18 @@ object SystemicTreatmentAnalyser {
...
+    }
+
+    fun potentialRecentTreatments(
```
spdeleeuw: It seems that, outside `SystemicTreatmentAnalyserTest`, all call sites throw away the second of the pair, so we're partitioning into a Pair without making use of it. You could simplify and just filter.
With that change, you could delete this function entirely, and just filter properties directly using `potentialTreatmentSinceMinDate` (as doing in the function body now).

e.g. in `HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting`: 
val potentiallyRecentTreatments = includedIntentTreatments.filter { potentialTreatmentSinceMinDate(it, referenceDate.minusMonths(MONTHS_TO_SUBTRACT)) }

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -29,12 +29,18 @@ object SystemicTreatmentAnalyser {
         return systemicTreatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
     }
 
-    fun partitionRecentTreatments(
+    fun certainRecentTreatments(
         treatments: List<TreatmentHistoryEntry>,
         minDate: LocalDate,
```
spdeleeuw: Trailing comma could be removed?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -29,12 +29,18 @@ object SystemicTreatmentAnalyser {
...
-        includeUnknown: Boolean
     ): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> {
-        return treatments.partition { TreatmentVersusDateFunctions.treatmentSinceMinDate(it, minDate, includeUnknown) }
+        return treatments.partition { TreatmentVersusDateFunctions.certainTreatmentSinceMinDate(it, minDate) }
+    }
+
+    fun potentialRecentTreatments(
+        treatments: List<TreatmentHistoryEntry>,
+        minDate: LocalDate,
```
spdeleeuw: Trailing comma could be removed?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
jbartletthmf: Is there any reason to call `toSet()` here? The list of treatments shouldn't contain any duplicates, and we won't introduce any by partitioning or filtering.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -20,10 +20,12 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
         val (_, includedIntentTreatments) = SystemicTreatmentAnalyser.partitionByIntent(priorSystemicTreatments, intentsToIgnore)
         val palliativeIntentTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) == true }
         val nonPalliativeIncludedTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) != true }
```
jbartletthmf: This is a good place to use `partition`:
```suggestion
val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) = includedIntentTreatments.partition { it.intents?.contains(Intent.PALLIATIVE) == true }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSomeTreatmentsWithCategoryAndTypeWithIntents.kt
```diff
@@ -69,11 +70,16 @@ class HasHadSomeTreatmentsWithCategoryAndTypeWithIntents(
 
     private fun drugTypeString(entries: List<TreatmentHistoryEntry>): String {
         return allowedTypes?.let {
-            " ${Format.concatItemsWithAnd(it.intersect(entries.flatMap { e -> e.treatments }.flatMap { t -> t.types() }.toSet()))}"
+            val types = Format.concatItemsWithAnd(it.intersect(entries.flatMap { e -> e.treatments }.flatMap { t -> t.types() }.toSet()))
+            if (types.isNotEmpty()) " $types" else ""
```
jbartletthmf:
```suggestion
            entries.flatMap(TreatmentHistoryEntry::treatments)
                .flatMap(Treatment::types)
                .filter(allowedTypes::contains)
                .ifEmpty { null }
                ?.let { " ${Format.concatItemsWithAnd(it)}" }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSpecificTreatmentSinceDate.kt
```diff
@@ -9,8 +13,32 @@ import java.time.LocalDate
...
     override fun evaluate(record: PatientRecord): Evaluation {
-        return TreatmentVersusDateFunctions.evaluateTreatmentMatchingPredicateSinceDate(
-            record, minDate, "matching '${treatment.display()}'"
-        ) { it.name == treatment.name }
+        val matchingTreatments = record.oncologicalHistory
+            .mapNotNull { entry ->
+                TreatmentHistoryEntryFunctions.portionOfTreatmentHistoryEntryMatchingPredicate(
+                    entry,
+                    { it.name == treatment.name })
+            }
```
jbartletthmf: Formatting seems off here? Also, the final function parameter can come outside of the parentheses:
```suggestion
        val matchingTreatments = record.oncologicalHistory.mapNotNull { entry ->
            TreatmentHistoryEntryFunctions.portionOfTreatmentHistoryEntryMatchingPredicate(entry) {
                it.name == treatment.name
            }
        }
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTherapyWithAnyIntent.kt
```diff
@@ -68,18 +74,17 @@ class HasHadSystemicTherapyWithAnyIntent(
     }
 
     private fun anyTreatmentPotentiallySinceMinDate(treatmentEntries: Iterable<TreatmentHistoryEntry>): Boolean {
-        return refDate == null || treatmentEntries.any { treatmentSinceMinDate(it, refDate, true) }
+        return refDate == null || treatmentEntries.any { potentialTreatmentSinceMinDate(it, refDate) }
     }
 
     private fun anyTreatmentPotentiallyBeforeMaxDate(treatmentEntries: Iterable<TreatmentHistoryEntry>): Boolean {
-        return refDate == null || treatmentEntries.any { treatmentBeforeMaxDate(it, refDate, true) }
+        return refDate == null || treatmentEntries.any { certainTreatmentBeforeMaxDate(it, refDate) }
```
jbartletthmf: Should use `potentialTreatmentBeforeMaxDate`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTherapyWithAnyIntent.kt
```diff
@@ -33,16 +35,20 @@ class HasHadSystemicTherapyWithAnyIntent(
...
             }
 
-            evaluateWithinWeeks == true && evaluateTreatments(matchingTreatments, ::treatmentSinceMinDate, false) -> {
+            evaluateWithinWeeks == true && evaluateTreatments(matchingTreatments, ::certainTreatmentSinceMinDate) -> {
                 EvaluationFactory.pass("Received $intentsLowercase systemic therapy within the last $weeks weeks")
             }
 
-            evaluateWithinWeeks == false && evaluateTreatments(matchingTreatments, ::treatmentBeforeMaxDate, false) -> {
+            evaluateWithinWeeks == false && evaluateTreatments(matchingTreatments) { entry, date ->
+                certainTreatmentBeforeMaxDate(entry, date)
```
jbartletthmf: could use a function reference here as above

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTherapyWithAnyIntent.kt
```diff
@@ -33,16 +35,20 @@ class HasHadSystemicTherapyWithAnyIntent(
...
+            } -> {
                 EvaluationFactory.pass("Received $intentsLowercase systemic therapy at least $weeks weeks ago")
             }
 
-            (evaluateWithinWeeks == true && evaluateTreatments(matchingTreatments, ::treatmentSinceMinDate, true)) ||
-                    (evaluateWithinWeeks == false && evaluateTreatments(matchingTreatments, ::treatmentBeforeMaxDate, true)) -> {
+            (evaluateWithinWeeks == true && evaluateTreatments(matchingTreatments, ::potentialTreatmentSinceMinDate)) ||
+                    (evaluateWithinWeeks == false && evaluateTreatments(matchingTreatments) { entry, date ->
+                        potentialTreatmentBeforeMaxDate(entry, date)
```
jbartletthmf: could use a function reference?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -20,10 +20,12 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
...
-        val (recentUncertainTreatments, _) =
-            SystemicTreatmentAnalyser.partitionRecentTreatments(nonPalliativeIncludedTreatments, referenceDate.minusMonths(6), true)
+        val (recentAndPotentiallyRecentTreatments, _) = SystemicTreatmentAnalyser.potentialRecentTreatments(
+            nonPalliativeIncludedTreatments,
+            referenceDate.minusMonths(6)
```
jbartletthmf: Should this reference `MONTHS_TO_SUBTRACT` like `HasHadSystemicTreatmentWithUnknownOrSpecificIntentAndSetting`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/washout/HasRecentlyReceivedTrialMedication.kt
```diff
@@ -19,26 +19,30 @@ class HasRecentlyReceivedTrialMedication(
             return EvaluationFactory.undetermined("Recent trial medication undetermined (required stop date prior to registration date)")
         }
 
-        val hadRecentTrialTreatment =
-            record.oncologicalHistory.any { it.isTrial && TreatmentVersusDateFunctions.treatmentSinceMinDate(it, minStopDate, false) }
+        val hadRecentTrialTreatment = record.oncologicalHistory.any {
+            it.isTrial && TreatmentVersusDateFunctions.certainTreatmentSinceMinDate(
+                it,
+                minStopDate
+            )
```
jbartletthmf: Could be one line?

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTest.kt
```diff
@@ -0,0 +1,234 @@
...
+private val OLDER_DATE = TARGET_DATE.minusYears(1)
+
+class TreatmentVersusDateFunctionsTest {
+
+    @Test
+    fun `Should return true for certainTreatmentSinceMinDate if stop date is after target date`() {
+        val treatment = treatment(stopYear = RECENT_DATE.year, stopMonth = RECENT_DATE.monthValue)
+        val result = certainTreatmentSinceMinDate(treatment, TARGET_DATE)
+
+        assertThat(result).isTrue
```
jbartletthmf: `result` could be inlined here and below.

pom.xml
```diff
@@ -26,7 +26,7 @@
         <maven.compiler.release>${java.version}</maven.compiler.release>
         <kotlin.compiler.jvmTarget>${java.version}</kotlin.compiler.jvmTarget>
 
-        <actin-datamodel.version>4.2.0</actin-datamodel.version>
+        <actin-datamodel.version>4.5.0-beta.1</actin-datamodel.version>
```
jbartletthmf: Is there a related datamodel PR to consider here? Otherwise, we shouldn't need to use a beta version in the pom.

ninajacobs:
Thanks both, hope to find the time to address your feedback today. This one should come with datamodel PR indeed: https://github.com/hartwigmedical/actin-datamodel/pull/46/changes

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
ninajacobs: Sorry the "(" and ")" and the toSet are remainders from when I had a slightly different implementation first (combining two lists). Will remove and also clarify the variable name.

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTest.kt
ninajacobs: Thanks, didn't know this (also don't know how I made sure I got this import :) )

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
ninajacobs: agreed. The code was already as is and ignored it, but I agree this is cleaner, will adjust!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
ninajacobs: is i our convention to remove it ?  I thought it's also fine to have it.

jbartletthmf: I don't believe we set a convention for it and [the Kotlin style guide is mildly in favour of it](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas), so I don't mind keeping them.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadSystemicTherapyWithAnyIntent.kt
ninajacobs: Yes, ugly, thanks for catching this one.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
ninajacobs: I think that's fair.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -17,15 +18,16 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
...
-        val (_, includedIntentTreatments) = SystemicTreatmentAnalyser.partitionByIntent(priorSystemicTreatments, intentsToIgnore)
-        val palliativeIntentTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) == true }
-        val nonPalliativeIncludedTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) != true }
-        val (recentAndPotentiallyRecentTreatments, _) = SystemicTreatmentAnalyser.potentialRecentTreatments(
-            nonPalliativeIncludedTreatments,
-            referenceDate.minusMonths(6)
-        )
+        val (_, includedIntentTreatments) = SystemicTreatmentAnalyser.partitionTreatmentsByIntent(priorSystemicTreatments, intentsToIgnore)
+        val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) = includedIntentTreatments.partition { it.intents?.contains(Intent.PALLIATIVE) == true }
+        val recentAndPotentiallyRecentNonPalliativeTreatments = nonPalliativeIncludedTreatments.filter {
```
jbartletthmf: Can we just call this `potentiallyRecentNonPalliativeTreatments`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -17,15 +18,16 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
...
+            potentialTreatmentSinceMinDate(
+                it,
+                referenceDate.minusMonths(MONTHS_TO_SUBTRACT)
+            )
```
jbartletthmf: Could be one line?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -22,27 +22,20 @@ object SystemicTreatmentAnalyser {
         UNKNOWN
     }
 
-    fun partitionByIntent(
-        systemicTreatments: List<TreatmentHistoryEntry>,
+    fun partitionTreatmentsByIntent(
+        treatments: List<TreatmentHistoryEntry>,
```
jbartletthmf: Do these functions still belong in `SystemicTreatmentAnalyser`? `partitionTreatmentsByCertainlySinceMinDate`, for example, could go in something like `TreatmentVersusDateFunctions`.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -22,27 +22,20 @@ object SystemicTreatmentAnalyser {
...
         intentsToIgnore: Set<Intent>
```
jbartletthmf: I would rename this to `intents` since it's up to the caller which intents are used and which are ignored.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -22,27 +22,20 @@ object SystemicTreatmentAnalyser {
...
     ): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> {
-        return systemicTreatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
+        return treatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
     }
 
-    fun certainRecentTreatments(
+    fun partitionTreatmentsByCertainlySinceMinDate(
```
jbartletthmf: How about `partitionTreatmentsByKnownOccurrenceSinceMinDate`? or `Certain` in place of `Known`?

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
```diff
@@ -17,15 +18,16 @@ class HasHadAtMostSystemicTreatmentLinesInSpecificSetting(
...
         val priorSystemicTreatments = record.oncologicalHistory.filter { it.treatments.any(Treatment::isSystemic) }
-        val (_, includedIntentTreatments) = SystemicTreatmentAnalyser.partitionByIntent(priorSystemicTreatments, intentsToIgnore)
-        val palliativeIntentTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) == true }
-        val nonPalliativeIncludedTreatments = includedIntentTreatments.filter { it.intents?.contains(Intent.PALLIATIVE) != true }
-        val (recentAndPotentiallyRecentTreatments, _) = SystemicTreatmentAnalyser.potentialRecentTreatments(
-            nonPalliativeIncludedTreatments,
-            referenceDate.minusMonths(6)
-        )
+        val (_, includedIntentTreatments) = SystemicTreatmentAnalyser.partitionTreatmentsByIntent(priorSystemicTreatments, intentsToIgnore)
+        val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) = includedIntentTreatments.partition { it.intents?.contains(Intent.PALLIATIVE) == true }
```
jbartletthmf: We can use the partition function here too since it works the same way whether you want to include or exclude the given intent:
```suggestion
        val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) = SystemicTreatmentAnalyser.partitionTreatmentsByIntent(includedIntentTreatments, setOf(Intent.PALLIATIVE))
```

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
```diff
@@ -22,27 +22,20 @@ object SystemicTreatmentAnalyser {
...
     }
 
-    fun partitionByIntent(
-        systemicTreatments: List<TreatmentHistoryEntry>,
+    fun partitionTreatmentsByIntent(
+        treatments: List<TreatmentHistoryEntry>,
         intentsToIgnore: Set<Intent>
     ): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> {
-        return systemicTreatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
+        return treatments.partition { it.intents?.any { intent -> intentsToIgnore.contains(intent) } == true }
```
jbartletthmf:
```suggestion
        return treatments.partition { it.intents?.any(intentsToIgnore::contains) == true }
```

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTest.kt
```diff
@@ -19,193 +19,145 @@ class TreatmentVersusDateFunctionsTest {
     @Test
     fun `Should return true for certainTreatmentSinceMinDate if stop date is after target date`() {
         val treatment = treatment(stopYear = RECENT_DATE.year, stopMonth = RECENT_DATE.monthValue)
-        val result = certainTreatmentSinceMinDate(treatment, TARGET_DATE)
-
-        assertThat(result).isTrue
+        assertThat( certainTreatmentSinceMinDate(treatment, TARGET_DATE)).isTrue
```
jbartletthmf: extra space

algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentVersusDateFunctionsTest.kt
```diff
@@ -19,193 +19,145 @@ class TreatmentVersusDateFunctionsTest {
...
+        assertThat(certainTreatmentSinceMinDate(treatment, TARGET_DATE)).isFalse
     }
 
     @Test
     fun `Should return false for certainTreatmentSinceMinDate if both stop date and start date are unknown`() {
         val treatment = treatment()
-        val result = certainTreatmentSinceMinDate(treatment, TARGET_DATE)
-
-        assertThat(result).isFalse
+        assertThat( certainTreatmentSinceMinDate(treatment, TARGET_DATE)).isFalse
```
jbartletthmf: extra space

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
ninajacobs: I found this clearer, but no strong opinion, happy to change.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
ninajacobs: I guess partitionTreatmentsByIntent could go into TreatmentHistoryEntryFunctions

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
ninajacobs: I thought so too and I started rewriting this part, but then the title of the test here

    fun `Should place all treatments in included when no intents are ignored`() {
        val entries = listOf(
            systemicTreatmentHistoryEntry.copy(intents = setOf(Intent.CURATIVE)),
            systemicTreatmentHistoryEntry.copy(intents = null)
        )
        val (excluded, included) = partitionTreatmentsByIntentsToIgnore(entries, emptySet())
        assertThat(excluded).isEmpty()
        assertThat(included).containsExactlyElementsOf(entries)
    }

doesn't make sense anymore. The original logic for this test was in the context of wanting to ignore intents - which doesn't make sense anymore if we start looking for an empty set now.

Though we never use emptyset() in code now anywhere for this function, so I think it's okay if the readability of the functions improves at the cost of this test. We can create a solution with intentsToIgnore = null instead and only partition if intentsToIgnore is not null if a case comes up.

I will remove this test.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
ninajacobs: I wanted to use certain since it aligns with certainTreatmentSinceMinDate, will rename to Certain!

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/HasHadAtMostSystemicTreatmentLinesInSpecificSetting.kt
ninajacobs: will rename the function name to make it clearer too.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/SystemicTreatmentAnalyser.kt
jbartletthmf: I don't think it's safe to assume that intents is never empty, because someone could easily add a new rule that uses e.g. `HasHadAtMostSystemicTreatmentLinesInSpecificSetting` with `intentsToIgnore = emptySet()`. Can't we just rename? E.g.
```kotlin
fun `Should return all treatments in non-matching list when intent list is empty`() {
    val entries = listOf(
        systemicTreatmentHistoryEntry.copy(intents = setOf(Intent.CURATIVE)),
        systemicTreatmentHistoryEntry.copy(intents = null),
        systemicTreatmentHistoryEntry.copy(intents = emptySet())
    )
    val (matching, nonMatching) = partitionTreatmentsByIntents(entries, emptySet())
    assertThat(matching).isEmpty()
    assertThat(nonMatching).containsExactlyElementsOf(entries)
}
```
I added a test entry with `intents = emptySet()` to cover that case as well.

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentHistoryEntryFunctions.kt
```diff
@@ -29,6 +30,13 @@ object TreatmentHistoryEntryFunctions {
     fun TreatmentHistoryEntry.containsTreatment(treatmentNameToFind: String) =
         allTreatments().any { it.name.equals(treatmentNameToFind, ignoreCase = true) }
 
+    fun partitionTreatmentsByIntent(
+        treatmentHistory: List<TreatmentHistoryEntry>,
+        intents: Set<Intent>
+    ): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> {
+        return treatmentHistory.partition { it.intents?.any(intents::contains) == true }
+    }
```
jbartletthmf: If you like, you could use an extension function here to make this a little easier to work with. This essentially lets you define a new function for the "receiver", which in this case is a `Collection<TreatmentHistoryEntry>`, instead of providing it as a parameter:
```kotlin
    fun Collection<TreatmentHistoryEntry>.partitionByIntent(intents: Set<Intent>): Pair<List<TreatmentHistoryEntry>, List<TreatmentHistoryEntry>> =
        partition { it.intents?.any(intents::contains) == true }
```
`partition` here implicitly works on the receiver collection. It's equivalent to calling `this.partition` since `this` refers to the receiver.

What I like about this is it turns calls like:
```kotlin
        val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) =
            TreatmentHistoryEntryFunctions.partitionTreatmentsByIntent(
                includedIntentTreatments,
                setOf(Intent.PALLIATIVE)
            )
```
into:
```kotlin
        val (palliativeIntentTreatments, nonPalliativeIncludedTreatments) =
            includedIntentTreatments.partitionByIntent(setOf(Intent.PALLIATIVE))
```
which is not just easier to type, it also reads more intuitively. Note that you have to import the extension function if it's not in scope.

jbartletthmf (approved):

algo/src/main/kotlin/com/hartwig/actin/algo/evaluation/treatment/TreatmentHistoryEntryFunctions.kt
ninajacobs: Thanks John, I can see why it reads more nicely within the class.

PR #1575, opened 2026-08-17
ACTIN-3929: Restrict ICD code slot assignment (main vs extension)

spdeleeuw:
`actin-clinical-llm`: https://github.com/hartwigmedical/actin-clinical-llm/pull/38
`actin-clinical`: https://github.com/hartwigmedical/actin-clinical/pull/496

- `IcdCode(mainCode, extensionCode?)` had no notion of which codes belong in which slot, so nothing stopped am extension code being stored as a main code. ICD matching in `IcdModel` is strictly slot-directional and every allergy/intolerance evaluator already assumes condition = main, substance = extension, so this PR aligns the model with what algo believes.
- Slot identity is only derivable from ICD-11 chapter X ("Extension Codes"), which IcdDeserializer already computes and then discarded. It's now persisted on `IcdNode.isExtension` and `IcdModel` derives per-slot title maps from it. 
- `resolveCodeForTitle`, `isValidIcdTitle` and `isValidIcdCode` are now slot-directional. `invalidTitleReason` / `invalidMainTitleReason` / `invalidExtensionTitleReason` give consumers a shared message distinguishing unknown title from known title, wrong slot.

### DIFF
--- a/common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
+++ b/common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
@@ -63,10 +63,10 @@ class IcdModel(
             else -> invalidMainTitleReason(slots.first) ?: slots.second?.let(::invalidExtensionTitleReason)
         }
 
-    fun invalidMainTitleReason(mainTitle: String): String? =
+    private fun invalidMainTitleReason(mainTitle: String): String? =
         unknownOrMisplacedTitleReason(mainTitle, mainTitleToCodeMap, extensionTitleToCodeMap, EXTENSION_SLOT, MAIN_SLOT)
 
-    fun invalidExtensionTitleReason(extensionTitle: String): String? =
+    private fun invalidExtensionTitleReason(extensionTitle: String): String? =
         unknownOrMisplacedTitleReason(extensionTitle, extensionTitleToCodeMap, mainTitleToCodeMap, MAIN_SLOT, EXTENSION_SLOT)
 
     private fun unknownOrMisplacedTitleReason(
@@ -79,7 +79,7 @@ class IcdModel(
         when (title.lowercase()) {
             in expectedSlotTitles -> null
             in otherSlotTitles -> "ICD title [$title] is $otherSlot and cannot be used as $expectedSlot"
-            else -> "ICD title [$title] is not known - check for existence in ICD model"
+            else -> "ICD title [$title] is not known. Check for existence in ICD model"
         }
 
     private fun malformedTitleReason(icdTitle: String): String =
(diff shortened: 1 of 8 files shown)


common/src/test/kotlin/com/hartwig/actin/icd/IcdModelTest.kt
```diff
@@ -49,14 +52,76 @@ class IcdModelTest {
     @Test
     fun `Should handle title validation correctly for regular and extended titles`() {
         assertThat(icdModel.isValidIcdTitle("targetMainTitle")).isTrue()
-        assertThat(icdModel.isValidIcdTitle("targetExtensionTitle")).isTrue()
```
spdeleeuw, line 52: The function takes a combined title string now, so a string containing an extension only is invalid. See test coverage below.

common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
```diff
@@ -16,30 +20,74 @@ class IcdModel(
     val titleToCodeMap: Map<String, String>
 ) {
 
-    fun isValidIcdTitle(icdTitle: String): Boolean {
-        val titles = icdTitle.split('&')
-        return titles.size in 1..2 && titles.map(String::lowercase).all(titleToCodeMap::containsKey)
+    val mainTitleToCodeMap = titleToCodeMapForSlot(extensionSlot = false)
+    val extensionTitleToCodeMap = titleToCodeMapForSlot(extensionSlot = true)
```
jbartletthmf: I think this should happen in `create()` in the companion object and these could be passed into the constructor. We could partition the nodes and compose the maps for efficiency:
```kotlin
val (extensionNodes, mainNodes) = icdNodes.partition(IcdNode::isExtension)
val mainTitleToCodeMap = mainNodes.associate { it.title.lowercase() to it.code }
val extensionTitleToCodeMap = extensionNodes.associate { it.title.lowercase() to it.code }
val titleToCodeMap = mainTitleToCodeMap + extensionTitleToCodeMap
```
This is assuming we still need a unified title-to-code map, which may no longer be true.

common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
```diff
@@ -16,30 +20,74 @@ class IcdModel(
...
+
+    private fun titleToCodeMapForSlot(extensionSlot: Boolean): Map<String, String> =
+        codeToNodeMap.values.filter { it.isExtension == extensionSlot }
+            .associate { it.title.lowercase() to it.code }
```
jbartletthmf: Private functions should go at the bottom of the class.

jbartletthmf (approved):
minor comments

common/src/main/kotlin/com/hartwig/actin/icd/IcdModel.kt
spdeleeuw: Indeed not needed anymore :)

PR #1580, opened 2026-08-21
ACTIN-4016: Update algo messages

cbruel:


algo/src/test/kotlin/com/hartwig/actin/algo/evaluation/bloodtransfusion/HasHadRecentBloodTransfusionTest.kt
```diff
@@ -12,13 +12,29 @@ class HasHadRecentBloodTransfusionTest {
...
+            EvaluationResult.FAIL,
+            function.evaluate(BloodTransfusionTestFactory.withBloodTransfusions(emptyList())),
+            "Has not received recent thrombocyte blood transfusion"
+        )
         val tooOld = create(TransfusionProduct.THROMBOCYTE, minDate.minusWeeks(4))
-        assertEvaluation(EvaluationResult.FAIL, function.evaluate(BloodTransfusionTestFactory.withBloodTransfusion(tooOld)))
+        assertEvaluation(
+            EvaluationResult.FAIL,
+            function.evaluate(BloodTransfusionTestFactory.withBloodTransfusion(tooOld)),
+            "Has not received recent thrombocyte blood transfusion"
```
andreia-hartwig: As an improvement, you could use constants for the expected messages. That would make it easier to immediately see which messages are produced by each class.

andreia-hartwig (approved):
I was not expecting to see changes in the source code with this PR.
Given the size of the PR I did not check every single test but did random checks in all packaged and it look good.

